PR #4, opened 2019-02-22
fairer benchmark with -O2 for everyone

gelisam:
too-fast-too-free was compiled without -O2, which put it at a
disadvantage compared to the code from the competing libraries, which is
inside the benchmark project and was thus compiled with -O2.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -26,6 +26,7 @@ dependencies:
 - mmorph
 
 library:
+  ghc-options: -O2
   source-dirs: src
 
 executables:
@@ -54,7 +55,6 @@ benchmarks:
   too-fast-too-free-bench:
     source-dirs: bench
     main: countDown.hs
-    ghc-options: -O2
     dependencies:
     - criterion
     - free


gelisam:
I'm not sure whether the `freer-simple` and `mtl` libraries are compiled with `-O2`, though?

isovector:
Good question---let me check.

isovector:
Looks like yes. I made the following change to `freer-simple` and still see O(700us) for its benchmarks:

```diff
diff --git a/package.yaml b/package.yaml
index 890fef3..99e4157 100644
--- a/package.yaml
+++ b/package.yaml
@@ -33,6 +33,7 @@ ghc-options:
 - -Wincomplete-record-updates
 - -Wincomplete-uni-patterns
 - -Wredundant-constraints
+- -O2
```

gelisam:
I now think this was a mistake, sorry.

With this PR, we have: default optimizations in `freer-simple:lib`, `-O2` in `too-fast-too-free:lib`, and `-O2` in `too-fast-too-simple:bench`. With that configuration, `stack bench` gives the expected results.

You tried (and I confirm): `-O2` optimizations in `freer-simple:lib`, `-O2` in `too-fast-too-free:lib`, and `-O2` in `too-fast-too-simple:bench`. With that configuration, `stack bench` gives the expected results.

It is tempting to conclude that the default optimization is `-O2`. But that doesn't explain what we observed earlier, when we had: default optimizations in `freer-simple:lib`, default optimizations in `too-fast-too-free:lib`, and `-O2` in `too-fast-too-simple:bench`. With that configuration, `stack bench` gives poorer results than expected for `too-fast-too-simple`, and `stack bench --ghc-options=-O2 --force-dirty` gives the expected results. If the default optimization is `-O2`, then forcing `--ghc-options=-O2` shouldn't change anything.

Confusingly, running `stack bench` _after_ running `stack bench --ghc-options=-O2 --force-dirty` give the expected results again, so there are some side-effects involved here which might have affected some of our measurements :(

Anyway. Using the default optimizations on `freer-simple:lib` doesn't seem to make much of a difference, but using the default optimizations on `too-fast-too-free:lib` does seem to make a difference. Since `freer-simple` is using the default optimizations on its library and `-O2` on its benchmark, I think it would be more fair to follow suit and use the default optimizations on `too-fast-too-free:lib`. Not to worry, though: I can recover the expected results once again by moving the `countDown` implementation from `too-fast-too-free:lib` to `too-fast-too-free:bench`. PR incoming.

PR #5, opened 2019-02-22
Fair benchmark

gelisam:
As explained [here](https://github.com/isovector/too-fast-too-free/pull/4#issuecomment-466264802), I think it would be fairer to the other libraries being compared if `too-fast-too-free:lib` was _not_ compiled with `-O2` after all. The proper solution to get `stack bench` to produce as good results as `stack --ghc-options="-O2" --force-dirty` is apparently to move `countDown` to the benchmark file. It's certainly more uniform that way.


app/Main.hs
```diff
@@ -1,7 +0,0 @@
-module Main where
```
isovector: I'd prefer not to chop this file out entirely. I agree it should disappear in the long term, but it's still useful for inspecting core.

isovector:
Not sure how to merge this any more, or even if it was helpful. Feel free to reopen later!

PR #11, opened 2019-04-11
Replace Semantic with Sem for easy typings

CaptJakk:
Because nobody has time to type that shit out :)

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -1,10 +1,9 @@
 module Polysemy
   ( -- * Core Types
-    Semantic ()
-  , Sem
+    Sem ()
   , Member
 
-  -- * Running Semantic
+  -- * Running Sem
   , run
   , runM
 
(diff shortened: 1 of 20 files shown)


src/Polysemy.hs
```diff
@@ -112,12 +113,13 @@ module Polysemy
...
+  )
+where
 
-import Data.Typeable
-import Polysemy.Internal
-import Polysemy.Internal.Combinators
-import Polysemy.Internal.TH.Effect
-import Polysemy.Internal.TH.Performance
-import Polysemy.Internal.Tactics
+import           Data.Typeable
```
CaptJakk: looks like my brittany config murdered this. Will fix and repush

isovector:
I figure we might as well just go all the way and `%s/Semantic/Sem`.

CaptJakk:
I mean, I'm down, I can fix and reupload.

isovector:
let's do it!

CaptJakk:
Done.

isovector:
Looks good to me. Can you follow up with `{-# DEPRECATED #-}` exports for all of the old names? Then we can release this as a minor version change.

CaptJakk:
totally can, question is, can you mark newtype constructors as deprecated?

isovector:
Just add `type Semantic = Sem` and `runSemantic = runSem` and deprecate
both?

CaptJakk:
take a look here. export lists might need a little love, but otherwise I think this is close

CaptJakk:
bump

isovector:
Sorry -- in the middle of traveling! Hopefully I'll have some time to look at this tomorrow.

PR #12, opened 2019-04-12
Extend examples, make them easy to copy and paste

jhenahan:
For the purpose of getting new users off on the right foot immediately, I added the necessary pragmas to get these working, and built out a trivial example of how to actually use `runConsoleIO`. I wanted to write something like `runConsoleBoring` as an example of mocking using `Input` and `Output` maybe(?), but I wasn't sure how to go about it. I also haven't tried writing a `Resource` example, yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,6 @@ import Polysemy.Input
 import Polysemy.Output
 import Polysemy.Error
 import Polysemy.Resource
-import qualified Control.Exception as X
 
 -- Using Teletype effect from above
 
--- a/README.md
+++ b/README.md
@@ -65,10 +65,8 @@ Teletype effect:
 
 ```haskell
 {-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE BlockArguments, DataKinds, FlexibleContexts, GADTs, LambdaCase,
-             PolyKinds, RankNTypes, ScopedTypeVariables, TypeApplications,
-             TypeOperators, TypeFamilies, UnicodeSyntax
-#-}
+{-# LANGUAGE LambdaCase, BlockArguments #-}
+{-# LANGUAGE GADTs, FlexibleContexts, TypeOperators, DataKinds, PolyKinds #-}
 
 import Polysemy
 import Polysemy.Input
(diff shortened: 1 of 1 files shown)


isovector:
Cool, great idea! You can write `runConsoleBoring` as follows:

```haskell
runConsoleBoring :: [String] -> Semantic (Console ': r) a -> Semantic r ([String], a)
runConsoleBoring inputs = runFoldMapOutput id . runListInput inputs . reinterpret2 \case
  ReadLine -> input
  WriteLine msg -> output msg
```

jhenahan:
GHC seems unhappy with that unless I give it

```haskell
runConsoleBoring ::
  (Member (Input String) r, Member (Output String) r) =>
    [String] ->
      Semantic (Console ': r) a -> Semantic r ([String], a)
```

But when I go to write

```haskell
echoBoring :: Semantic '[Input String, Output String] ([String], a)
echoBoring = runConsoleBoring echo

revBoring :: Semantic '[Input String, Output String] ([String], a)
revBoring = runConsoleBoring rev
```

I get errors like:
```
• Ambiguous use of effect 'Input'
      Possible fix:
        add (Member (Input [Char]) r1) to the context of 
          the type signature
      If you already have the constraint you want, instead
        add a type application to specify
          '[Char]' directly
    • In the expression: runConsoleBoring echo
      In an equation for ‘echoBoring’: echoBoring = runConsoleBoring echo
```

First, extreme praise for the very nice error messages. They're really helpful when you're exploring. My only problem is that I _think_ I have the constraint I want, but I don't know for sure that I do. If I do, I don't know _where_ the type application needs to go. If there's a chunk of the docs (which, by the way, are also very good) that makes this clear (or a way to make the errors just a skosh more informative), I may have missed it. I'll go reading again, but if I'm just missing the obvious I wouldn't mind a nudge. 😆

isovector:
Whoops, try this instead! I forgot runInputList gives you a maybe :)

```haskell
runConsoleBoring :: [String] -> Semantic (Console ': r) a -> Semantic r ([String], a)
runConsoleBoring inputs
  = runFoldMapOutput (:[])
  . runListInput inputs
  . reinterpret2 \case
      ReadLine -> maybe "" id <$> input
      WriteLine msg -> output msg
```

jhenahan:
That would explain why it was trying to get me to add a constraint with a `Maybe String` in there. 😄 

That does let me write

```haskell
echoBoring :: [String] -> Semantic '[] ([String], ())
echoBoring = flip unConsoleBoring echo

revBoring :: [String] -> Semantic '[] ([String], ())
revBoring = flip runConsoleBoring rev
```
which is quite nice.

I'm a little confused by `runListInput`. The documentation suggests that it'll run my `Input` effect on each element of the list, but if I try something like

```haskell
fst $ run $ echoBoring ["hello", "there"]
```
I only get `["hello"]` back, when I'd expect `["hello", "there"]`.

Rewriting as

```haskell
runConsoleBoring :: String -> Semantic (Console ': r) a -> Semantic r (String, a)
runConsoleBoring i
  = runFoldMapOutput id
  . runConstInput i
  . reinterpret2 \case
      ReadLine -> input
      WriteLine msg -> output msg
```

seems in line with the idea (read a line of input, then spit a line out), though I'd still be interested in figuring out what I'm missing about `runListInput`, which may bring me back around on using the `[String]` version.

With that definition, I can write

```haskell
echoBoring :: String -> Semantic '[] (String, ())
echoBoring = flip runConsoleBoring echo

revBoring :: String -> Semantic '[] (String, ())
revBoring = flip runConsoleBoring rev

test :: [String]
test = do
  let a = fst . run $ echoBoring "hello"
      b = fst . run $ revBoring "there"
  [a, b]
```

What I really want to write is something like

```haskell
test :: [String] -> [String]
test = ??? (fst . run) [echoBoring, revBoring]
```

So that I can write something like

```haskell
test ["hello", "there"]
-- ["hello", "ereht"]

test ["even", "more", "words"]
-- ["even", "erom", "words"]
```

But I'm not really sure what `???` would be.

isovector:
The problem is that you're only calling `readLine` once in echo and rev; the semantics of `runListInput` is that they will provide a different element of the list for each call to `input` (or `readLine` in your example). Try this instead:

```haskell
echo :: Member Console r => Semantic r ()
echo = do
  msg <- readLine
  writeLine msg
  when (msg /= "") echo
```

jhenahan:
Refactored the examples to use CPS (so `echo` and `rev` are just fixpoints), which made it much more natural to express what I was going for, and added `Done` to the console API. If this has unexpected performance implications that I'm not seeing (or there's just a better way to write it) let me know. I'm going to try to build up a `Resource` example before I'll consider this ready for a merge.

jhenahan:
Added a `Resource` example:

```haskell
fileOpen :: Member (Lift IO) r => String -> IOMode -> Semantic (Resource ': r) ()
fileOpen name mode = do
  bracket (sendM $ openFile name mode)
          (sendM . hClose)
          \handle -> do
            f <- sendM $ hGetContents handle
            sendM $ putStrLn f

cat :: String -> IO ()
cat = runM . runResource runM . flip fileOpen ReadMode
```

I kind of just threw `sendM`s at `fileOpen` until it compiled, so if there's a nicer way to write that then I'm happy to update.

isovector:
Thanks for tackling this documentation stuff!

Using bracket to show off file access doesn't seem like a great example
however; after all, you can do that out of the box with regular-strength
`base`.

I'd prefer to use something like the `Error` effect and maybe the
`Teletype` effect from above; to show how the cleanup action from `bracket`
is called even if a custom (non-IO) error is thrown via `throw`. This is a
nontrivial example that AFAIK can't be done via ExceptT and MonadUnliftIO.

jhenahan:
Sounds good to me! I'll throw some time at that idea and let you know if I get stuck.

isovector:
What's the state of this? Ready for review?

jhenahan:
Sorry, got caught up in work fires. I've set aside some time today, so I'll try to get this wrapped up and ready.

jhenahan:
```haskell
data CustomException = ThisException | ThatException deriving Show
instance X.Exception CustomException

program :: Members '[Resource, Teletype, Error CustomException] r => Sem r ()
program = catch work $ \e -> writeTTY ("Caught " ++ show e)
  where work = bracket (readTTY) (const $ writeTTY "hello") $ \input -> do
          writeTTY "entering bracket"
          when (input == "explode") $ throw ThisException
          writeTTY input
          writeTTY "exiting bracket"
```

How about that? I can run it with `runM . runResource runM . runError @CustomException . runTeletypeIO $ program`, though I'm having trouble getting `runErrorInIO` to play nice.

jhenahan:
@isovector Updated examples with the new changes, and tried to demonstrate `Teletype`, `Error`, and `Resource` effects all working together.

README.md
```diff
@@ -110,6 +207,27 @@ runResource finish = interpretH $ \case
         runIt = finish .@ runResource
 
     sendM $ X.bracket (runIt a) (runIt . d) (runIt . u)
+
+data CustomException = ThisException | ThatException deriving Show
+instance X.Exception CustomException
```
isovector: You shouldn't need this instance.

README.md
```diff
@@ -61,39 +61,136 @@ Make sure you read the [Necessary Language
...
+makeSem ''Teletype
 
-runConsoleIO :: Member (Lift IO) r => Sem (Console ': r) a -> Sem r a
-runConsoleIO = interpret $ \case
+runTeletypeIO :: Member (Lift IO) r => Sem (Teletype ': r) a -> Sem r a
+runTeletypeIO = interpret $ \case
   ReadTTY      -> sendM getLine
   WriteTTY msg -> sendM $ putStrLn msg
+
+runTeletypeBoring :: [String] -> Sem (Teletype ': r) a -> Sem r ([String], a)
```
isovector: Why is this "boring"?

README.md
```diff
@@ -61,39 +61,136 @@ Make sure you read the [Necessary Language
...
+  . runListInput i
+  . reinterpret2 \case
+      ReadTTY -> maybe "" id <$> input
+      WriteTTY msg -> output msg
+
+-- Continuations so we can build up chains of behavior
+echoK :: Member Teletype r => Sem r () -> Sem r ()
+echoK k = do
+  a <- readTTY
+  when (a == "") $ pure ()
```
isovector: `case a of "" -> ...; _ -> ...`

README.md
```diff
@@ -61,39 +61,136 @@ Make sure you read the [Necessary Language
...
+  when (a /= "") do
+    writeTTY (reverse a)
+    k
+
+-- Fixpoints
+
+echo :: Member Teletype r => Sem r ()
+echo = fix echoK
```
isovector: wat?

README.md
```diff
@@ -61,39 +61,136 @@ Make sure you read the [Necessary Language
...
+
+revIO :: Sem '[Lift IO] ()
+revIO = runTeletypeIO rev
+
+alternatingIO :: Sem '[Lift IO] ()
+alternatingIO = runTeletypeIO alternating
+
+-- cycle echo and reverse until a blank line
+main :: IO ()
+main = runM alternatingIO
```
isovector: This example seems pretty complicated to me. I'm not sure what's supposed to happen, nor why there are `fix`es. Maybe there's a simpler demonstration where we could just give a pure and an IO interpretation of `Teletype`, write a quick echo server, and then evaluate it both ways?

README.md
```diff
@@ -61,39 +61,136 @@ Make sure you read the [Necessary Language
...
+{-# LANGUAGE BlockArguments, DataKinds, FlexibleContexts, GADTs, LambdaCase,
+             PolyKinds, RankNTypes, ScopedTypeVariables, TypeApplications,
+             TypeOperators, TypeFamilies, UnicodeSyntax
+#-}
 
+import Prelude hiding (throw, catch, bracket)
+import Polysemy
+import Polysemy.Input
+import Polysemy.Output
+import Polysemy.Error
```
isovector: Might as well just import `Resource` here too rather than redefining it, no?

README.md
```diff
@@ -61,39 +61,136 @@ Make sure you read the [Necessary Language
...
+runTeletypeIO :: Member (Lift IO) r => Sem (Teletype ': r) a -> Sem r a
+runTeletypeIO = interpret $ \case
+  ReadTTY      -> sendM getLine
+  WriteTTY msg -> sendM $ putStrLn msg
+
+runTeletypeBoring :: [String] -> Sem (Teletype ': r) a -> Sem r ([String], a)
+runTeletypeBoring i
+  = runFoldMapOutput pure
+  . runListInput i
+  . reinterpret2 \case
```
isovector: I really like this, but a comment or two would probably go a long way.

README.md
```diff
@@ -110,6 +207,27 @@ runResource finish = interpretH $ \case
...
+          writeTTY "no exceptions"
+
+boringProgram :: IO (Either CustomException ([String], ()))
+boringProgram = (runM .@ runResource) . runError @CustomException . runTeletypeBoring ["words"] $ program
+
+boringExplosion :: IO (Either CustomException ([String], ()))
+boringExplosion = (runM .@ runResource) . runError @CustomException . runTeletypeBoring ["explode"] $ program
+
+ioProgram :: IO (Either CustomException ())
+ioProgram = (runM .@ runResource .@@ runErrorInIO @CustomException) . runTeletypeIO $ program
```
isovector: What do you think about moving these into a `docs/` page, with more prose/explanation power, + a link from the readme?

jhenahan:
Made things much less weird!

isovector:
Good enough for me! Thanks for putting this together!

PR #24, opened 2019-04-25
Add Members and (>@>) type synonyms

TheMatten:
When having many effects in one function, type signatures tend to get hard to read. I am proposing adding two new type synonyms - `Members`, which takes list of effects instead of one and `(>@>)` (_at-fish_ or whatever you want to call it :smile:), which constructs type from list of member effects and return type:
```haskell
foo :: ( Member (Output Int)  r
       , Member (Output Bool) r
       , Member Console       r
       )
    => Sem r ()
```
could be written simply as:
```haskell
foo :: '[Output Int, Output Bool, Console] >@> ()
```

There are few things I am not completely sure about:
- [x] (https://github.com/isovector/polysemy/pull/24#issuecomment-487040049) Should `Members` be even available, or is `(>@>)` sufficient enough for all purposes?
- [ ] If yes, should `r` in `Members es r` in  appear before list to make constraint more readable on multiple lines?
- [x] (https://github.com/isovector/polysemy/pull/24#issuecomment-487040049) Can rank-2 `r` create some complications in specific cases?
- [x] (https://github.com/isovector/polysemy/pull/24#issuecomment-487040049) What should operator look like? `(>@>)` does not pop up in Stackage search (and I personally like it :smile:). Should it actually be an operator?
- [ ] Should types of both synonyms be restricted somehow?

### DIFF
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -7,7 +7,6 @@ module Polysemy.Internal
   ( Sem (..)
   , Member
   , Members
-  , type (>@>)
   , send
   , sendM
   , run
@@ -154,28 +153,6 @@ newtype Sem r a = Sem
 type family Members es r :: Constraint where
   Members '[]       r = ()
   Members (e ': es) r = (Member e r, Members es r)
-
-
-------------------------------------------------------------------------------
--- | Compact alternative to 'Sem' type with 'Member' effects:
---
--- @
--- foo :: \'['Polysemy.Output.Output' Int, 'Polysemy.Output.Output' Bool] >@> ()
--- @
---
--- translates into:
---
--- @
--- foo :: ( 'Member' ('Polysemy.Output.Output' Int) r
---        , 'Member' ('Polysemy.Output.Output' Bool) r
---        )
---     => 'Sem' r ()
--- @
---
--- (NB: keep in mind that 'r' variable representing list of all effects is
--- hidden behind rank-2 type --- use 'Members' instead if you need access to it.)
-infixr 1 >@>
-type es >@> a = ∀ r. Members r es => Sem r a
 
 
 ------------------------------------------------------------------------------
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -174,7 +174,7 @@ type family Members es r :: Constraint where
 --
 -- (NB: keep in mind that 'r' variable representing list of all effects is
 -- hidden behind rank-2 type --- use 'Members' instead if you need access to it.)
-infix 0 >@>
+infixl 1 >@>
 type es >@> a = ∀ r. Members r es => Sem r a
 
 
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -138,6 +138,7 @@ newtype Sem r a = Sem
 -- @
 -- foo :: 'Members' \'[ 'Polysemy.Output.Output' Int
 --                 , 'Polysemy.Output.Output' Bool
+--                 , 'Polysemy.State' String
 --                 ] r
 --     => 'Sem' r ()
 -- @
@@ -147,6 +148,7 @@ newtype Sem r a = Sem
 -- @
 -- foo :: ( 'Member' ('Polysemy.Output.Output' Int) r
 --        , 'Member' ('Polysemy.Output.Output' Bool) r
+--        , 'Member' ('Polysemy.State' String) r
 --        )
 --     => 'Sem' r ()
 -- @
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -6,6 +6,8 @@
 module Polysemy.Internal
   ( Sem (..)
   , Member
+  , Members
+  , type (>@>)
   , send
   , sendM
   , run
@@ -23,6 +25,7 @@ import Control.Applicative
 import Control.Monad.Fix
 import Control.Monad.IO.Class
 import Data.Functor.Identity
+import Data.Kind
 import Polysemy.Internal.Effect
 import Polysemy.Internal.Fixpoint
 import Polysemy.Internal.Lift
@@ -127,6 +130,52 @@ newtype Sem r a = Sem
         => (∀ x. Union r (Sem r) x -> m x)
         -> m a
   }
+
+
+------------------------------------------------------------------------------
+-- | Makes constraints of functions that use multiple effects more readable by
+-- translating single list of effects into multiple 'Member' constraints:
+--
+-- @
+-- foo :: 'Members' \'[ 'Polysemy.Output.Output' Int
+--                 , 'Polysemy.Output.Output' Bool
+--                 ] r
+--     => 'Sem' r ()
+-- @
+--
+-- translates into:
+--
+-- @
+-- foo :: ( 'Member' ('Polysemy.Output.Output' Int) r
+--        , 'Member' ('Polysemy.Output.Output' Bool) r
+--        )
+--     => 'Sem' r ()
+-- @
+type family Members es r :: Constraint where
+  Members '[]       r = ()
+  Members (e ': es) r = (Member e r, Members r es)
+
+
+------------------------------------------------------------------------------
+-- | Compact alternative to 'Sem' type with 'Member' effects:
+--
+-- @
+-- foo :: \'['Polysemy.Output.Output' Int, 'Polysemy.Output.Output' Bool] >@> ()
+-- @
+--
+-- translates into:
+--
+-- @
+-- foo :: ( 'Member' ('Polysemy.Output.Output' Int) r
+--        , 'Member' ('Polysemy.Output.Output' Bool) r
+--        )
+--     => 'Sem' r ()
+-- @
+--
+-- (NB: keep in mind that 'r' variable representing list of all effects is
+-- hidden behind rank-2 type --- use 'Members' instead if you need access to it.)
+infix 0 >@>
+type es >@> a = ∀ r. Members r es => Sem r a
 
 
 ------------------------------------------------------------------------------
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -174,7 +174,7 @@ type family Members es r :: Constraint where
 --
 -- (NB: keep in mind that 'r' variable representing list of all effects is
 -- hidden behind rank-2 type --- use 'Members' instead if you need access to it.)
-infixl 1 >@>
+infixr 1 >@>
 type es >@> a = ∀ r. Members r es => Sem r a
 
 
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -153,7 +153,7 @@ newtype Sem r a = Sem
 -- @
 type family Members es r :: Constraint where
   Members '[]       r = ()
-  Members (e ': es) r = (Member e r, Members r es)
+  Members (e ': es) r = (Member e r, Members es r)
 
 
 ------------------------------------------------------------------------------
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -3,7 +3,6 @@ module Polysemy
     Sem ()
   , Member
   , Members
-  , type (>@>)
 
   -- * Running Sem
   , run
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -2,6 +2,8 @@ module Polysemy
   ( -- * Core Types
     Sem ()
   , Member
+  , Members
+  , type (>@>)
 
   -- * Running Sem
   , run


src/Polysemy/Internal.hs
```diff
@@ -129,6 +132,52 @@ newtype Sem r a = Sem
...
+--
+-- @
+-- foo :: ( 'Member' ('Polysemy.Output.Output' Int) r
+--        , 'Member' ('Polysemy.Output.Output' Bool) r
+--        )
+--     => 'Sem' r ()
+-- @
+type family Members es r :: Constraint where
+  Members '[]       r = ()
+  Members (e ': es) r = (Member e r, Members r es)
```
isovector: This should be `Members es r`, no?

TheMatten:
Nice :smile:
Yeah, I mostly wanted to make types more clean, but this is good point too :smiley:
Will look at it, but it's 10 pm here, so let's wait until tomorrow.

src/Polysemy/Internal.hs
TheMatten: Oops, thanks! I should probably test revised version before pushing...

TheMatten:
> I'd like a story around how this interacts with interpreters before merging it, because there we do want to make sure the `r`s are the same! Consider:
> 
> ```haskell
> runTraceIO :: Member (Lift IO) r => Sem (Trace ': r) a -> Sem r a
> ```
> 
> Let's assume there are more member constraints here --- how can `>@>` help us?

Well, basically it seems like there are two different use-cases for `Sem`:
1. We are users of it, we only care about member effects and result type (and nothing else?)
2. We are interpreting some effects and we need access to both the members we want to translate actions into and list itself to match on effects we want to remove or add (this does not seem to even possibly work with `(>@>)`, except for `intercept` maybe).

Now, 2. works relatively nicely with `Members`:
```haskell
runTraceIO :: Members '[Input String, Output Int, Lift IO] r
           => Sem (Trace ': r) a -> Sem r a
```
We _could_ create synonym for this workflow with `r` (and maybe even `a`) hidden, e.g.:
```haskell
runTraceIO :: RunWith '[Input String, Output Int, Lift IO] Trace

type RunWith ms e = forall r a. Members ms r => Sem (e ': r) a -> Sem r a
```
But IMHO this would be totally unreadable to newcomers...
It may be better with list transformations being explicit (and this way we would support transformations on mutiple effects at once, if this is something we may want):
```haskell
runTraceIO :: MapWith '[Input String, Output Int, Lift IO] (Trace ': r) r

type MapWith ms r r' = forall a. Members ms r' => Sem r a -> Sem r' a
```
`(>@>)` describes one concrete concept of adding effects to context of some value (From the user perspective, I like it, because it seems to capture exactly the goal of effects and nothing else :smiley:).
But for the other side, it may be beneficial to actually see what's going on; plus, I am not completely sure about how would `Tactics` compose with these other combinators...

isovector:
I do like the ergonomics of `>@>` for writing application code, but yeah, I am worried about the "WTF is going on here" factor. What used to be just one concept _use `Member` to constraint `Sem`_ becomes:

* Application code gets written with `>@>`
* Interpreters still use `Member` to constrain `Sem`
* Effect actions are somewhere in between? What type does `catch` have? `'[Error e] >@> a -> (e -> '[Error e] >@> a) -> '[Error e] >@> a`? Yikes!

I'm leaning towards just making this PR be `Members` and relegating `>@>` to be defined by someone on their own, outside of the main library. What do you think?

TheMatten:
Seems reasonable, ok.

I will look into this - if I find some way to integrate all the aspects of the library into similar solution in some way, should I make a new PR?

At the end, we will probably need to "pipe" that list through somehow - theoretically, it could be possible by defining it in some top level part of the signature and then passing it to the other combinators through additional parameter... (We don't have currying at the type level, but there are some ways... :smile:)

src/Polysemy/Internal.hs
```diff
@@ -129,6 +131,30 @@ newtype Sem r a = Sem
...
 
 
+------------------------------------------------------------------------------
+-- | Makes constraints of functions that use multiple effects more readable by
+-- translating single list of effects into multiple 'Member' constraints:
+--
+-- @
+-- foo :: 'Members' \'[ 'Polysemy.Output.Output' Int
+--                 , 'Polysemy.Output.Output' Bool
+--                 ] r
```
isovector: This might be a little clearer if it uses an effect that isn't the same between each case. Maybe two `Output`s and a `State`?

TheMatten: Done.

isovector:
Merged. Thank you so much! Feel free to open another PR if you come up with a nice story for `>@>` -- and I'll keep it in mind for the next time I write an app with this stuff :)

TheMatten:
Nice! :smile:
Will do - I'm already thinking about one way of doing this, hopefully it will work with some tinkering :smiley:

jhenahan:
@TheMatten 

```haskell
type family Curry (xs :: [Type]) (y :: Type) where
  Curry '[] y = y
  Curry (x ': xs) y = x -> Curry xs y
```

Something like that?

isovector:
I'm not sure what y'all want currying for, but if you really and truly need it, maybe it's worth waiting for [this](https://www.microsoft.com/en-us/research/uploads/prod/2019/03/ho-haskell.pdf) to land.

adamConnerSax:
Oh! ```Curry``` turns the list of effects into a signature that you can use to handle one.  Or all of them.? Would you also need ```Uncurry``` so you could handle just one?

TheMatten:
@jhenahan I meant what @isovector's  [link](https://www.microsoft.com/en-us/research/uploads/prod/2019/03/ho-haskell.pdf) describes - with higher-order type functions, I could pass hidden argument (our list) to unsaturated types.

TheMatten:
This compiles and runs :smile: (I did not test this thoroughly, but it works with simple program with home-made effects and typechecks with e.g. runTraceIO signature):

```haskell
foo :: Member (Lift IO) r => '[RandomInt, ConsoleIO] >@r@> ()
```

It's actually a simple trick:

```haskell
infix 2 @>
type (@>) = Semantic

type family SemList s where SemList (Semantic r _) = r

infix 1 >@
type (>@) es s = Members es (SemList s) => s
```

So we could have variant with hidden and available list variable, depending on what we need...

isovector:
This is disgusting and I love it. Can you show me what `runTraceIO` and `runResource` look like under this scheme?

TheMatten:
> This is disgusting and I love it.

:smile:

We could write `runTraceIO` like this:

```haskell
runTraceIO :: Trace ': r @> a -> '[Lift IO] >@ r @> a
```

but in this case, again, it may be more readable with `Member` (let's just write cons without quote, you can't define type level construct with same name anyway from what I tried):

```haskell
runTraceIO :: Member (Lift IO) r => Trace:r @> a -> r @> a
```

Both types evaluate into same result as the original:

```haskell
runTraceIO :: Member (Lift IO) r => Sem (Trace : r) a -> Sem r a
```
Same for `runResource`:

```haskell
runResource :: (forall x. r @> x -> IO x) -> Resource:r @> a -> '[Lift IO] >@ r @> a

runResource :: Member (Lift IO) r => (forall x. r @> x -> IO x) -> Resource:r @> a -> r @> a

runResource :: Member (Lift IO) r => (forall x. Sem r x -> IO x) -> Sem (Resource : r) a -> Sem r a
```

`>@ r @>` really feels like an extension to `>@>` while it technically preserves "shape" of signature comparing to `MapWith`/`RunWith`, and it should work with `Tactics` too.

Again, we can question if it's readable enough for newcomers and isn't just an example of "operator syndrome" some Haskellers (well, including me) may have. But personally, I would love to have it in `polysemy` - this way I would write most of actions without ever mentioning `r` and have consistent alternative for interpreters with explicit lists too. And sufficient documentation should be enough for people coming to this library.

Some points:
- `('[Lift IO] >@)` on every action should work too, but it creates separate constraints for all different versions of lists, plus constraint on `Resource:r` _only_ won't work - it's not the same list as `r` for GHC apparently...
- how should `>@ r @>`/`>@r@>` be written "officially"?
- Hmm, should `Member`/`Members` get their own operators too then?... :smiley:

TheMatten:
(If this won't get into library itself, I will probably create `polysemy-aquarium` :smile:)

isovector:
Sorry to bikeshed, but my eyes glaze over at so many `>`s, and I find it hard to quickly and visually distinguish `@>` from `->`. 

I wonder if we can do something with sufficiently clever polykinded type families here. Maybe like

```
type family Z (e :: k) r :: Type where
  Z (e :: (Type -> Type) -> Type -> Type) r    = Member e r
  Z (es :: [(Type -> Type) -> Type -> Type]) r = Members es r
```

(for some bikeshedded name/symbol `Z`)

This would get rid of the necessary brackets around `Lift` -- and if we're feeling cute, could add a `Z (e :: Type -> Type) r = Member (Lift e) r` case too

TheMatten:
Ok, will look into possible alternatives for current symbol.
May be good idea, ok.
Aaah, those timezones :smile: - it's late here, will look at it tomorrow.

TheMatten:
What about this?:
```haskell
foo         :: '[RandomInt, ConsoleIO] @|@ ()
runTraceIO  :: Trace:r|@ a -> '[Lift IO] @|r|@ a
runResource :: (forall x. r|@ x -> IO x) -> Resource:r|@ a -> '[Lift IO] @|r|@ a
```

But I can't get "universal" members constraint to work properly - problem can be illustrated on modified `Members`:

```haskell
type family Members es r :: Constraint where
  Members '[]       r = ()
  Members (e ': es) r = (Member e r, Members es r)
```

works fine, but if we make `es` kind polymorphic (what we need because of other cases):

```haskell
type family Members (es :: k) r :: Constraint where
  Members '[]       r = ()
  Members (e ': es) r = (Member e r, Members es r)
```

GHC fails with terrible error:

```
Expected kind ‘k’, but ‘e : es’ has kind ‘[(* -> *) -> * -> *]’
```

This can be fixed by specifying kind of `r` (I would like to know why... :smile:):

```haskell
type family Members (es :: k) (r :: k') :: Constraint where
  Members '[]       r = ()
  Members (e ': es) r = (Member e r, Members es r)
```

But now, with `k` and `k'`, `Members` doesn't reduce...:

```haskell
λ :k! forall r. Members '[RandomInt, ConsoleIO] r
forall r. Members '[RandomInt, ConsoleIO] r :: Constraint
= Members '[RandomInt, ConsoleIO] r
```

which means that signatures using it won't do their work:

```haskell
foo :: '[RandomInt, ConsoleIO] @|@ ()
```

```
Could not deduce: Members '[RandomInt, ConsoleIO] r
from the context: Members '[RandomInt, ConsoleIO] r
  bound by the type signature for:
             foo :: forall k. '[RandomInt, ConsoleIO] @|@ ()
```

_except_... when we concretize `k'` and specify some `Lift`-ed `Monad` as `Member`...:

```haskell
type family Members (es :: k) (r :: [(Type -> Type) -> Type -> Type]) :: Constraint where
  Members '[]       r = ()
  Members (e ': es) r = (Member e r, Members es r)
```

```haskell
λ :k! forall r. Members '[RandomInt, ConsoleIO, Lift IO] r
forall r. Members '[RandomInt, ConsoleIO, Lift IO] r :: Constraint
= ((Polysemy.Internal.Union.Find r RandomInt, ...
```

Hmm...

Same happens with separate family that uses original `Members` internally.

TheMatten:
At the end, it seems to be problem with inference of kind of effects in list - there may be way of doing this, but I don't think it is worth it - simple list is relatively fine anyway.

isovector:
Sorry, why is `es` kind-polymorphic here?

TheMatten:
Sorry, I didn't rename that "universal" constraint to different name -  well, it is basically just `Members` with polymorphic argument and added patterns (but of course, list case could be directed to `Members` as well) - so, that `Members` here is actually `Z` with list only - I stopped there, because it already doesn't reduce (because of ambiguous types in effects arguments from my testing).

TheMatten:
Potential complete version (if we get it to infer types properly) named `Effs`, because I don't have any other ideas for a name...:
```
type family Effs (x :: k) (r :: k') :: Constraint where
  -- List of member effects
  Effs (es :: [(Type -> Type) -> Type -> Type]) r = Members es r
  -- Single effect - allows to omit brackets in '(@|)'
  Effs (e  :: (Type -> Type) -> Type -> Type  ) r = Member e r
  -- -- Single monad - allows to omit 'Lift'
  Effs (m  :: Type -> Type                    ) r = Member (Lift m) r
```

TheMatten:
Really, I'm not sure if it's worth additional type complexity, comparing to those additional brackets and `Lift`... Even if it will work, how will errors look like? Won't it confuse people about differences between lifted monads and effects?

TheMatten:
Okay, so assuming effects would get concrete types of monad and return value, it will work - I will prepare implementation with good errors and documentation.

PR #37, opened 2019-05-05
Replace Typeable with SNat equality

googleson78:
I am not sure if it "erases" as well as the `Typeable` version, because looking at it:
`prj` is used only in `interceptH`
`interceptH` is used only in `intercept` (and exported)
`intercept` is never never used internally, only exported, i.e. none of this is tested.

I would be happy to write some tests, if you have any suggestions/guidelines. (this is the `WIP` part of the PR)

The only thing that is getting called is `weaken`, which is the same as before, except freed from having to provide a `Typeable` context.

### DIFF
--- a/src/Polysemy/Internal/Union.hs
+++ b/src/Polysemy/Internal/Union.hs
@@ -31,7 +31,7 @@ module Polysemy.Internal.Union
 
 import Data.Functor.Compose
 import Data.Functor.Identity
-import Data.Typeable
+import Data.Type.Equality
 import Polysemy.Internal.Effect
 
 #ifdef ERROR_MESSAGES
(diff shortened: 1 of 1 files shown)


src/Polysemy/Internal/Union.hs
```diff
@@ -113,27 +113,26 @@ type Member' e r =
...
 ------------------------------------------------------------------------------
 -- | A singleton for 'Nat'.
 data SNat :: Nat -> * where
   SZ :: SNat 'Z
-  SS :: Typeable n => SNat n -> SNat ('S n)
-  deriving Typeable
+  SS :: SNat n -> SNat ('S n)
+
+instance TestEquality SNat where
+  testEquality SZ     SZ     = (Just Refl)
```
isovector: No parens necessary

isovector:
This is fantastic, thanks! The relevant test is [here](https://github.com/isovector/polysemy/blob/master/test/FusionSpec.hs#L33) so it seems like this doesn't break any fusion. I'm happy to merge when you give me the :+1:

googleson78:
I pushed a ""test"" to see if I can get something to break.
Honestly I'm not sure if this will even reach `prj` at all, I just filled in type holes until it compiled+passed, using examples from the rest of the library. If it's useless I'll just remove it.

isovector:
`prj` is only called by `intercept`, so you could test it with something like this:

```haskell
teeOutput :: (Member (Output String) r, Member (Lift IO) r) => Sem r ()
teeOutput = intercept $ \case
  Output msg -> sendM $ putStrLn msg
```

and then maybe

```haskell
foo :: IO ([String], ())
foo = runM . runOutputFoldMap pure . teeOutput $ do
  output "hello"
  output "world"
```

isovector:
Just looked at your test failure; I think the same program would fail with the original `Typeable` constraint too, at least until [!668](https://gitlab.haskell.org/ghc/ghc/merge_requests/668/) gets merged.

googleson78:
I tried it out with master - it does indeed fail.

I'm removing the test then - will hold it off for eventually putting it in after that patch is merged.

Don't think I have anything else I wanted to try.

isovector:
Love it, thanks!

PR #38, opened 2019-05-05
Support versions of GHC without QuantifiedConstraints

jkachmar:
Addresses #29 by implementing the [workaround used by `fused-effects`](https://github.com/fused-effects/fused-effects/blob/4d0e056c95e4016d6c8d8e14909702874e17863a/src/Control/Effect/Carrier.hs#L16) that @patrickt suggested.

### DIFF
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -40,7 +40,7 @@ swap ~(a, b) = (b, a)
 -- | The simplest way to produce an effect handler. Interprets an effect @e@ by
 -- transforming it into other effects inside of @r@.
 interpret
-    :: FirstOrder e m0 "interpret"
+    :: FirstOrder m0 e "interpret"
     => (∀ x m. e m x -> Sem r x)
        -- ^ A natural transformation from the handled effect to other effects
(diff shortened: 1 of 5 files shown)


isovector:
Looks great! Just to confirm, the new tyvars in `FirstOrder` don't cause the error messages from coming up, do they?

jkachmar:
It looks like the error messages still come through just fine:

```haskell
runResource
    :: forall r a
     . Member (Lift IO) r
    => (∀ x. Sem r x -> IO x)
       -- ^ Strategy for lowering a 'Sem' action down to 'IO'. This is likely
       -- some combination of 'runM' and other interpreters composed via '.@'.
    -> Sem (Resource ': r) a
    -> Sem r a
runResource finish = interpret undefined
```
...results in the following error message:
```
    /Users/jkachmar/src/polysemy/src/Polysemy/Resource.hs:66:23: error:
        • 'Resource' is higher-order, but 'interpret' can help only
          with first-order effects.
          Fix:
            use 'interpretH' instead.
        • In the expression: interpret undefined
          In an equation for ‘runResource’:
              runResource finish = interpret undefined
       |
    66 | runResource finish = interpret undefined
       |     
```

PR #39, opened 2019-05-05
Expand CI coverage

isovector:
Now that #29 is finished, let's see which versions of GHC this thing compiles with.


jkachmar:
Is there anything I can do to help with this? Looks like it's diverged a bit since my last PR.

isovector:
@jkachmar we're waiting on getting #67 merged first, which should solve all of the TH woes. Thanks for staying keen about it though!

src/Polysemy/Writer.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE BlockArguments  #-}
+{-# LANGUAGE CPP             #-}
```
isovector: why

isovector:
okay! so core polysemy works! But i'm not sure how to get cabal to build polysemy-plugin against the polysemy that exists in the repo

jkachmar:
> But i'm not sure how to get cabal to build `polysemy-plugin` against the `polysemy` that exists in the repo

#123 should address this; you can tell `cabal-install` about projects that are in the same directory by setting a `cabal.project` file with some of this information called out there.

ocharles:
"Mute this thread"

jhenahan:
This is my favorite repo on all of GitHub.

PR #40, opened 2019-05-06
Removes BlockArguments (pre-GHC 8.6 compat)

jkachmar:
Travis was failing on versions of GHC pre-8.6 because of the `BlockArguments` language extension.

Removing it makes the code inarguably less cool to read, but it should be fix the compatibility issue so maybe that's okay.


jkachmar:
PR's against the `more-travis` branch so we'll see if this actually fixes anything.

jkachmar:
Closed in favor of #41

Issue #27, opened 2019-04-28
Documentation: Make sure to add `polysemy-plugin` to your package.yaml


isovector:
The documentation on using `polysemy-plugin` doesn't mention it needs to exist in your package yaml dependencies.

PR #43, opened 2019-05-14
polysemy-plugin doc

csasarak:
This is meant to address issue #27 

In addition to documenting where to add polysemy-plugin I also changed the second example to not require it. The motivation being that documentation for polysemy-plugin can be in one place and there'd be no need to add a caveat near the resource-effect example that polysemy-plugin must be installed.


isovector:
Fantastic, thanks!

Issue #57, opened 2019-05-22
Higher order effects bug


Simre1:
There seems to be a bug with higher order effects. Here is an example with the current version of polysemy on hackage (0.1.2.1) and ghc-8.6.4.

```
main :: IO ()
main = runM . runReader 4 $ test

test :: Sem '[Reader Int, Lift IO] ()
test = do
  ask @Int >>= liftIO . print
  local (+(4 :: Int)) $ do
    ask @Int >>= liftIO . print
```
Executing the main function, I would expect:
`4`
`8`

What's actually happening is:
`4`
`4`

isovector:
What an embarrassing bug! I can duplicate it, but not entirely sure _why_. I'll dig in tomorrow!

mitchellwrosen:
Randomly chiming in here, do not be embarrassed, software is reeeeeally hard... ;)

isovector:
Thanks for finding this --- it's a pretty duplicitous bug in the fundamental combinators. Essentially they'd handle the higher-order effects automatically using the currently-run interpreter, but then would call `lift` on it so it seemed like you still had to handle that thing. The behavior "works" for any interpreter that just wants to call itself, but `runReader` isn't in that class.

isovector:
I've released the fix as version 0.2.0.0, along with deprecating the other versions on hackage. Thanks again for the bug report!

Simre1:
Thanks for fixing the problem so quickly! It works perfectly now.

PR #58, opened 2019-05-23
Fix broken higher order effects

isovector:
#57 pointed out that `local` doesn't work in `Reader`. Some investigation showed that `runReader` was not to blame. Uh oh!

The bug was something stupid; when I introduced the `Tactics` machinery I didn't retest my original interpreters; and instead was relying on the typesystem. But the implementation of `interpretH` and friends would hard-code the current interpreter into the natural transformation in `Tactics`, which meant that the signature for `runT` et al was lying. They promised to give you some effects you could handle, but in fact _they were already handled!_

I've fixed the bug and added a small test case to prove `local` now works.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -50,7 +50,7 @@ flags:
     manual:      True
 
 library:
-  ghc-options: -O2 -Wall
+  ghc-options: -Wall
   source-dirs: src
   other-modules: Polysemy.Internal.PluginLookup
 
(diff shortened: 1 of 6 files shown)


isovector:
I've released the fix as version `0.2.0.0`, along with deprecating the other versions on hackage. Thanks again for the bug report!

PR #59, opened 2019-05-23
Remove erroneous Prelude import

jhenahan:
I left that in when writing the README example because I use a custom Prelude. Whoops.


isovector:
Thanks!

PR #62, opened 2019-05-23
Fix the plugin

### LINKED ISSUE #61: polysemy-plugin-0.1.0.1 doesn't work
#46 points out that the example code doesn't work when built with stack. The stackage guys are mad at me because it's broken there too (https://github.com/commercialhaskell/stackage/issues/4572)

In my tests, downgrading `polysemy-plugin` to `0.1.0.0` seems to fix the problem.

isovector:
Fixes #61; see the discussion there for what went wrong.


isovector:
Confirmed to fix https://github.com/commercialhaskell/stackage/issues/4572

Issue #56, opened 2019-05-22
Rewrite Polysemy.Internal.TH.Effect in terms of th-abstraction


isovector:
The [`th-abstraction`](http://hackage.haskell.org/package/th-abstraction-0.3.1.0/docs/Language-Haskell-TH-Datatype.html) package gives a significantly nicer view of GADTs. This would make it much easier to fix #48 and any other hidden TH bugs that nobody has run into yet.

googleson78:
I'll try looking at this in the weekend if someone isn't already and it doesn't need to be done asap.

isovector:
I think @TheMatten is currently tackling this!

TheMatten:
Should have written here :smile:

Will make PR when I will have something usable.

isovector:
Fixed by #67

Issue #48, opened 2019-05-15
TH generates ambiguous smart constructors for phantom parameters


isovector:
The TH currently spits out smart constructors like this:

```haskell
data Foo e m a where
  FooC :: Foo e m ()

fooC :: forall e r. Member (Foo e) r => Sem r ()
fooC = send FooC
```

Uh oh! The usage of `FooC` here is ambiguous! The plugin will correctly determine what it should be (`send $ FooC @e`), but it'd be nicer if the TH just did the right thing to begin with.

We want to generate this:

```haskell
fooC :: forall e r. Member (Foo e) r => Sem r ()
fooC = send (FooC @e)
```

googleson78:
```haskell
fooC :: forall e r. Member (Foo e) r => Sem r ()
fooC = send (FooC @e)
```
is also ambiguous? (according to `ghc`)

isovector:
It is, but differently ambiguous! The code as its written today is ambiguous in the effect its trying to send, but the correct code is only ambiguous in terms of `AllowAmbiguousTypes` :)

isovector:
Fixed by #67

PR #67, opened 2019-05-27
Rewrite 'Polysemy.Internal.TH.Effect' in terms of 'th-abstraction'

TheMatten:
This PR should resolve (at least) #56, #48 and add support for other variants of datatype declaration (even data families maybe?). See TODOs in code for things that need to be resolved.

### DIFF
--- a/src/Polysemy/Internal/TH/Effect.hs
+++ b/src/Polysemy/Internal/TH/Effect.hs
@@ -1,9 +1,7 @@
 {-# OPTIONS_HADDOCK not-home #-}
 
-{-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE TemplateHaskell     #-}
 {-# LANGUAGE TupleSections       #-}
-{-# LANGUAGE MonadComprehensions #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 
 -- | This module provides Template Haskell functions for automatically generating
@@ -37,6 +35,8 @@ import Prelude hiding ((<>))
 import Control.Monad
 import Data.Bifunctor
 import Data.Char                      (toLower)
+import Data.Either
+import Data.Generics
 import Data.List
 import Language.Haskell.TH
 import Language.Haskell.TH.PprLib
@@ -47,6 +47,7 @@ import Polysemy.Internal.CustomErrors (DefiningModule)
 import qualified Data.Map.Strict as M
 
 -- TODO: update documentation based on other decisions
+-- TODO: write tests for what should (not) compile
 
 ------------------------------------------------------------------------------
 -- | If @T@ is a GADT representing an effect algebra, as described in the
@@ -83,151 +84,171 @@ makeSem_ = genFreer False
 -- | Generates declarations and possibly signatures for functions to lift GADT
 -- constructors into 'Sem' actions.
 genFreer :: Bool -> Name -> Q [Dec]
-genFreer shouldMkSigs typeName = do
-  dtInfo <- reifyDatatype typeName
-  -- TODO: data families seem to work just fine - should they be restricted?
-  clInfos  <- traverse (mkCLInfo dtInfo) $ datatypeCons dtInfo
-  defModFI <- TySynInstD ''DefiningModule
-          .   TySynEqn [ConT $ datatypeName dtInfo]
-          .   LitT
-          .   StrTyLit
-          .   loc_module
-          <$> location
-  sigs     <- if shouldMkSigs then traverse genSig clInfos
-                              else return []
-  decs     <- traverse genDec clInfos
+genFreer should_mk_sigs type_name = do
+  dt_info       <- reifyDatatype type_name
+  cl_infos      <- traverse (mkCLInfo dt_info) $ datatypeCons dt_info
+  tyfams_ext_on <- isExtEnabled TypeFamilies
+  def_mod_fi    <- sequence [ TySynInstD ''DefiningModule
+                                . TySynEqn [ConT $ datatypeName dt_info]
+                                . LitT
+                                . StrTyLit
+                                . loc_module
+                              <$> location
+                            | tyfams_ext_on
+                            ]
+  sigs          <- if should_mk_sigs
+                      then traverse genSig cl_infos
+                      else return []
+  decs          <- traverse genDec cl_infos
 
-  return $ defModFI : join (sigs:decs)
+  return $ join $ def_mod_fi : sigs : decs
+
+------------------------------------------------------------------------------
+-- | Generates signature for lifting function and type arguments to apply in
+-- it's body on effect's data constructor.
+genSig :: ConLiftInfo -> Q Dec
+genSig
+  (CLInfo
+    { cliTypeName
+    , cliFunName
+    , cliEffTArgs
+    , cliMName
+    , cliResArg
+    , cliConCxt
+    , cliConArgs
+    })
+  = do
+  r <- newName "r"
+  let member_cxt = ConT ''Member `AppT` eff `AppT` VarT r
+      eff        = foldl' AppT (ConT cliTypeName) cliEffTArgs
+      -- TODO: can use of 'm' blow up somehow?
+      f_args     = replaceMArg cliMName r cliConArgs
+      return_t   = ConT ''Sem `AppT` VarT r `AppT` cliResArg
+
+  return $ SigD cliFunName $ quantifyType $
+    ForallT [] (member_cxt : cliConCxt) $ foldArrows $ f_args ++ [return_t]
+
+------------------------------------------------------------------------------
+-- | Replaces use of @m@ in type with @Sem r@.
+replaceMArg :: TypeSubstitution t => Name -> Name -> t -> t
+replaceMArg m r = applySubstitution $ M.singleton m $ ConT ''Sem `AppT` VarT r
+
+------------------------------------------------------------------------------
+-- | Builds a function definition of the form @x a b c = send $ X a b c@.
+genDec :: ConLiftInfo -> Q [Dec]
+-- TODO: apply effect's type arguments to avoid "bad ambiguity"
+genDec (CLInfo { cliFunName, cliConName, cliConArgs }) = do
+  f_args_names <- replicateM (length cliConArgs) $ newName "x"
+
+  return
+    [ PragmaD $ InlineP cliFunName Inlinable ConLike AllPhases
+    , FunD cliFunName
+        [ Clause (VarP <$> f_args_names)
+                 (NormalB $ AppE (VarE 'send) $
+                   foldl1' AppE $ ConE cliConName : (VarE <$> f_args_names))
+                 []
+        ]
+    ]
 
 -------------------------------------------------------------------------------
 -- | Info about constructor being lifted; use 'mkCLInfo' to create one.
-data ConLiftInfo = CLInfo {
-    typeName :: Name
-  , funName  :: Name
-  , conName  :: Name
-  , effTArgs :: [Type]
-  , mName    :: Name
-  , resArg   :: Type
-  , conCxt   :: Cxt
-  , conArgs  :: [Type]
+data ConLiftInfo = CLInfo
+  { cliTypeName :: Name    -- ^ type constructor of effect
+  , cliFunName  :: Name    -- ^ final lifting function
+  , cliConName  :: Name    -- ^ data constructor being lifted
+  , cliEffTArgs :: [Type]  -- ^ effect specific arguments to type constructor
+  , cliMName    :: Name    -- ^ monad argument in effect
+  , cliResArg   :: Type    -- ^ result argument in effect
+  , cliConCxt   :: Cxt     -- ^ constraints introduced by constructor
+  , cliConArgs  :: [Type]  -- ^ constructor's arguments
   } deriving Show
 
 ------------------------------------------------------------------------------
 -- | Creates info about constructor being lifted from parent type's and it's
 -- own info.
 mkCLInfo :: DatatypeInfo -> ConstructorInfo -> Q ConLiftInfo
--- mkCLInfo dtInfo conInfo = do
-mkCLInfo DatatypeInfo {
-             datatypeName = typeName
-           , datatypeVars
-           , datatypeInstTypes
-           }
-         ConstructorInfo {
-             constructorName = conName
-           , constructorContext
-           , constructorFields
-           }
+mkCLInfo
+  (DatatypeInfo
+    { datatypeName = cliTypeName
+    , datatypeVars
+    , datatypeInstTypes
+    })
+  (ConstructorInfo
+    { constructorName = cliConName
+    , constructorContext
+    , constructorFields
+    })
   = do
-  let funName           = mapNameBase (mapHead toLower) conName
+  let cliFunName            = mapNameBase (mapHead toLower) cliConName
+      all_eff_t_args        = normalizeType <$> datatypeInstTypes
+      cliConArgs            = normalizeType <$> constructorFields
+      normalizeType         = removeTypeKind . applySubstitution eq_pairs
       -- We extract equality constraints with variables to unify them
-      -- manually - this makes type errors more readable
-      (eqPairs, conCxt) = first (M.fromList . maybeAddResKind)
-                        $ partitionWith eqPairOrCxt constructorContext
-      -- We constrain kind of result to 'Type' if we get a type variable
+      -- manually - this makes type errors more readable. Plus we replace
+      -- kind of result with 'Type' if it is a type variable.
+      (eq_pairs, cliConCxt) = first (M.fromList . maybeAddResKind)
+                            $ partitionEithers
+                            $ eqPairOrCxt <$> constructorContext
       -- TODO: we probably want to add requirement for KindSignatures into
       -- documentation - PolyKinds are already mentioned...
-      -- TODO: This makes function signatures ugly - what is the best way to
-      -- restrict occurence of kind signature or completely remove it?
       -- TODO: take result type by name or position?
-      maybeAddResKind   = maybe id ((:) . (, StarT))
-                        $ tvName' $ tvKind $ last datatypeVars
-      allEffTArgs       = applySubstitution eqPairs datatypeInstTypes
-      conArgs           = applySubstitution eqPairs constructorFields
+      maybeAddResKind       = maybe id (\k ps -> (k, StarT) : ps)
+                            $ tVarName $ tvKind $ last datatypeVars
 
-  (effTArgs, [mArg, resArg]) <- case splitFromEnd 2 allEffTArgs of
+  (cliEffTArgs, [m_arg, cliResArg]) <-
+    case splitFromEnd 2 all_eff_t_args of
       r@(_, [_, _]) -> return r
-      _             -> fail $ show $ missingEffTArgs typeName
-  mName <-
-    -- May happen when used on data families
-    maybe (fail $ show $ mArgNotVar typeName mArg) return $ tvName' mArg
+      _             -> fail $ show $ missingEffTArgs cliTypeName
 
-  return CLInfo {
-    typeName, funName, conName, effTArgs, mName, resArg, conCxt, conArgs
-    }
-  where
-    mArgNotVar name mArg
-      =   text "Monad argument ‘" <> ppr mArg <> text "’ in effect ‘"
-          <> ppr name <> text "’ is not a type variable"
+                    -- May happen when used on data families
+  cliMName <- maybe (fail $ show $ mArgNotVar cliTypeName m_arg)
+                    return
+                    (tVarName m_arg)
 
-    missingEffTArgs name
-      =   text "Effect ‘" <> ppr name
-          <> text "’ has not enough type arguments"
+  return CLInfo { cliTypeName
+                , cliFunName
+                , cliConName
+                , cliEffTArgs
+                , cliMName
+                , cliResArg
+                , cliConCxt
+                , cliConArgs
+                }
+
+------------------------------------------------------------------------------
+-- Error messages
+
+mArgNotVar :: Name -> Type -> Doc
+mArgNotVar name mArg
+  =  text "Monad argument ‘" <> ppr mArg <> text "’ in effect ‘"
+  <> ppr name <> text "’ is not a type variable"
+
+missingEffTArgs :: Name -> Doc
+missingEffTArgs name
+  | base <- mapNameBase id name
+  , args <- PlainTV . mkName <$> ["m", "a"]
+
+  =   text "Effect ‘" <> ppr name
+      <> text "’ has not enough type arguments"
+  $+$ nest 4
+      (   text "At least monad and result argument are required, e.g.:"
       $+$ nest 4
-          (   text "At least monad and result argument are required, e.g.:"
-          $+$ nest 4
-              (   text ""
-              $+$ ppr (DataD [] base args Nothing [] []) <+> text "..."
-              $+$ text ""
-              )
+          (   text ""
+          $+$ ppr (DataD [] base args Nothing [] []) <+> text "..."
+          $+$ text ""
           )
-      where
-        base = mapNameBase id name
-        args = PlainTV . mkName <$> ["m", "a"]
+      )
 
 ------------------------------------------------------------------------------
--- | Generates signature for lifting function and type arguments to apply in
--- it's body on effect's data constructor.
-genSig :: ConLiftInfo -> Q Dec
-genSig CLInfo {
-    typeName, funName, effTArgs, mName, resArg, conCxt, conArgs
-  } = do
-  r <- newName "r"
-  let memberCxt  = ConT ''Member `AppT` eff `AppT` VarT r
-      eff        = foldl' AppT (ConT typeName) effTArgs
-      -- TODO: can use of 'm' blow up somehow?
-      funArgs    = replaceMArg mName r conArgs
-      returnType = ConT ''Sem `AppT` VarT r `AppT` resArg
-
-  return $ SigD funName $ quantifyType $
-    ForallT [] (memberCxt : conCxt) $ foldArrows $ funArgs ++ [returnType]
-
-  where
-    -- Replaces use of @m@ in type with @Sem r@.
-    replaceMArg m r = applySubstitution
-                    $ M.singleton m (ConT ''Sem `AppT` VarT r)
-
-------------------------------------------------------------------------------
--- | Builds a function definition of the form @x a b c = send $ X a b c@.
-genDec :: ConLiftInfo -> Q [Dec]
--- TODO: apply effect's type arguments to avoid "bad ambiguity"
-genDec CLInfo { funName, conName, conArgs } = do
-  funArgsNames <- replicateM (length conArgs) $ newName "x"
-
-  return [
-      PragmaD $ InlineP funName Inlinable ConLike AllPhases
-    , FunD funName [
-        Clause (VarP <$> funArgsNames)
-               (NormalB $ AppE (VarE 'send) $
-                 foldl1 AppE $ ConE conName : fmap VarE funArgsNames)
-               []
-      ]
-    ]
-
-------------------------------------------------------------------------------
--- | Folds a list of 'Type's into a right-associative arrow 'Type'.
-foldArrows :: [Type] -> Type
-foldArrows = foldr1 $ AppT . AppT ArrowT
-
-------------------------------------------------------------------------------
--- | Extracts name from type variable (possibly nested in signature and/or
--- some context), returns 'Nothing' otherwise.
-tvName' :: Type -> Maybe Name
-tvName' = \case
-  ForallT _ _ t -> tvName' t
-  SigT t _      -> tvName' t
-  VarT n        -> Just n
-  ParensT t     -> tvName' t
-  _             -> Nothing
+-- Removes 'Type' kind signatures from type
+removeTypeKind :: Type -> Type
+removeTypeKind = everywhere $ mkT $ \case
+  SigT t StarT    -> t
+  ForallT bs cs t -> ForallT (goBndr <$> bs) cs t
+    where
+      goBndr (KindedTV n StarT) = PlainTV n
+      goBndr b                  = b
+  t               -> t
 
 ------------------------------------------------------------------------------
 -- | Converts equality constraint with type variable to name and type pair or
@@ -239,6 +260,22 @@ eqPairOrCxt p = case asEqualPred p of
   _                -> Right p
 
 ------------------------------------------------------------------------------
+-- | Folds a list of 'Type's into a right-associative arrow 'Type'.
+foldArrows :: [Type] -> Type
+foldArrows = foldr1 $ AppT . AppT ArrowT
+
+------------------------------------------------------------------------------
+-- | Extracts name from type variable (possibly nested in signature and/or
+-- some context), returns 'Nothing' otherwise.
+tVarName :: Type -> Maybe Name
+tVarName = \case
+  ForallT _ _ t -> tVarName t
+  SigT t _      -> tVarName t
+  VarT n        -> Just n
+  ParensT t     -> tVarName t
+  _             -> Nothing
+
+------------------------------------------------------------------------------
 -- | Constructs name from base of input name with function applied
 mapNameBase :: (String -> String) -> Name -> Name
 mapNameBase f = mkName . f . nameBase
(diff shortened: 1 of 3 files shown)


src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
-getDeclName = mkName . overFirst toLower . nameBase
- where
-  overFirst f (a : as) = f a : as
-  overFirst _ as       = as
+genFreer shouldMkSigs typeName = do
+  dtInfo <- reifyDatatype typeName
+  -- TODO: data families seem to work just fine - should they be restricted?
+  clInfos  <- traverse (mkCLInfo dtInfo) $ datatypeCons dtInfo
+  defModFI <- TySynInstD ''DefiningModule
+          .   TySynEqn [ConT $ datatypeName dtInfo]
```
isovector: Nit: align like this:

```haskell
   <-
    .
  <*>
```

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+      where
+        base = mapNameBase id name
+        args = PlainTV . mkName <$> ["m", "a"]
+
+------------------------------------------------------------------------------
+-- | Generates signature for lifting function and type arguments to apply in
+-- it's body on effect's data constructor.
+genSig :: ConLiftInfo -> Q Dec
+genSig CLInfo {
+    typeName, funName, effTArgs, mName, resArg, conCxt, conArgs
```
isovector: Nit: I realize it's not particularly widespread around the codebase, but I'd like to start naming local variables with `snake_case`, leaving `camelCase` for top-level things.

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+------------------------------------------------------------------------------
 -- | Folds a list of 'Type's into a right-associative arrow 'Type'.
 foldArrows :: [Type] -> Type
-foldArrows = foldr1 (AppT . AppT ArrowT)
+foldArrows = foldr1 $ AppT . AppT ArrowT
+
+------------------------------------------------------------------------------
+-- | Extracts name from type variable (possibly nested in signature and/or
+-- some context), returns 'Nothing' otherwise.
+tvName' :: Type -> Maybe Name
```
isovector: why the prime?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+splitFromEnd :: Int -> [a] -> ([a], [a])
+splitFromEnd n lst = go lst $ drop n lst where
+  go xs     []     = ([], xs)
+  go (x:xs) (_:ys) = first (x:) $ go xs ys
 
-freeVars :: Data a => a -> [Name]
-freeVars = mkQ [] $ \case
-  VarT n -> [n]
-  _ -> []
+  go []     (_:_)  = undefined
```
isovector: yucky -- why is this `undefined`?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+------------------------------------------------------------------------------
+-- TODO: separate into some module for general utilities (or find them in
+-- available dependencies)
+
+partitionWith :: (a -> Either b c) -> [a] -> ([b], [c])
+partitionWith f = foldl' choose ([], []) where
+  choose = flip $ either (first . (:)) (second . (:)) . f
 
+splitFromEnd :: Int -> [a] -> ([a], [a])
+splitFromEnd n lst = go lst $ drop n lst where
```
isovector: Nit: `where` on its own line, indented 2

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+
+------------------------------------------------------------------------------
+-- TODO: separate into some module for general utilities (or find them in
+-- available dependencies)
+
+partitionWith :: (a -> Either b c) -> [a] -> ([b], [c])
+partitionWith f = foldl' choose ([], []) where
+  choose = flip $ either (first . (:)) (second . (:)) . f
 
+splitFromEnd :: Int -> [a] -> ([a], [a])
```
isovector: Maybe just `bimap reverse reverse . split . reverse`?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+-- | Constructs name from base of input name with function applied
+mapNameBase :: (String -> String) -> Name -> Name
+mapNameBase f = mkName . f . nameBase
+
+------------------------------------------------------------------------------
+-- TODO: separate into some module for general utilities (or find them in
+-- available dependencies)
+
+partitionWith :: (a -> Either b c) -> [a] -> ([b], [c])
+partitionWith f = foldl' choose ([], []) where
```
isovector: `= partitionEithers . fmap f`

isovector: But just inline this since it's only used once

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
-          (ConE cName)
-          dTypeVars
-        )
-        []
+genDec :: ConLiftInfo -> Q [Dec]
+-- TODO: apply effect's type arguments to avoid "bad ambiguity"
+genDec CLInfo { funName, conName, conArgs } = do
+  funArgsNames <- replicateM (length conArgs) $ newName "x"
+
+  return [
```
isovector: Nit: `[`s on their own line, indented two; align with `,`s, eg:

```
foo = 
  [ bar
  , baz
  ]
```

Here and elsewhere

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+          <$> location
+  sigs     <- if shouldMkSigs then traverse genSig clInfos
+                              else return []
+  decs     <- traverse genDec clInfos
+
+  return $ defModFI : join (sigs:decs)
 
+-------------------------------------------------------------------------------
+-- | Info about constructor being lifted; use 'mkCLInfo' to create one.
+data ConLiftInfo = CLInfo {
```
isovector: I'd prefer some haddock on these fields

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+    typeName :: Name
```
isovector: Please prefix record names with an acronym of the datatype; here `cli`

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+  , conCxt   :: Cxt
+  , conArgs  :: [Type]
+  } deriving Show
+
+------------------------------------------------------------------------------
+-- | Creates info about constructor being lifted from parent type's and it's
+-- own info.
+mkCLInfo :: DatatypeInfo -> ConstructorInfo -> Q ConLiftInfo
+-- mkCLInfo dtInfo conInfo = do
+mkCLInfo DatatypeInfo {
```
isovector: I find this hard to read; maybe parens around the record unpacking would help?

```haskell
mkCLInfo
    (DatatypeInfo
      { datatypeName = typeName
      , ...
      })
    (ConstructorInfo
      { ...
      }) = do
  let funName = ...
```

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+      -- manually - this makes type errors more readable
+      (eqPairs, conCxt) = first (M.fromList . maybeAddResKind)
+                        $ partitionWith eqPairOrCxt constructorContext
+      -- We constrain kind of result to 'Type' if we get a type variable
+      -- TODO: we probably want to add requirement for KindSignatures into
+      -- documentation - PolyKinds are already mentioned...
+      -- TODO: This makes function signatures ugly - what is the best way to
+      -- restrict occurence of kind signature or completely remove it?
+      -- TODO: take result type by name or position?
+      maybeAddResKind   = maybe id ((:) . (, StarT))
```
isovector: Too point-free. What's going on here?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+                        $ tvName' $ tvKind $ last datatypeVars
+      allEffTArgs       = applySubstitution eqPairs datatypeInstTypes
+      conArgs           = applySubstitution eqPairs constructorFields
+
+  (effTArgs, [mArg, resArg]) <- case splitFromEnd 2 allEffTArgs of
```
isovector: Nit: Case on a new line

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+      r@(_, [_, _]) -> return r
+      _             -> fail $ show $ missingEffTArgs typeName
+  mName <-
+    -- May happen when used on data families
+    maybe (fail $ show $ mArgNotVar typeName mArg) return $ tvName' mArg
+
+  return CLInfo {
```
isovector: `{` on a new line

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+-- TODO: apply effect's type arguments to avoid "bad ambiguity"
+genDec CLInfo { funName, conName, conArgs } = do
+  funArgsNames <- replicateM (length conArgs) $ newName "x"
+
+  return [
+      PragmaD $ InlineP funName Inlinable ConLike AllPhases
+    , FunD funName [
+        Clause (VarP <$> funArgsNames)
+               (NormalB $ AppE (VarE 'send) $
+                 foldl1 AppE $ ConE conName : fmap VarE funArgsNames)
```
isovector: Maybe some helpers for this kind of thing would go a long way. I wouldn't be surprised if they exist in `TH.Syntax`

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -1,52 +1,56 @@
...
-  ReadFile :: 'FilePath' -> FileSystem 'String'
-  WriteFile :: 'FilePath' -> 'String' -> FileSystem ()
-'makeSem' ''FileSystem
-@
-
-This will automatically generate the following functions:
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE TemplateHaskell     #-}
+{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE MonadComprehensions #-}
```
googleson78: I think this is unused?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Hmm, why I remembered this making syntax errors somewhere?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+  dtInfo <- reifyDatatype typeName
+  -- TODO: data families seem to work just fine - should they be restricted?
+  clInfos  <- traverse (mkCLInfo dtInfo) $ datatypeCons dtInfo
+  defModFI <- TySynInstD ''DefiningModule
+          .   TySynEqn [ConT $ datatypeName dtInfo]
+          .   LitT
+          .   StrTyLit
+          .   loc_module
+          <$> location
+  sigs     <- if shouldMkSigs then traverse genSig clInfos
```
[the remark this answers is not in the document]
TheMatten: Like this?:
```hs
  sigs     <- if shouldMkSigs
                 then traverse genSig clInfos
                 else return []
```

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
-
--- | Given the name of a GADT constructor, return the name of the corresponding
--- lifted function.
-getDeclName :: Name -> Name
-getDeclName = mkName . overFirst toLower . nameBase
- where
-  overFirst f (a : as) = f a : as
-  overFirst _ as       = as
+genFreer shouldMkSigs typeName = do
+  dtInfo <- reifyDatatype typeName
```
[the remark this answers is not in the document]
TheMatten: It works - `reifyDatatype` takes both type constructor or _some_ data constructor of type

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+          .   TySynEqn [ConT $ datatypeName dtInfo]
+          .   LitT
+          .   StrTyLit
+          .   loc_module
+          <$> location
+  sigs     <- if shouldMkSigs then traverse genSig clInfos
+                              else return []
+  decs     <- traverse genDec clInfos
+
+  return $ defModFI : join (sigs:decs)
```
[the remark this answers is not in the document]
TheMatten: Oops...

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -75,135 +79,185 @@ makeSem = genFreer True
...
+          <> text "’ has not enough type arguments"
+      $+$ nest 4
+          (   text "At least monad and result argument are required, e.g.:"
+          $+$ nest 4
+              (   text ""
+              $+$ ppr (DataD [] base args Nothing [] []) <+> text "..."
+              $+$ text ""
+              )
+          )
+      where
```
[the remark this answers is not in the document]
TheMatten: Sure, they just seemed a little bit special purpose.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok

src/Polysemy/Internal/TH/Effect.hs
TheMatten: If we've got `Just` name of kind, put it in a pair with `StarT` for substitution and make it append to supplied list of substitutions, else leave list as-is.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Done

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Good point - why doesn't compiler warn about unused extensions with `-Wall`?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: `tvName` from `th-abstraction` works on `TyVarBndrs` - I wanted to differentiate them somehow, but `tvName` is the most sensible name for both...

src/Polysemy/Internal/TH/Effect.hs
TheMatten: `-Wincomplete-patterns`...

TheMatten: It's a bad idea to disable it, so I'm being explicit instead - this is not Idris, so I can't just prove this doesn't happen...

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Yeah, I'm just always cautious about expensive `reverse` - If we're not moving this into separate module (where it may be used on more than e.g `5` element lists), then it's probably more sensible implementation.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: I completely missed that in `Data.Either`, sorry.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Will do

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Like `cliTypeName` or `cli_typeName`?

src/Polysemy/Internal/TH/Effect.hs
googleson78: Unsure.

`hlint` catches some of them.

Here's something I found on the subject: https://gitlab.haskell.org/ghc/ghc/issues/3085

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Changed to `tVarName`.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Will look into it.

src/Polysemy/Internal/TH/Effect.hs
isovector: I meant more generally :) like why _can't_ this case happen? Maybe add a comment with your reasoning. Change the `undefined` to an error with the name of the function as well, so if it _does_ go wrong we have a chance of finding it.

src/Polysemy/Internal/TH/Effect.hs
isovector: Some explicit lambdas would help here!

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
isovector: My feeling is "if you found them useful, somebody else will too" :)

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
isovector: so if i go:

```
foo :: MyEffect m a
foo = undefined

makeSem 'foo
```

will just work fine? Could we make that an error?

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
isovector: yeah baby yeah

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
TheMatten: Oh, not any name - just type constructor/data constructor/field name of some effect (latter because of data families - you won't be able to choose specific instance otherwise) - and `th-abstraction` actually has nice errors when user tries something stupid :smile:

src/Polysemy/Internal/TH/Effect.hs
jhenahan: `\case { Nothing -> id; Just rk -> ((rk, StarT):) }` is similarly terse but more obvious, and `LambdaCase` is already common in the codebase.

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
TheMatten: Done :slightly_smiling_face:

src/Polysemy/Internal/TH/Effect.hs
TheMatten:
```
      maybeAddResKind       = (\case Nothing -> id
                                     Just k  -> ((k, StarT) :))
                            $ tVarName $ tvKind $ last datatypeVars
```
looks fine :slightly_smiling_face:

isovector: I'm happy with `maybe`, it's the `(:) . (, StarT)` but I don't want to have to unravel in my brain

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Can't find any useful - should I create some helpers locally? Because I'm not sure if it's really worth it.

TheMatten: It's pity that quotations aren't more flexible - I can't for example splice function name into declaration...

isovector: Don't worry about it in that case. There are more important things in the world than pretty-looking TH.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Hmm, what about `maybe id (\k ps -> (k, StarT) : ps)`?

isovector: Much better; now I understand what's happening here!

TheMatten:
Some things to do:
- [ ] update documentation (we can use data families now! + `makeSem` accepts some data constructor or field name of effect too + what about `KindSignatures`?)
- [ ] write tests (DFs, playing with monad and result argument, custom arguments, constraints, non-GADT syntax, etc.)
- [ ] apply effect's custom type arguments to data constructor in body (#48)
- [ ] `splitFromEnd` and `mapHead` should be in some general module instead

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Great!

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Done.

isovector:
Of these TODOs, I'm most interested in getting #48 finished. You can steal the [freer-simple test](https://github.com/lexi-lambda/freer-simple/blob/master/tests/Tests/TH.hs) for some of the hairier cases. Let's hold off on the fourth point until some other code in the codebase wants to use them.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Added descriptive `error`.

TheMatten:
It seems to be nearly finished - we just need to resolve some small questions and make sure that I am not doing something stupid :smile:

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
...
+    base = mapNameBase id name
+    args = PlainTV . mkName <$> ["m", "a"]
+
+checkExtensions :: [Extension] -> Q ()
+checkExtensions exts = do
+  states <- zip exts <$> traverse isExtEnabled exts
+  maybe (return ())
+        (\(ext, _) -> fail $ show
+          $ char '‘' <> text (show ext) <> char '’'
+            <+> text "extension needs to be enabled\
```
isovector: what is this trailing sorcery?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
...
+-- | Converts names of all type variables in type to capturable ones based on
+-- original name base. Use with caution, may create name conflicts!
+capturableTVars :: Type -> Type
+capturableTVars = everywhere $ mkT $ \case
+  VarT n          -> VarT $ mapNameBase id n
+  ForallT bs cs t -> ForallT (goBndr <$> bs) (capturableTVars <$> cs) t
+    where
+      goBndr (PlainTV n   ) = PlainTV $ mapNameBase id n
+      goBndr (KindedTV n k) = KindedTV (mapNameBase id n) $ capturableTVars k
+  t               -> t
```
isovector: Maybe just `t -> t` without the massive indent

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
...
+
+------------------------------------------------------------------------------
+-- | Converts names of all type variables in type to capturable ones based on
+-- original name base. Use with caution, may create name conflicts!
+capturableTVars :: Type -> Type
+capturableTVars = everywhere $ mkT $ \case
+  VarT n          -> VarT $ mapNameBase id n
+  ForallT bs cs t -> ForallT (goBndr <$> bs) (capturableTVars <$> cs) t
+    where
+      goBndr (PlainTV n   ) = PlainTV $ mapNameBase id n
```
isovector: Is the `mapNameBase id` bit necessary?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
...
+  $+$ nest 4
+      (   text "At least monad and result argument are required, e.g.:"
+      $+$ nest 4
+          (   text ""
+          $+$ ppr (DataD [] base args Nothing [] []) <+> text "..."
+          $+$ text ""
+          )
+      )
+  where
+    base = mapNameBase id name
```
isovector: why map id?

src/Polysemy/Internal/TH/Effect.hs
isovector: what is going on here? it's in a few places; maybe `mapNameBase` needs a more descriptive name

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
...
+simplifyKinds :: Type -> Type
+simplifyKinds = everywhere $ mkT $ \case
+  SigT t StarT    -> t
+  SigT t VarT{}   -> t
+  ForallT bs cs t -> ForallT (goBndr <$> bs) (simplifyKinds <$> cs) t
+    where
+      goBndr (KindedTV n StarT ) = PlainTV n
+      goBndr (KindedTV n VarT{}) = PlainTV n
+      goBndr b                  = b
+  t               -> t
```
isovector: same

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
...
+-- Removes 'Type' and variable kind signatures from type
+simplifyKinds :: Type -> Type
+simplifyKinds = everywhere $ mkT $ \case
+  SigT t StarT    -> t
+  SigT t VarT{}   -> t
+  ForallT bs cs t -> ForallT (goBndr <$> bs) (simplifyKinds <$> cs) t
+    where
+      goBndr (KindedTV n StarT ) = PlainTV n
+      goBndr (KindedTV n VarT{}) = PlainTV n
+      goBndr b                  = b
```
isovector: too much indent for not enough visual readability. just `goBndr b = b` is fine

src/Polysemy/Internal/TH/Effect.hs
isovector: I think this is worth doing. Slow reverse isn't really a problem for n=5. Especially since you don't ever inspect the `snd` of the result, so the compiler won't actually perform it.

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -64,146 +76,259 @@ makeSem = genFreer True
 -- makeSem_ ''Lang
 --
 -- -- | Output a string.
--- output :: Member Lang r
---        => String         -- ^ String to output.
+-- output :: forall r
```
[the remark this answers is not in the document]
TheMatten: Should I use `Output` or make some different example?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: 79-character limit :smile: Should I format it differently?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
isovector: Either is fine

src/Polysemy/Internal/TH/Effect.hs
isovector: I didn't know about this feature, so I'm mostly surprised. I don't love it, but maybe I will come to, in time.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Will do this differently.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Changed implementation.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Is `splitAtEnd n = swap . join bimap reverse . splitAt n . reverse` fine?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Is `splitAtEnd n = swap . join bimap reverse . splitAt n . reverse` fine?

src/Polysemy/Internal/TH/Effect.hs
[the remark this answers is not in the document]
TheMatten: Okay, will copy `Output` :smile:

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Done.

TheMatten:
Is documentation fine? And there's this extracting of result argument using `last` that I'm (maybe overly) nervous about :smile:

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -1,54 +1,67 @@
...
--- | If @T@ is a GADT representing an effect algebra, as described in the module
--- documentation for "Polysemy", @$('makeSem' ''T)@ automatically
+-- TODO: write tests for what should (not) compile
+
+------------------------------------------------------------------------------
+-- | If @T@ is a datatype representing an effect algebra, as described in the
+-- module documentation for "Polysemy", @$('makeSem' ''T)@ automatically
 -- generates a smart constructor for every data constructor of @T@.
 --
+-- Datatype may be created with GADT syntax, traditional syntax and can be
```
isovector: I think I'd prefer to not publish this fact; life is easier for everyone if we all agree just to use GADT syntax.

isovector: Maybe just have a comment "this also will work for data families", so that people who have a use case won't be scared off, but people who don't won't be inspired to try to make things more complicated.

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -58,152 +71,279 @@ makeSem = genFreer True
...
--- Note that 'makeSem_' must be used /before/ the explicit type signatures.
+-- Because of limitations in Template Haskell, signatures have to follow some
+-- rules to work properly:
+--
+-- * 'makeSem_' must be used /before/ the explicit type signatures
+-- * signatures have to specify argument of 'Sem' representing union of
+-- effects as @r@ (e.g. @'Sem' r ()@)
+-- * all arguments in effect's type constructor that are type variables in
+-- action's signature have to use the same names as in type constructor; if
+-- they became more specific, they should use names from constructor's
```
isovector: I don't necessarily understand what you mean by "more specific" here. Maybe elaborate?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -58,152 +71,279 @@ makeSem = genFreer True
...
+                         ]
+  decs       <- traverse (genDec should_mk_sigs) cl_infos
+
+  let sigs = [ genSig <$> cl_infos | should_mk_sigs ]
+
+  return $ join $ def_mod_fi : sigs ++ decs
+
+------------------------------------------------------------------------------
+-- | Generates signature for lifting function and type arguments to apply in
+-- it's body on effect's data constructor.
```
isovector: The smallest of all nits: "its"

isovector:
Three last comments on the comments; then lets get this puppy merged!!

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Done.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: More specific = contains some concrete types; e.g. `()`, `Int` or `Maybe x` - in those cases we get concrete types (not really important) or new names from constructor's signature (like `x` above). Should I change the documentation?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Still waiting for English syntax checker :smile:

src/Polysemy/Internal/TH/Effect.hs
isovector: I suspect people will figure out the right thing to do. Maybe let's just chop out this bit about becoming more specific, and give the two examples.

TheMatten:
Okay, can't wait! :smile:

I will make another PR for tests - `Complicated` from `freer-simple` seems to work perfectly, but I want to be sure :slightly_smiling_face:

TheMatten:
Is this version fine? :slightly_smiling_face:

Issue #69, opened 2019-05-29
it sucks that you can't call callbacks


isovector:
Consider this:

```haskell
    wv     <- input
    h      <- bindT hook
    istate <- getInitialStateT

    void $ sendM $ after wv #keyPressEvent $ \kp -> do
      kv   <- get kp #keyval
      void $ lower .@ runNavigation $ h $ chr (fromIntegral kv) <$ istate
      pure False
```

even though `hook` returns a `Bool`, `h` returns an `f Bool`. But an `f Bool` is no good because I need a regular bool for the line below!

DANG DOES THIS EVER SUCK.

The problem is that we don't know which higher-order interpreters may have already run that have thrown off our chances for static evacuation of `f`. Ok, but the interpreters themselves know. What if we added a constructor to `Yo`:

```haskell
evacuate :: (forall x. f x -> Maybe x)
```

Effects like state could say "yeah no problem bruh." Effects like `NonDet` or `Error`, nope, no can do. But I'm reasonably sure the only `Yo`s you have to deal with are the ones that have already run above you, so as long as you keep your `Error`s low in the stack, it should be fine?

If `evacuate` is feasible, we can stick it into `Tactics` and then in some cases call the callback above. Is this a good idea? I don't know, it's 2am.

isovector:
I guess `Error` can do it in the `Right` case. So that's dope. Maybe there should be another bit:

```haskell
data EvacuationState = Always | Sometimes | Never

evacuationState :: EvacuationState
```

so that if you're concerned you can check and see what's what. is this valuable I HAVE NO IDEA. Maybe it should be reified at the TYPE LEVEL :o :o :o

isovector:
Ok, what if `r` instead of just having the effect row also kept track of static properties of the effect. It could be like this:

```haskell
newtype Row (status :: EvacuationSatus) = Row
  { effects :: [(* -> *) -> * -> *]
  }
```

then `runError` can change the evacuation status in `r`, and we can write interpreters like the motivating example which REQUIRE THE ABILITY TO EVACUATE

isovector:
```haskell
runError :: Sem (Row (Error e ': r) status) a -> Sem (Row r 'Sometimes) (Either e a)
```

isovector:
Most code doesn't give a shit, so maybe this could be packaged into `Sem` if it were a tysym:

```haskell
type Sem = forall status. SEM status
```

isovector:
This way  `r` stays just the effect row

[...]

PR #71, opened 2019-05-29
GetInspectorT

isovector:
Some of the ideas in #69

### DIFF
--- a/src/Polysemy/Internal/Tactics.hs
+++ b/src/Polysemy/Internal/Tactics.hs
@@ -1,17 +1,17 @@
-{-# LANGUAGE AllowAmbiguousTypes   #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
 
 {-# OPTIONS_HADDOCK not-home #-}
 
 module Polysemy.Internal.Tactics
   ( Tactics (..)
   , getInitialStateT
-  , getEvacuatorT
+  , getInspectorT
+  , Inspector (..)
   , runT
   , bindT
   , pureT
   , liftT
   , runTactics
-  , runEvacuator
   , Tactical
   , WithTactics
   ) where
@@ -19,11 +19,6 @@ module Polysemy.Internal.Tactics
 import Polysemy.Internal
 import Polysemy.Internal.Effect
 import Polysemy.Internal.Union
-
-
-newtype Evacuator f = Evacuator
-  { runEvacuator :: forall x. f x -> Maybe x
-  }
 
 
 ------------------------------------------------------------------------------
@@ -85,7 +80,7 @@ type WithTactics e f m r = Tactics f m (e ': r) ': r
 data Tactics f n r m a where
   GetInitialState     :: Tactics f n r m (f ())
   HoistInterpretation :: (a -> n b) -> Tactics f n r m (f a -> Sem r (f b))
-  GetEvacuator        :: Tactics f n r m (Evacuator f)
+  GetInspector        :: Tactics f n r m (Inspector f)
 
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


isovector:
Look at the following program:

```haskell
{-# LANGUAGE TemplateHaskell #-}

import Control.Monad
import Polysemy
import Polysemy.State
import Polysemy.Error
import Polysemy.Internal.Tactics


doCB :: IO String -> IO ()
doCB m = m >>= putStrLn


data Callback m a where
  Callback :: m String -> Callback m ()

makeSem ''Callback


runCallback
    :: Member (Lift IO) r
    => (forall x. Sem r x -> IO x)
    -> Sem (Callback ': r) a
    -> Sem r a
runCallback lower = interpretH $ \case
  Callback cb -> do
    cb' <- runT cb
    ev <- runEvacuator <$> getEvacuatorT
    sendM $ doCB $ do
      v <- lower .@ runCallback $ cb'
      pure $ maybe ":(" id $ ev v
    getInitialStateT


main :: IO ()
main = do
  void . (runM .@ runCallback)
       . runState False
       $ do
    sendM $ putStrLn "hello world"
    callback $ show <$> get @Bool
    modify not
    callback $ show <$> get @Bool

  putStrLn "\n"

  void . (runM .@ runCallback)
       . runError @()
       $ do
    sendM $ putStrLn "hello world"
    callback $ throw ()
```

which prints

```
hello world
False
True


hello world
:(
```

isovector:
@jkachmar this would let you do your retry thing

isovector:
I don't see any reason not to merge this. The fancier stuff in #69 can be done in some sort of `polysemy-2` release.

TIME TO BIKESHED: 

What's a better name for this function: `f a -> Maybe a`, where `f` is some existential bastard?

jhenahan:
I might bikeshed `hush` to `rightMay` or `rightMaybe` or `eitherMaybe`. All those names come up in Hoogle, but `hush` is the least obvious unless you go find the type signature.

As for `f a -> Maybe a`, the closest analog I can find is something like `gcopoint` (generic `copoint`), but that's not exactly informative. That said, taking a lesson from `Control.Functor.Copointed`'s `extract :: f a -> a` , maybe you could call it `extractMay` or `extractMaybe`.

NickSeagull:
+1 to `extractMaybe`. In the end editors autocomplete the function names, so having long names that state what they do is awesome 😄

isovector:
I'm keen on neither `hush` nor `extractMaybe`, since both have existing intuitions in the ecosystem, which don't apply here:

* `hush`: from the `errors` package. A natural transformation `Either e ~> Maybe`, _which silences the errors_. I agree that the type looks similar here, but the concern is the difference in intentions. We don't want to silence the errors, we just want to get our hands on the insides!
* `extractMaybe`: Extracting is the right idea; but the name is already taken by comonads. 

Something else to keep in mind is that the thing we're naming isn't pure. Its' type is more this:

```haskell
  :: Sem (WithTactics e f m r) (forall x. f x -> Maybe x)
```

jhenahan:
Non-serious: `extractMaybe` -> `extractics`. 😆 

From the docs
```haskell
-- The @f@ type here is existential and corresponds to "whatever
-- state the other effects want to keep track of." @f@ is always
-- a 'Functor'.
```

Maybe something like `fromExternalState` or even `fromState`, though it's not necessarily `State`, so that could be misleading. `fromExternal` or `fromEnv`, maybe?

If you prefer punny names, `restate` might be cute.

NickSeagull:
I like `fromExternal` and `restate` 😁

src/Polysemy/Internal/Tactics.hs
```diff
@@ -96,6 +91,40 @@ getInitialStateT :: forall f m r e. Sem (WithTactics e f m r) (f ())
 getInitialStateT = send @(Tactics _ m (e ': r)) GetInitialState
 
 
+------------------------------------------------------------------------------
+-- | Get a natural transformation capable of potentially inspecting values
+-- inside of @f@. Binding the result of 'getInspectorT' produces a function that
+-- can sometimes peek inside values returned by 'bindT'.
```
NickSeagull: What are the cases when it cannot inspect the values? Perhaps it'd be nice to document them?

PR #76, opened 2019-05-31
Support operators as names of actions

TheMatten:
While current TH would work fine without them, someone may find them useful - operator names of lifting functions are made by replacing colon `:` (which denotes constructor) with 'at' symbol `@`. Additionally, fixities of both normal names and operators are now transferred from original constructors when generated through `makeSem`.

- [x] Should operators be mentioned in the documentation?
- [x] Are there any alternatives to `@` symbol?


TheMatten:
Oops, I reused `makesem-th` branch - should I recreate PR on new one?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -36,16 +36,16 @@ import Prelude hiding ((<>))
 
 import Control.Monad
 import Data.Bifunctor
-import Data.Char                      (toLower)
+import Data.Char                             (toLower)
```
isovector: Any chance we can just use one space for these?

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -140,18 +140,20 @@ genFreer should_mk_sigs type_name = do
...
 ------------------------------------------------------------------------------
 -- | Generates signature for lifting function and type arguments to apply in
 -- its body on effect's data constructor.
-genSig :: ConLiftInfo -> Dec
+genSig :: ConLiftInfo -> [Dec]
 genSig cli
-  = SigD (cliFunName cli) $ quantifyType
-  $ ForallT [] (member_cxt : cliFunCxt cli)
-  $ foldArrows $ cliFunArgs cli ++ [sem `AppT` cliResType cli]
+  =  maybe [] ((:[]) . (`InfixD` cliFunName cli)) (cliFunFixity cli)
```
isovector: I prefer `flip` instead of infix syntax, and `pure` instead of `(:[])`, as both trim down the amount of non-letters I need to read. Perhaps `maybe [] (pure . flip InfixD $ cliFunName cli)`?

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Ok.

src/Polysemy/Internal/TH/Effect.hs
TheMatten: Yeah, probably better.

isovector:
Could we maybe just drop the `:` for the value level? Ie `:>>>` becomes `>>>`?

TheMatten:
Hmm, okay.

TheMatten:
Will recreate in new branch with tests.

PR #82, opened 2019-06-04
Add a warning about use of type aliases

sevanspowell:
- Add a note to the documentation warning users to not use type
  aliases to distinguish between effect types.

### DIFF
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -130,6 +130,43 @@ import Polysemy.Internal.Union
 -- Notice that we must use @-XTypeApplications@ to specify that we'd like to
 -- use the ('Polysemy.Output.Output' 'Int') effect.
 --
+-- Note:
+-- Don't use type aliases to distinguish between different types of
+-- effects, the differences are erased at compile-time. The following
+-- will not work as you might expect:
+--
+-- @
+-- type A = String
+-- type B = String
+--
+-- foo
+--   :: ( Member (Input A) r
+--      , Member (Input B) r
+--      )
+--   => Sem r a
+-- foo = do
+--   a <- input @A
+--   b <- input @B
+-- @
+--
+-- The "A" input will bound to both "a" and "b".
+--
+-- Instead, use newtypes:
+--
+-- @
+-- newtype A = A String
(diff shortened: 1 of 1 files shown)


sevanspowell:
Let me know what you think, I'm not aware of your intended tone or style of documentation.

src/Polysemy/Internal.hs
```diff
@@ -130,6 +130,43 @@ import Polysemy.Internal.Union
...
+--   :: ( Member (Input A) r
+--      , Member (Input B) r
+--      )
+--   => Sem r a
+-- foo = do
+--   a <- input @A
+--   b <- input @B
+-- @
+--
+-- The "A" input will bound to both "a" and "b".
```
isovector: Well, the `String` input will be; there is no distinction between `A` and `B`.

src/Polysemy/Internal.hs
```diff
@@ -130,6 +130,43 @@ import Polysemy.Internal.Union
...
+-- Don't use type aliases to distinguish between different types of
+-- effects, the differences are erased at compile-time. The following
+-- will not work as you might expect:
+--
+-- @
+-- type A = String
+-- type B = String
+--
+-- foo
+--   :: ( Member (Input A) r
```
isovector: In all of your code samples, please change:

* `Member` => `'Member'`
* `Input` => `'Polysemy.Input.Input'`
* `input` => `'Polysemy.Input.input'`
* `Sem` => `'Sem'`

which will make the haddocks link properly.

isovector:
Upon further thought. I wonder if this might be better cleaned up just by turning on `-Wall`, which should warn about the redundant `Member (Input B) r` constraint.

sevanspowell:
Huh, I didn't even know that would happen, that sounds best. I'll give it a go tomorrow.

ocharles:
I don't think -Wall implies -fwarn-redundant-constraints

isovector:
@sevanspowell does `-fwarn-redundant-constraints` solve your problem?

sevanspowell:
Sorry, `-fwarn-redundant-constraints` does resolve the problem. It warns about:

PR #83, opened 2019-06-04
Split out effect lib

isovector:
This PR teases apart the reusable pieces of the current TH machinery. Doing so allows us to build other TH abstractions in polysemy --- such as an RPC effect.

Todo:

- [x] Haddock


isovector:
The RPC stuff should live in the zoo, but I wanted to make sure the EffectLib changes were finished before moving it there.

isovector:
@TheMatten what do you think?

TheMatten:
Seems fine. And that RPC effect is interesting 🙂
When it comes to name for that newly created module - if it's going to contain more shared code for different purposes, maybe something like `Polysemy.Internal.TH.Common` could work.

isovector:
I'm wondering if maybe this stuff shouldn't live in its own package at the end of the day --- `polysemy-th`, and then move everything there.

TheMatten:
So you are thinking about possibility to use `polysemy` without TH and it's dependencies?

isovector:
This PR teases apart the reusable pieces of the current TH machinery. Doing so allows us to build other TH abstractions in polysemy --- such as an RPC effect.

Issue #73, opened 2019-05-30
MonadUnliftIO support?


ocharles:
#36 talks about interop with `mtl`, but could we also support `MonadUnliftIO`? I haven't tried anything at all yet, just raising this from the lack of `instance MonadUnliftIO (Sem r)`.

isovector:
I think the solution is to use @adamConnerSax's new `absorb` machinery to magick up an instance of `MoandUnliftIO` given a natural transformation `forall a. Sem r a -> IO a`

isovector:
I'd definitely accept a PR in https://github.com/isovector/polysemy-zoo for this. See the examples in https://github.com/isovector/polysemy-zoo/pull/8 for inspiration.

ocharles:
Ok, jamming some code into `Polysemy.MTL` I got this:

```haskell
data UnliftIO m a where
  UnliftIO :: ( ( forall x. m x -> IO x ) -> IO a ) -> UnliftIO m a

makeSem ''UnliftIO
    
absorbMonadUnliftIO :: forall r a. Member UnliftIO r
  => (S.MonadUnliftIO (Sem r) => Sem r a) -> Sem r a
absorbMonadUnliftIO = absorb @S.MonadUnliftIO

instance ReifiableConstraint1 S.MonadUnliftIO where
  data Dict1 S.MonadUnliftIO m = MonadUnliftIO
    { unliftIO_ :: forall a. ( ( forall x. m x -> IO x ) -> IO a ) -> m a
    }
  reifiedInstance = Sub Dict

instance ( Monad m
         , R.Reifies s' (Dict1 S.MonadUnliftIO m)
         ) => S.MonadIO (ConstrainedAction S.MonadUnliftIO m s') where
  liftIO io = ConstrainedAction $ unliftIO_ (R.reflect $ Proxy @s') (\_ -> io)

instance ( Monad m
         , S.MonadIO (ConstrainedAction S.MonadUnliftIO m s')
         , R.Reifies s' (Dict1 S.MonadUnliftIO m)
         ) => S.MonadUnliftIO (ConstrainedAction S.MonadUnliftIO m s') where
  withRunInIO f =
    ConstrainedAction $ unliftIO_ (R.reflect $ Proxy @s') (\k -> f (\(ConstrainedAction a) -> k a))


instance Member UnliftIO r => IsCanonicalEffect S.MonadUnliftIO r where
  canonicalDictionary = MonadUnliftIO unliftIO


runUnliftIO
[...]

PR #87, opened 2019-06-08
Better Resource bracketing

isovector:
https://github.com/isovector/polysemy/issues/73#issuecomment-500042179 pointed out that the `Resource` effect was missing a lot of things. This PR fixes that!

* [x] Change log needs to be updated
* [x] Update readme example
* [x] Update benchmark


isovector:
@lexi-lambda does this address your issues?

PR #100, opened 2019-06-11
Update polysemy-plugin version bound

Infinisil:
The test suite is made for 0.3 and doesn't compile with earlier versions due to Yo now taking another argument since https://github.com/isovector/polysemy/pull/71


isovector:
Thanks!

PR #105, opened 2019-06-14
Allow bidirectional typechecking in the plugin

### LINKED ISSUE #95: Plugin doesn't propagate unification information from inside to out
```haskell
flipShouldBe 11 . sum . fst . run . runFoldMapOutput id $ do
  output [1]
  output $ replicate 2 5
```

This thing should infer from `[1]` that `sum` is at `@[]`.  But it doesn't.

The problem is that the given is `Output (t a)`, but the _wanted_ is `Output [Int]`. This is caused by there being no real givens, and just cheating here:

https://github.com/isovector/polysemy/blob/ca0c68e1c35be4ec45662c36b7dbb12980d29e81/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs#L128

This logic is backwards from everywhere else in which we do not want to unify a polymorphic given against a concrete wanted. I'm not sure what the fix is.

isovector:
The plugin used to choke on this:

```haskell
flipShouldBe 11 . sum . fst . run . runFoldMapOutput id $ do
  output [1]
  output $ replicate 2 5
```

because it would fail to unify `Output (t Int)` (the polymorphic `Traversable` variable in `sum`) with `Output [Int]`.

In this case, there are no given constraints, so the plugin is attempting to solve a `Member (Output (t Int)) '[Output [Int]]`. We reuse the codepath for unifying wanted/givens, pretending like the
`Output (t Int)` is a given.[^1]

[^1]: Pretending it's a wanted breaks the common case of trying to call, eg. `runState (5 :: Num a => a)`  on an effect `State Int`.

So now we have a "wanted" `Output [Int]`, and a "given" `Output (t Int)`. In general, this thing isn't OK to solve. Consider a real example:

```haskell
foo :: Member (Output (t Int)) r => Sem r ()
foo = output [5]
```

This is a type error, because the polymorphism goes the wrong way. We have no guarantees that `t` is supposed to be `[]`.

But in our original example, this isn't a problem, because our `t` isn't actually in given position. It's just an artifact of reusing the code! `mkWanted` now takes a new parameter for whether or not it's OK to allow
the givens to be polymorphic.

Such a thing necessitates a change though. We never want to unify a polymorphic *effect* in given position. Doing so will break Haskell's regular type inference that determines what the effect row should be, based on the order in which the interpreters are run.

Fixes #95

### DIFF
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -34,8 +34,10 @@ module Polysemy.Plugin.Fundep (fundepPlugin) where
 
 import           Class
 import           CoAxiom
+import           Control.Applicative
 import           Control.Monad
 import           Data.Bifunctor
+import           Data.Bool
 import           Data.Function (on)
 import           Data.IORef
 import           Data.List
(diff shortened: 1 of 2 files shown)


isovector:
I've tested this against `polysemy-zoo`. I'm also trying to run it against [`hskit`](https://github.com/isovector/hskit), but currently stuck in dependency hell. Will report back when I have the results.

TheMatten:
If someone tries to put this into ghci (as I did :smile:) - the reason it works here is that ghci has `ExtendedDefaultRules` enabled, which probably finds `Foldable t` from `sum` and fits `[]` in there.

Issue #107, opened 2019-06-14
Member constraints can't reference existentially quantified variables


isovector:
https://github.com/isovector/polysemy/issues/103#issuecomment-502027385 gives an example in which an existentially scoped type variable `s` is attempted to be used in the context of a `Member (State s) r` constraint.

[Trac #9223](https://gitlab.haskell.org/ghc/ghc/issues/9223) suggests that this is caused by `Member`'s use of `~`:

https://github.com/isovector/polysemy/blob/18e16f68db2ae77d54d23b18aafa1b82a4f64b0c/src/Polysemy/Internal/Union.hs#L105

This is more of a problem in GHC than in polysemy, though thankfully there is a workaround. If we introduce the `Member` constraint, and then existentialize afterwards, everything works. See https://github.com/isovector/polysemy/issues/103#issuecomment-502145910 for details.

isovector:
It's probably worth opening a GHC bug for this.

isovector:
Apparently section 5.2 of [https://www.microsoft.com/en-us/research/publication/outsideinx-modular-type-inference-with-local-assumptions/](OutsideIn) is good reading on this topic.

incertia:
i'm not quite sure how to adapt the given workaround to the following code trying to abstract around input streams (sockets, stdin/out, files, websockets, etc)

```haskell
{-# LANGUAGE AllowAmbiguousTypes #-}
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE LambdaCase #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeOperators #-}

module Main where

import Data.Function ((&))
import Polysemy
import System.IO

data Stream s m a where
  Recv :: Stream s m String
  Send :: String -> Stream s m ()
makeSem ''Stream

streamToIO :: forall s r a. (Member (Embed IO) r)
           => String
           -> (Sem (Stream s ': r) a)
           -> Sem r a
streamToIO name action = do
  interpret (embed @IO . \case
      Recv     -> putStr (name ++ "> ") >> getLine
      Send buf -> putStr (name ++ "> ") >> putStrLn buf
    ) action

[...]

PR #110, opened 2019-06-15
Update 'Union' with use of more precise proof

TheMatten:
Use of more precise:
```hs
data Elem :: Eff -> [Eff] -> Type where
  Here ::              Elem e (e ': es)
  In   :: Elem e es -> Elem e (d ': es)
```
instead of `Nat`+`SNat` allows us to prove important properties and makes implementation a little bit simpler, e.g.:
```hs
absurdU :: Union '[] m a -> a
absurdU = absurdU
```
can now be proved with `EmptyCase`:
```hs
absurdU (Union p  _) = case p of {}
```
TODO:
- [ ] Decide on custom type errors
- [ ] Make fundep plugin work again

### DIFF
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -54,8 +54,6 @@ import           TcSMonad hiding (tcLookupClass)
 import           TyCoRep (Type (..))
 import           Type
 
-import           Outputable
-
 
 polysemyInternalUnion :: ModuleName
 polysemyInternalUnion = mkModuleName "Polysemy.Internal.Union"
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -52,6 +52,8 @@ import           TcSMonad hiding (tcLookupClass)
 import           TyCoRep (Type (..))
 import           Type
 
+import           Outputable
+
 
 polysemyInternalUnion :: ModuleName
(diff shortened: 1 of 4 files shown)


isovector:
Before going too far on this, I'd want to look at the perf. I have a sneaking suspicion that this thing will allocate much harder than `Union`. I spent a looooong time making the perf good on the current impl, and suspect that it's quite brittle to changes like these. Maybe we should hold off until 8.10 when it becomes easy to test something like this?

TheMatten:
Okay, no problem - should I close it for now?

isovector:
Your call! I'm curious what tangible benefits this buys us --- does it allow us to do anything we couldn't otherwise, or is it merely aesthetic?

TheMatten:
Actually, #107 comes to my mind - `Member'` constraint would avoid need for equality (but there may be some other way too)

isovector:
I'm going to close this for now --- feel free to reopen if you feel we should reconsider.

KingoftheHomeless:
Stuff like #281 has made more more and more attracted to a solution like this. It'd make the machinery behind those kinds of stuff prettier. `membership :: Member e r => Elem r e` is also really powerful, and would enable stuff like a `Compose :: EffectRow -> Effect` effect, which I have trouble implementing using the current machinery.

You've commented about performance, but shouldn't `Elem` have just the same internal representation as `SNat`?

isovector:
That's a reasonable argument. Anyone want to tackle this?

KingoftheHomeless:
I already have a solution in the works, so I'm up for it, but @TheMatten is probably also interested.

TheMatten:
@KingoftheHomeless I currently don't have time to finish this - go for it if you want :slightly_smiling_face:

PR #113, opened 2019-06-15
Custom type error testing

isovector:
This PR adds doctests allowing us to write tests for the custom type errors. Having this stuff reified in the test suite means we can iterate on the implementations and give much better QA.


isovector:
Many of the crap error messages have obvious solutions :)

PR #123, opened 2019-06-19
Trying to fix some Travis stuff

jkachmar:
Changes:
- add `cabal.project` to let `cabal v2-*` commands know about `polysemy-plugin`
  - also provides some nice `ghc-options` that should speed up compilation (just slightly)
- drops `transformers` lower bound to `0.5.2.0` (which is what it looks like GHC 8.2.2 ships with)
- adds `.ghc.environment` and `dist-newstyle` files to the `.gitignore` since the new `cabal` makes these files and they're annoying
- bumps `stack` LTS to 13.25 to pull in GHC 8.6.5 since that's the latest version of the GHC 8.6 release train
  - bumps GHC version in the Travis test matrix

---

Running `cabal v2-build` works for both `polysemy` and `polysemy-plugin` for GHC 8.2.2 -> GHC 8.6.5, but running `cabal v2-test polysemy-plugin` fails for GHC 8.2.2 with the following error:

<details><summary>Error Log Snippet</summary>
<p>

```
test/PluginSpec.hs:20:8: error:
    • Ambiguous use of effect 'State'
      Possible fix:
        add (Member (State s0) r) to the context of 
          the type signature
      If you already have the constraint you want, instead
        add a type application to specify
          's0' directly, or activate polysemy-plugin which
            can usually infer the type correctly.
    • In a stmt of a 'do' block: s <- get
      In the expression:
        do s <- get
           put s
      In an equation for ‘idState’:
          idState
            = do s <- get
                 put s
   |
20 |   s <- get
   |        ^^^

[... more of the same ...]
```
</p>
</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ stack.yaml.lock
 core-html/
 *~
 *.prof
+.ghc.environment.*
+dist-newstyle
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ matrix:
   include:
     - ghc: "8.6.5"
     - ghc: "8.4.4"
-    - ghc: "8.2.2"
+#     - ghc: "8.2.2"
+# Disabling GHC 8.2.2 as polysemy-plugin tests do not pass
 
 install:
  - cabal --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cabal: "2.4"
 
 matrix:
   include:
-    - ghc: "8.6.3"
+    - ghc: "8.6.5"
     - ghc: "8.4.4"
     - ghc: "8.2.2"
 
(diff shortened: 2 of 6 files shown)


jkachmar:
I tested this by installing [`ghcup`](https://gitlab.haskell.org/haskell/ghcup) and then using that to install GHC 8.2.2, 8.4.4, and 8.6.5 along with `cabal-install-2.4.1.0`, then building and testing `polysemy` and `polysemy-plugin` against those GHC versions.

isovector:
Easy as #123. I'm not sure if this means we should drop support for polysemy at 8.2.2 due to the plugin's failure, or find a way to work around it in the CI.

isovector:
Thanks Joe! You're a peach!

Issue #84, opened 2019-06-05
State is discarded inside `bracket` dealloc actions


lexi-lambda:
Both these results seem wrong to me:

```haskell
ghci> runM .@ runResource $ runState 0 $ bracket
        (pure ())
        (const (sendM . print =<< get @Integer))
        (const (put 2))
0
(2,())
ghci> runM .@ runResource $ runState 0 $ bracket
        (pure ())
        (const (put 2))
        (const (pure ()))
(0,())
```

It seems like it ought to be possible to fix this by using a custom version of `bracket` inside `runResource` that threads the state through appropriately, but if that’s necessary, then I suppose I’m a little confused about what makes the approach `polysemy` takes any better than `MonadBaseControl` with respect to dropping state.

isovector:
Both of these seem wrong to me too; I'm sure that this used to work. Let me dive in and write some tests.

isovector:
Nevermind, this is "expected behavior," for some definition of expected.

The type of base's `bracket` is `bracket :: IO a -> (a -> IO b) -> (a -> IO c) -> IO c`, which is pretty explicit that it's dropping the `b`. polysemy, fused-effects, and indeed all implementations based on "Effect Handlers in Scope" are going to have this same limitation, since they manipulate the return types to push state around.

There's a workaround, which is to use [`runStateInIORef`](https://hackage.haskell.org/package/polysemy-0.3.0.0/docs/Polysemy-State.html#v:runStateInIORef).

isovector:
To be more explicit about why this doesn't work, polysemy instantiates base's bracket at

```
bracket :: IO (s, a) -> ((s, a) -> IO (s, b)) -> ((s, a) -> IO (s, c)) -> IO (s, c)
```

which is why the `c`-scoped changes don't appear in the `b`-scope, and why the `b`-scope changes get ignored in the result.

isovector:
In #73 we've been chatting about this problem more generally. 

I think that in this particular case, the bug might just be a quirk of the `runState` interpretation of `State` --- "**Note**: this thing does not preserve state changes run outside of the main thread. Use `runStateInIORef` instead if you want those semantics." 

Maybe it's a cop-out, but also maybe this is more of a documentation bug than anything else?

ocharles:
Another `bracket` to consider is that in `exceptions`: https://hackage.haskell.org/package/exceptions-0.10.2/docs/Control-Monad-Catch.html#v:generalBracket

[...]

Issue #80, opened 2019-06-03
How do I write an `async` effect?


ocharles:
I have the following:

```haskell
{-# language BlockArguments #-}
{-# language DataKinds #-}
{-# language FlexibleContexts #-}
{-# language GADTs #-}
{-# language RankNTypes #-}
{-# language ScopedTypeVariables #-}
{-# language TemplateHaskell #-}
{-# language TypeOperators #-}

module Polysemy.Async where

import Control.Concurrent.Async ( Async, async )
import Control.Monad.IO.Class
import Polysemy
import Prelude


data Concurrent m a where
  Fork :: m a -> Concurrent m ( Async a )


makeSem ''Concurrent


runWithConcurrency
  :: forall r a.
     Member ( Lift IO ) r
  => ( forall x. Sem r x -> IO x )
  -> Sem ( Concurrent ': r ) a
  -> Sem r a
runWithConcurrency toIO =
  interpretH \( Fork m ) -> do
    fm <- runT m

    let run :: forall a. Sem ( Concurrent ': r ) a -> IO a
        run = toIO .@ runWithConcurrency
    
    fmap _ ( liftIO ( async ( run fm ) ) )
```

But I'm stuck at the final hole:

```
Found hole: _ :: Async (f a1) -> f (Async a1)
```

I need to distribute `Async` over `f`, but I can't see how to do that.

ocharles:
I think the only sensible thing we can do semantically here is to just discard the state changes in the child thread. But this is not obvious behaviour to calling code. TBH, it's part of the reason I like `MonadUnliftIO` - it essentially removes the option of using stateful monads with `IO` (instead you would propagate state via a mutable reference).

isovector:
**In general, you can't.** Recall that `f` corresponds to "all of the effects that might have been run" -- for example, what would you expect to be the behavior of this:

```haskell
put False
fork $ put True
get
```

as you point out, this indeed requires a mutable reference. **You can instead implement it via the `getInspectorT` machinery**, and then wrap it back into `f` via `pureT`, while delegating the statefulness into `runStateInIO`.

But yes, you make a good argument that these semantics aren't obvious to calling code. I'm not sure what the right thing to do here is, but I'd love any ideas you have.

ocharles:
In `simple-effects`, each effect specifies a class of transformers it can lift through:

```haskell
class Effect e where
  type CanLift e (t :: (* -> *) -> * -> *) :: Constraint
```

[...]

PR #124, opened 2019-06-20
GlobalState that safely runs under bracket

isovector:
This is a possible solution for #84. It's a `State`-like effect that can be deferred until the very end of the day, ensuring it can be run _underneath_ things like `Resource` and `Async`.

The idea is essentially to keep a big map around that you can read from and write to. In IO this is implemented as an `unsafePerformIO`-scoped `IORef` that is atomically updated. Purely, it's just a regular `State` effect.

Elements added to this map have their lifetimes managed via the `allocate` effect, during which you may use `getGlobal` and `putGlobal` on the resulting key. There's a small amount of type-level protection to prevent the keys from escaping their scope, but @Lysxia points out that it's not foolproof.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -90,7 +90,7 @@ tests:
     - hspec-discover >= 2.0
     dependencies:
     - polysemy
-    - inspection-testing >= 0.4.2.1 && < 0.5
+    - inspection-testing >= 0.4.2 && < 0.5
     - hspec >= 2.6.0 && < 3
     - doctest >= 0.16.0.1 && < 0.17
 
--- a/polysemy-plugin/package.yaml
+++ b/polysemy-plugin/package.yaml
@@ -43,7 +43,7 @@ tests:
     - polysemy-plugin
     - hspec >= 2.6.0 && < 3
     - should-not-typecheck >= 2.1.0 && < 3
-    - inspection-testing >= 0.4.2.1 && < 0.5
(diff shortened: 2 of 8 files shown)


isovector:
@lexi_lambda @gelisam @ocharles @jhenahan I'd love for any feedback on this idea.

jhenahan:
This looks really promising. Are the interpreters not inlined intentionally?

ocharles:
I can't say I'm filled with joy, though I haven't take a detailed look at it. I understand it solves the problem, but these days I'm so far convinced that the fundamental principle of `MonadUnliftIO` is much cleaner way of working (which is basically saying `StT t ~ Identity`, as we've established else where).

isovector:
> Are the interpreters not inlined intentionally?

No, it's just boilerplate I didn't want to do until i was convinced this is merge-worthy.

isovector:
> I can't say I'm filled with joy, though I haven't take a detailed look at it. I understand it solves the problem, but these days I'm so far convinced that the fundamental principle of MonadUnliftIO is much cleaner way of working

It doesn't tickle my toes either --- when testing this I tried to put a `trace` in the cleanup action, but to no avail. That being said, as mentioned in #80, I don't think the MonadUnliftIO approach is feasible, so we need some other idea.

isovector:
#126 strikes me as a more elegant solution to this problem.

isovector:
This thing sucks.

PR #125, opened 2019-06-21
Create loopbreakers for inlining recursive calls after renaming

### LINKED ISSUE #91: InlineRecursiveCalls doesn't appear to help the simplifier
The exploration in https://github.com/isovector/polysemy/pull/90 shows that even though the plugin is generating the manually loop-broken definitions, the simplifier refuses to use them. I'm not sure why, but my guess is that some sort of analysis has already been performed by the time the core plugin runs.

A potential solution is to implement this as a `parsedResultAction` plugin instead of as a coretodo. The interesting bits appear to be under `hpm_module > hsmodDecls > ValD > FunBind > fun_matches` , where we'd need to run the existing replace logic.

It's not clear how to check if we have the `Polysemy` dependency we want to be imported.

TheMatten:
This PR strives to resolve #91.

- [x] Find out why test currently fails (https://gitlab.com/snippets/1868163) and how it actually affects generated Core
- [ ] Fix TODOs


TheMatten:
Sorry - I later realized that it is actually not going to be used with versions older than 8.6 (because of plugin interface changes).

polysemy-plugin/src/Polysemy/Plugin.hs
```diff
@@ -61,51 +61,34 @@ module Polysemy.Plugin
...
 plugin = defaultPlugin
-    { tcPlugin = const $ Just fundepPlugin
-    , installCoreToDos = const installTodos
+    { tcPlugin            = const $ Just fundepPlugin
 #if __GLASGOW_HASKELL__ >= 806
-    , pluginRecompile = const $ pure NoForceRecompile
+    , pluginRecompile     = purePlugin
+    , renamedResultAction = \_ env group ->
+        traverse inlineRecCallsAction (env, group)
+    , installCoreToDos    = const installTodos
```
isovector: You can still do this on older versions

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+      pure $ case maybeReplaceVarNames fun_name loopb_name fun_matches of
+        Just fun_matches' -> (,)
+          ( Recursive
+          , listToBag
+              [ L fun_loc fun_bind{ fun_matches = fun_matches' }
+              , loopb_decl
+              ]
+          )
+          (  [ inline alwaysInlinePragma fun_name
+             , inline neverInlinePragma  loopb_name
```
isovector: Neverinline doesn't do what you'd expect! Try this one out for size:

```haskell
noInlinePragma :: InlinePragma
noInlinePragma = defaultInlinePragma { inl_inline = NoInline }
```

polysemy-plugin/src/Polysemy/Plugin.hs
TheMatten: `installCoreToDos`? Is it useful with `renamedResultAction` absent?

isovector: It's always useful, but only on GHC >= 8.10 --- maybe we could make that more clear here.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Thank you so much! :smile:

polysemy-plugin/src/Polysemy/Plugin.hs
TheMatten: Should I move it outside and add `-- Passes to get more out of 8.10 specializing`?

TheMatten:
Should I restrict test on `>=806`?

isovector:
Good idea about restricting the test. It's annoying, but we're already in a situation where different versions of GHC are going to have dramatically different performance, so what's one more?

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
isovector: I think this is a bug in GHC. I'll send a patch.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+  ( if optLevel dyn_flags > 0
+      then inlineRecCalls
+      else pure
+    ) group
+
+------------------------------------------------------------------------------
+inlineRecCalls :: MonadUnique m => HsGroup GhcRn -> m (HsGroup GhcRn)
+inlineRecCalls group@(hs_valds -> XValBindsLR (NValBinds binds sigs)) = do
+
+  (binds', extra_sigs) <- second concat . unzip
```
isovector: I don't understand why you're returning all the `binds'` on the left, but only the *new* sigs on the `right`?

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+
+  ( if optLevel dyn_flags > 0
+      then inlineRecCalls
+      else pure
+    ) group
+
+------------------------------------------------------------------------------
+inlineRecCalls :: MonadUnique m => HsGroup GhcRn -> m (HsGroup GhcRn)
+inlineRecCalls group@(hs_valds -> XValBindsLR (NValBinds binds sigs)) = do
+
```
isovector: nit: no need for this line

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+------------------------------------------------------------------------------
+-- | Forces compiler to inline functions by creating loopbreaker with
+-- NO_INLINE pragma, changing recursive calls to use it and by adding INLINE
+-- pragma to the original function.
+inlineRecCallsAction :: (MonadUnique m, HasDynFlags m)
+                     => HsGroup GhcRn -> m (HsGroup GhcRn)
+inlineRecCallsAction group = do
+  dyn_flags <- getDynFlags
+
+  ( if optLevel dyn_flags > 0
```
isovector: maybe `bool inlineRecCalls pure (optLevel dyn_flags > 0) group`?

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+    ) group
+
+------------------------------------------------------------------------------
+inlineRecCalls :: MonadUnique m => HsGroup GhcRn -> m (HsGroup GhcRn)
+inlineRecCalls group@(hs_valds -> XValBindsLR (NValBinds binds sigs)) = do
+
+  (binds', extra_sigs) <- second concat . unzip
+                      <$> traverse inlineRecCall (pairTypes binds sigs)
+
+  pure group{ hs_valds = XValBindsLR $ NValBinds binds' $ sigs ++ extra_sigs }
```
isovector: Maybe a bidirectional pattern sig for `XValBindsLR . NValBinds` would clean this up and the view pattern above.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+
+------------------------------------------------------------------------------
+-- | Takes binding group and type of binding inside if there's only one and
+-- returns group with inlining of recursive calls and signatures to be added
+-- to the environment.
+inlineRecCall :: MonadUnique m
+              => ((RecFlag, LHsBinds GhcRn), Maybe (LHsSigWcType GhcRn))
+              -> m ((RecFlag, LHsBinds GhcRn), [LSig GhcRn])
+inlineRecCall (b, mtype)
+  | (Recursive, (bagToList -> [
```
isovector: I find this guard super hard to read.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+  noLoc $ TypeSig NoExt [noLoc loopb_name] fun_type
+
+------------------------------------------------------------------------------
+inline :: InlinePragma -> IdP GhcRn -> LSig GhcRn
+inline how name = noLoc $ InlineSig NoExt (noLoc name) how
+
+------------------------------------------------------------------------------
+-- | Returns value with every variable expression id replaced or 'Nothing' if
+-- there's none to be replaced.
+maybeReplaceVarNames :: Data a => Name -> Name -> a -> Maybe a
```
isovector: Why combine these two things in one rather than doing a separate pass for checking recursion? Perf worries?

isovector: Aren't we guaranteed it will be `Just` if the binding group is `Recursive` and only contains one element?

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+  pure group{ hs_valds = XValBindsLR $ NValBinds binds' $ sigs ++ extra_sigs }
+
+inlineRecCalls _ = error "inlineRecCalls: expected renamed group"
+
+------------------------------------------------------------------------------
+-- | Pairs single binds with their types if they have one specified.
+pairTypes :: [(RecFlag, LHsBinds GhcRn)]
+          -> [LSig GhcRn]
+          -> [((RecFlag, LHsBinds GhcRn), Maybe (LHsSigWcType GhcRn))]
+-- TODO: better way of doing this would be great
```
isovector: Build a map and then do a lookup :)

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+          ( Recursive
+          , listToBag
+              [ L fun_loc fun_bind{ fun_matches = fun_matches' }
+              , loopb_decl
+              ]
+          )
+          (  [ inline alwaysInlinePragma                           fun_name
+             , inline defaultInlinePragma{ inl_inline = NoInline } loopb_name
+             ]
+          ++ maybeToList (loopbreakerSig loopb_name <$> mtype)
```
isovector: What happens if this is `Nothing` --- we just get no sig? Does that work?

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
```diff
@@ -1,91 +1,153 @@
...
+    _                 -> Nothing
+ where
+  go :: HsExpr GhcRn -> (Any, HsExpr GhcRn)
+  go (HsVar _ (L loc name))
+    | name == from = (Any True, HsVar NoExt $ L loc to)
+  go e             = (Any False, e)
+
+------------------------------------------------------------------------------
+findMapPop :: Monad m => (a -> Maybe b) -> StateT [a] m (Maybe b)
+findMapPop = state . go where
```
isovector: Just build a map! This thing has `O(n^2)` asymptotics, compared to `O(n log n)`.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Yes, thanks.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Made it and then wondered if it's not overkill for one function, but maybe not :smile:

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Yes
Good point - but should I separate logic for case of one and multiple bindings?

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: I'm probably just overspeculating things :slightly_smiling_face:

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: If there was no sig to begin with, then it's probably infererable and so we shouldn't have to worry - plus we don't really have other options in renamer :smile:

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Will do.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Yes
Good point :slightly_smiling_face:

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
isovector: Fair enough! Maybe add a comment explaining that!

TheMatten: Okay.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Moved pattern matching on `FunBind` to guard.

polysemy-plugin/src/Polysemy/Plugin/InlineRecursiveCalls.hs
TheMatten: Ok.

TheMatten:
Inlining will be separated to the `loopbreakers` library, closing.

PR #126, opened 2019-06-23
Message Passing as a potential solution to async

isovector:
Fundamentally, running state under async isn't really all that tractable. It's obvious how to fork state, but joining it has no meaning in general. So I started thinking about how distributed computing solves these problems, and then answer is that everyone has their own state and they use message passing to synchronize the bits they need.

This is an especially interesting idea in effects systems, where your effects are already the only messages you need. I'm imagining a sort of dispatch server, which is responsible for actually running all of the effects. This thing corresponds to the big `runM . runWhatever` stacks that exist today. However, the Async effect doesn't take a natural transformation `Sem r ~> IO` --- instead, it just implements every effect by message passing it to the dispatch server.

~Effect handlers now have the type `(e x -> Sem r (Sem r x)) -> Sem (e ': r) a -> Sem r a`, where the outside `Sem r` corresponds to effects you want to run on the dispatch server, and the inner `Sem r` are effects you want to run on the forked computations.~ Things like `State` want to be run on the dispatch server, but things like big blocking call in `IO` should be run asynchronously.

This elegantly solves the problem of being unable to run Async, and of state getting lost in bracket.
---

TODO:

* [ ] Haddock!!!
* [x] ~Currently all `IO` runs on the thread it originates on, and all other effects are passed back to the master thread. The API of `Response` allows the master to pass back `IO` to be run on the slave, which is good for effects that want to be interpreted via blocking. This functionality would need to be exposed (possibly via `Tactics`?) to the interpreters.~ Fuck it. Stretch goal.
* [x] Right now the `InChan` isn't very polymorphic and only can be used at a single `r`. I'd need some sort of weakening machinery for this to be used for multiple effects that want to do things outside the main thread of execution.
* [x] How can we get the master thread up and running automatically?
* [x] ~Effects like `Reader.local` probably are fine to run in the originating thread. Is it possible to tag only certain effects that should be run in a different thread? Possibly change the kind of effects to `(par :: MONAD) -> (seq :: MONAD) -> Type -> Type`, and then only send computations in `par` over the dispatch boundary.~ This breaks the principle of local effects, so no worries.


src/Polysemy/Dispatch.hs
```diff
@@ -0,0 +1,115 @@
...
+
+
+data Request r end = forall a. Request
+  { responseMVar :: MVar (Sem '[end] a)
+  , request      :: Union r (Sem r) a
+  }
+
+dispatchEverything
+    :: ( LastMembers end r
+       , Member (Lift IO) '[end]
```
googleson78: Curiousity:

Is this equivalent to
`Lift IO ~ end`
?
Or is `Member` required for some custom type error or `Find` and `IndexOf` machinery?

src/Polysemy/Internal/Union.hs
```diff
@@ -232,4 +237,20 @@ decompCoerce (Union p a) =
...
+class LastMembers end effs | effs -> end where
+  decompLast
+      :: Union effs m a
+      -> Either (Union effs m a) (Union '[end] m a)
+
+instance {-# OVERLAPPABLE #-} LastMembers end effs => LastMembers end (eff ': effs) where
+  decompLast (Union SZ u)     = Left $ Union SZ u
+  decompLast (Union (SS n) u) = first weaken $ decompLast $ Union n u
+
+instance LastMembers end (end ': '[]) where
```
googleson78: Curiosity:

Is it possible to use `'[end]` instead of `(end ': '[])` here?

src/Polysemy/Dispatch.hs
isovector: `Member` is the thing that actually lets `send` work. It's yucky, but I don't think it's a game stopper.

src/Polysemy/Internal/Union.hs
isovector: Probably!

src/Polysemy/Dispatch.hs
isovector: When I first wrote this `LastMembers` would match any suffix of `r`, thus the name and weird style around it. It ended up not working for reasons I didn't want to debug last night, but I'm not convinced it's a fundamental limitation.

isovector:
Split this into #128, #129 and #130. Closing.

PR #128, opened 2019-06-24
Forklift interpretations

isovector:
#126 is finished, but big. Instead I'm splitting it up into a few PRs.

@ocharles pointed out that my brilliant idea for solving the "run state underneath IO effects" solution already has a name --- [the forklift](https://apfelmus.nfshost.com/blog/2012/06/07-forklift.html).

This PR provides a single function `withLowerToIO`, which runs a desired `Sem r` effect all the way down to `IO`, without needing to know the natural transformation beforehand. It does it by running the desired code in a new thread, and shipping all of the unhandled effects back to the main thread. The main thread turns into an event loop for the duration of the `withLowerToIO` block.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -24,6 +24,8 @@ dependencies:
 - th-abstraction >= 0.3.1.0 && < 0.4
 - transformers >= 0.5.2.0 && < 0.6
 - first-class-families >= 0.5.0.0 && < 0.6
+- unagi-chan >= 0.4.0.0 && < 0.5
+- async >= 2.2 && < 3
 
 default-extensions:
   - DataKinds
@@ -90,7 +92,7 @@ tests:
     - hspec-discover >= 2.0
     dependencies:
     - polysemy
-    - inspection-testing >= 0.4.2.1 && < 0.5
+    - inspection-testing >= 0.4.2 && < 0.5
     - hspec >= 2.6.0 && < 3
     - doctest >= 0.16.0.1 && < 0.17
 
--- a/polysemy-plugin/package.yaml
+++ b/polysemy-plugin/package.yaml
@@ -43,7 +43,7 @@ tests:
     - polysemy-plugin
     - hspec >= 2.6.0 && < 3
     - should-not-typecheck >= 2.1.0 && < 3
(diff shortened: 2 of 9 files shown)


googleson78:
Questions:
* How does this play with the "zero-cost abstraction" part of the library? Is it still true?
* How is a user able to know when threads are being spawned, other than by looking through `polysemy` implementation?

ocharles:
> How does this play with the "zero-cost abstraction" part of the library? Is it still true?

Thanks for bringing this up, this was going to be one of my next comments. I'd be curious to know this too. My gut is naturally "there is a cost", but maybe GHC has more surprises.

isovector:
> How does this play with the "zero-cost abstraction" part of the library? Is it still true?

There's definitely a cost here. I'm not very worried about it. The claim is that the machinery is zero-cost, not the interpretations.

> How is a user able to know that threads are being spawned, other than by looking through polysemy implementation?

Who cares, so long as the semantics are correct?

googleson78:
> > How is a user able to know that threads are being spawned, other than by looking through polysemy implementation?
> 
> Who cares, so long as the semantics are correct?

I haven't actually tried this stuff out myself, so my worries could just be a lack of understanding but:

Users running without `-threaded` who also do blocking FFI calls [need to care](https://gitlab.haskell.org/ghc/ghc/wikis/commentary/rts/scheduler#haskell-threads). For an `Async` you would know beforehand to be careful, but for `runResourceBase` it might not be so obvious. (?)

isovector:
I'm happy to add a comment to `runResourceBase`!

mitchellwrosen:
Hello! Randomly chiming in here again:

```
Users running without -threaded who also do blocking FFI calls need to care.
```

I don't see the issue here. Sure, it is uncommon for today's libraries to have any sort of background thread or inter-thread communication, but I think that may be a symptom of immaturity of our library ecosystem for writing concurrent applications. Contrast ours to, say, erlang's, which has many commonly used libraries that spawn background gen_servers and provide some sort of mini service.

But to your point, some care needs to be taken by users. I think a comment in the library's documentation would suffice.

PR #129, opened 2019-06-24
Async effect

### LINKED ISSUE #80: How do I write an `async` effect?

isovector:
#126 is finished, but big. Instead I'm splitting it up into a few PRs.

Here's an `Async` effect that does exactly what you'd expect.

Fixes #80

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ stack.yaml.lock
 core-html/
 *~
 *.prof
+.ghc.environment.*
+dist-newstyle
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,8 @@ cabal: "2.4"
 
 matrix:
   include:
-    - ghc: "8.6.3"
-    # - ghc: "8.4.4"
-    # - ghc: "8.2.2"
-    # - ghc: "8.0.2"
-    # - ghc: "7.10.3"
-    # - ghc: "7.8.4"
-    # - ghc: "7.6.3"
+    - ghc: "8.6.5"
+    - ghc: "8.4.4"
 
 install:
  - cabal --version
@@ -34,3 +29,7 @@ script:
 # `cabal install --force-reinstalls dist/*-*.tar.gz`
  - SRC_TGZ=$(cabal info . | awk '{print $2;exit}').tar.gz &&
    (cd dist && cabal install --force-reinstalls "$SRC_TGZ")
+
+ - cabal v2-build polysemy-plugin
+ - cabal v2-test --enable-test polysemy-plugin
+
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -84,3 +84,6 @@
 
 ## Unreleased changes
 
+- Lots of hard work on the package and CI infrastructure to make it green on
+    GHC 8.4.4 (thanks to @jkachmar)
+
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -86,4 +86,6 @@
 
 - Lots of hard work on the package and CI infrastructure to make it green on
     GHC 8.4.4 (thanks to @jkachmar)
+- runTraceAsList
+- New effect: Async
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 # polysemy
 
-[![Build Status](https://api.travis-ci.org/isovector/polysemy.svg?branch=master)](https://travis-ci.org/isovector/polysemy)
+[![Build Status](https://api.travis-ci.org/polysemy-research/polysemy.svg?branch=master)](https://travis-ci.org/polysemy-research/polysemy)
 [![Hackage](https://img.shields.io/hackage/v/polysemy.svg?logo=haskell&label=polysemy)](https://hackage.haskell.org/package/polysemy)
 [![Hackage](https://img.shields.io/hackage/v/polysemy-plugin.svg?logo=haskell&label=polysemy-plugin)](https://hackage.haskell.org/package/polysemy-plugin)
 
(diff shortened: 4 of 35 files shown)


ocharles:
> Here's an Async effect that does exactly what you'd expect.

Well, almost :wink: Not happy about the `Maybe`, but we're making good progress. 

I was thinking about [`ExitCase`](https://hackage.haskell.org/package/exceptions-0.10.2/docs/Control-Monad-Catch.html#t:ExitCase) from `exceptions`, I wonder if we should use something like that rather than `Maybe`, for a little more clarity.

* `Just` -> `ExitCaseSuccess`
* `Nothing` -> `ExitCaseAbort`

But I suppose the type is isomorphic to `Maybe`, because we aren't catching exceptions (`async` already has machinery for that)

isovector:
I'm not a big fan of those sorts of types; they're invariably missing all of the useful combinators. Consider `Result` from `aeson` which is missing `Result a -> Either String a` and backwards, and `Result a -> Maybe a`, and everyone just ends up writing the functions they want anyway and it's a terrible experience.

I could be talked into a comment, though!

ocharles:
No strong feelings, but I do think it does help users WRT documentation, as it really spells out what is happening.

isovector:
That's a reasonable argument. Let's see where the reviewers get stuck in this mess of a PR and see what should be clarified.

ocharles:
I have a fairly serious reservation about this as a "solution" to `async` (in general) - it complete serialises threads (if I understand it correctly). Let's say I have a basic server that accepts connections. The master thread simply sits there doing a `accept` on a socket, and for each new connection, forks a thread. What the server does is immaterial to this example, but assume we have one very slow client (perhaps being [purposefully slow](https://en.wikipedia.org/wiki/Slowloris_(computer_security))). Whenever I `read` from a client in a forked thread, that read gets pushed down into the `master` thread, because that is where *all* effect handling occurs. So if a `client` is sending me data veeeery slowly, one client seems to take the entire program down.

In fact, I have a suspicious that just two threads using MVars could deadlock under this behavior, because the serialization is eager with no retrying/fairness.

isovector:
@ocharles this is a concern, but can be worked around.

Any IO actions that already exist in in the monad will be run on the originating thread. For example, consider an effect `Delay :: Int -> Delay m a` whose interpretation is `threadDelay`.

* `runAsync . runDelay` <-- delaying will occur in the forked threads
* `runDelay . runAsync` <-- delaying will occur in the master thread

ocharles:
Ok, that's a good start. Where does the `threadDelay` in `async (liftIO (threadDelay 1000))` run?

isovector:
Runs in async.

ocharles:
Ok, I had some time to muse on this and - now that I have a better understanding - I like it more. It's certainly nice that runState . runAsync is different to runAsync . runState. My initial confusion was that _everything_ ran in the parent, but I understand now that it's everything that's still to be handled. The delay stuff is actually really interesting - you get thread local delay or _global_ delay, something we've never really got 'for free'

However, I find the special treatment of IO a bit ugly - it would be great to just be able to say runAsync . runIO to have thread local IO. Also. Instead of returning Maybe in the Async, I wonder if we can just throw an exception into that thread. I might be able to think more on this later

isovector:
I like the idea of `runIO`, but I don't know how to make it happen. What's really going on under the hood is that `LastMember end r` will let you decompose a `Union r m a` into `Either (Union '[end] m a) (Union r m a)`.

spacekitteh:
So how does one actually use this for concurrency?

I have:

```haskell
sequenceConcurrently :: forall t r a. (Traversable t, Member Async r) => 
    t (Sem r a) -> Sem r (Maybe (t a))
sequenceConcurrently t = fmap DT.sequence $ DT.sequence $ fmap f t where
    f :: Sem r a -> Sem r (Maybe a)
    f a = do
        waiting <- async a
        await waiting
```
And in IO:
```Haskell
    ...
    (Just critters) <- runM
        $ asyncToIO
        $  runAtomicStateTVar var
        $  runCacheWithHashMapAsAtomicState
        $ sequenceConcurrently
        $ fmap (\seed -> runRandomSource seed $ doRepros var k config) seeds
    ...
```
I've tried a dozen permutations of interpreter orderings, with both `asyncToIO` and `asyncToIOFinal`, but I cannot figure how to get it to actually run concurrently.

ocharles:
Just to check, are you running with `-threaded` and `+RTS -N`?

spacekitteh:
> Just to check, are you running with -threaded and +RTS -N?

Yep. What I have with the `mapConcurrently` from the `async` package works, but I'd rather refactor it to use Polysemy's `Async`.

KingoftheHomeless:
```haskell
waiting <- async a
await waiting
```
Any time you're spawning an `async`ed thread, you're immediately `await`ing upon it afterwards. The result is that, effectively, the code is synchronous.

The fix is to wait until all thread are spawned before `await`ing upon them. This can be done like this:

```
sequenceConcurrently :: forall t r a. (Traversable t, Member Async r) => 
    t (Sem r a) -> Sem r (Maybe (t a))
sequenceConcurrently t = traverse async t >>= traverse await
```

spacekitteh:
Huh, weird. I actually tried that, but I must have gotten my ordering wrong with the interpreters somewhere. I tried again with a different order, and it works. Thanks!

PR #131, opened 2019-06-25
Better runMonadicInput type order

### LINKED ISSUE #122: runMonadicInput has unhelpful type order
We'd like to be able to type apply the Input, but instead it type applies the monad or the r or something stupid.  Check output too.

tempname11:
This was so disturbingly trivial, I feel like cheating :)

Fixes #122


isovector:
Love, it! Thanks so much

PR #138, opened 2019-06-28
New error combinators

isovector:
Helpful things for using errors in real life


ocharles:
Just for comparison, unliftio calls these fromEither and fromEitherM

isovector:
Thanks for the heads up. I'll rename these for confluence!

PR #140, opened 2019-06-29
Add runOutputAsList and a test for it

### LINKED ISSUE #139: runOutputAsList
This would be nice, intead of `runFoldMapOutput (:[])`

googleson78:
Closes #139

### DIFF
--- a/src/Polysemy/Output.hs
+++ b/src/Polysemy/Output.hs
@@ -8,11 +8,13 @@ module Polysemy.Output
   , output
 
     -- * Interpretations
+  , runOutputAsList
   , runFoldMapOutput
   , runIgnoringOutput
   , runBatchOutput
   ) where
 
+import Data.Bifunctor (first)
 import Polysemy
 import Polysemy.State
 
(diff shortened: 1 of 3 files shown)


isovector:
Fantastic, thanks! Mind changing this to use the new function while you're at it? https://github.com/polysemy-research/polysemy/blob/master/src/Polysemy/Trace.hs#L66

jhenahan:
Is there a performance reason to write it this way instead of just as `runFoldMapOutput pure`? Then you can call it `runOutputMonoid` or something. Also, may be worth updating the README example to use this.

googleson78:
@jhenahan Well, `runFoldMapOutput` does `<>` to the right which seems like O(n^2) to me - I'm not sure if this is subject to fusion, so I just wrote it like this directly.

@isovector Will do. 

Slightly off-topic but:
Why can't `ghc` figure out what `output`s type should be, if I remove the `@Int`? Is there any way to tell `ghc` that `Member (Output o0) '[Output Int]` implies `o0 ~ Int` (I'm guessing this has something to do with injectivity)?

isovector:
@googleson78 you get this inference when `polysemy-plugin` is enabled, though we can't enable it in the main project for dependency reasons.

PR #144, opened 2019-06-30
Use type-errors for our type errors!

isovector:
This PR has farmed out the type error machinery out to my new package https://github.com/isovector/type-errors. It's much nicer to work with, and a big red diff!


src/Polysemy/Internal/CustomErrors.hs
```diff
@@ -33,8 +39,11 @@ data T1 m a
 
 
 type family IfStuck (tyvar :: k) (b :: k1) (c :: Exp k1) :: k1 where
-  IfStuck T1 b c = b
-  IfStuck a  b c = Eval c
+  -- TODO(sandy): This behavior is wrong when k is not one of these things
+  IfStuck T1                  b c = b
+  IfStuck "NOT_A_REAL_SYMBOL" b c = b
+  IfStuck 9999999999999999999 b c = b
```
googleson78: May I ask what's going on here :smile:?

isovector: Consider the following:

```haskell
type family AlwaysStuck :: Symbol
```

when we don't have the `IfStuck "NOT_A_REAL_SYMBOL"` case, the following:

```haskell
type Bug = IfStuck AlwaysStuck (TypeError (Text "Stuck!")) (Pure (() :: Constraint))

foo :: Bug => ()
foo = ()
```

somehow the type family evaluates to `() :: Constraint`, and therefore `foo` doesn't the `Stuck!` type error. But `AlwaysStuck` is, as it implies, *always stuck.*

What's going on here? *The kinds matter.* `IfStuck` won't even compare against the `T1` or the `9999999999999` cases, because neither of those things has kind `Symbol`. And so it will just boop out the `IfStuck a b c = Eval c` case!

googleson78: Does this imply you need a case for every kind you want to use with `IfStuck`? If so is there a userland (ie without hacking ghc) way to avoid this? (thanks for taking the time to expain this! <3)

isovector: This does imply such a thing, unfortunately. I was thinking about writing a library that does everything it can --- all polykinded type constructors up to 64 arguments, likewise for constrainst, and then all of the promoted kinds in base (including tuples). But the limitation of any data kind that isn't in that set makes me hesitant that it'll go wrong sooner than later :(

isovector: This discussion lead to https://github.com/isovector/type-errors

PR #145, opened 2019-06-30
Use new, separate `loopbreaker` library for inlining

TheMatten:
Red diffs, hooray! :tada:

TODO:
- [x] GHC 8.4 (guard use of plugin)


isovector:
Hooray!!!

package.yaml
```diff
@@ -69,7 +72,7 @@ library:
         - dump-core
       ghc-options:
         - -fplugin=DumpCore
-        - -fplugin-opt DumpCore:core-html
+        - -fplugin-opt=DumpCore:core-html
```
isovector: I am reasonably sure this works without the change

polysemy-plugin/src/Polysemy/Plugin.hs
```diff
@@ -100,12 +94,9 @@ installTodos todos = do
     0 -> pure todos
     _ -> do
       mods <- moduleSetElts <$> getVisibleOrphanMods
-      pure $ case any ((== polysemyInternal) . moduleName) mods of
-        True  -> CoreDoPluginPass "Inline Recursive Calls" inlineRecursiveCalls
-               : todos
+      pure $ todos
```
isovector: This change makes `mods` unused in the < `810` case, right? That'll make lint unhappy

src/Polysemy/Writer.hs
```diff
@@ -53,7 +53,7 @@ runWriter = runState mempty . reinterpretH
         modify (`mappend` o) >>= pureT
       Listen m -> do
         mm <- runT m
-        -- TODO(sandy): this is fucking stupid
+        -- TODO(sandy): this is stupid
```
isovector: :angry:

isovector:
This brings true joy to my heart.

package.yaml
TheMatten: I wanted to be consistent and `=` was just my personal preference - should I discard this change?

polysemy-plugin/src/Polysemy/Plugin.hs
TheMatten: Good point, thanks.

package.yaml
isovector: I'm happy with it so long as it works.

polysemy-plugin/src/Polysemy/Plugin.hs
```diff
@@ -93,10 +93,11 @@ installTodos todos = do
   case optLevel dflags of
     0 -> pure todos
     _ -> do
-      mods <- moduleSetElts <$> getVisibleOrphanMods
-      pure $ todos
 #if __GLASGOW_HASKELL__ >= 810
-          ++ bool []
-                  (extraPhases dflags)
-                  (any ((== polysemyInternal) . moduleName) mods)
+      mods <- moduleSetElts <$> getVisibleOrphanMods
```
isovector: Just to ease my mind, did you check this branch to make sure it compiles?

TheMatten: Yes - is there something that seems questionable?

isovector: Nope, nothing! Just being overly cautious!

isovector:
I'm happy to merge when you give me the :+1:

package.yaml
```diff
@@ -43,10 +43,14 @@ default-extensions:
   - UnicodeSyntax
 
 when:
-  - condition: impl(ghc < 8.6)
+  condition: impl(ghc < 8.6)
+  then:
     default-extensions:
       - MonadFailDesugaring
       - TypeInType
+  else:
```
isovector: Let's write this as a separate condition, since it's not actually related to the `else` branch.

TheMatten:
CI passed - :+1:

PR #147, opened 2019-07-04
Parenthesize r when it's a cons

### LINKED ISSUE #74: Weird error message with no commas
I just got:

```
    • Ambiguous use of effect 'State'
      Possible fix:
        add (Member (State FeederId) Error
                                       InsertTraydatasAndTraytypesFailure
                                       : r1) to the context of 
          the type signature
```

I think this is actually telling me to add `( Member ( State FeederId ) r, Member ( Error InsertTraydatasAndTraytypesFailure ) r)`

isovector:
When emitting messages for ambiguous sends: use `IfStuck` on `r` to
determine if we can compose it. This means `r` is in one of three
states, stuck, cons or nil. We want to wrap it in parentheses iff it's
a cons.

Fixes #74

### DIFF
--- a/src/Polysemy/Internal/CustomErrors.hs
+++ b/src/Polysemy/Internal/CustomErrors.hs
@@ -35,7 +35,38 @@ type family DefiningModuleForEffect (e :: k) :: Symbol where
   DefiningModuleForEffect e     = DefiningModule e
 
 
-type AmbigousEffectMessage r e t vs =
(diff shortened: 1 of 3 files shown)


TheMatten:
Now I'm wondering if simply putting quotes around every type wouldn't be enough :slightly_smiling_face:

isovector:
NO!

TheMatten:
Ok then :smile:

Issue #112, opened 2019-06-15
On the type of `runBatchOutput`


Infinisil:
I'm a bit confused from the type of `runBatchOutput`:

```haskell
runBatchOutput :: Int -> Sem (Output [o] ': r) a -> Sem (Output [[o]] ': r) a 
```

Shouldn't this be:
```haskell
runBatchOutput :: Int -> Sem (Output [o] ': r) a -> Sem (Output [o] ': r) a 
```

Because it converts output of variable sizes into outputs of fixed sizes, no idea how another `[]` would make sense here. The implementation also only even outputs lists with a single element in already: https://github.com/isovector/polysemy/blob/b5d086b85999708d1da98f5d3f7aa5f7067bc8a8/src/Polysemy/Output.hs#L74-L83

A bit further, it seems that this type would even be a better fit:
```haskell
runBatchOutput :: Int -> Sem (Output o ': r) a -> Sem (Output [o] ': r) a 
```

Because an `Output o` effects already corresponds to it outputting a series of `o`'s. Converting this series of single events into a series of a bunch of events makes a lot of sense.

Does this make sense or am I missing something?

isovector:
You might be right that the type is wrong. I recall changing it while writing the [test](https://github.com/isovector/polysemy/blob/master/test/OutputSpec.hs); so I might have just been not thinking clearly.

The type `Int -> Sem (Output o ': r) a -> Sem (Output [o] ': r) a` seems too restrictive, however, since it means you can't batch things that already come in a list. I think probably `Int -> Sem (Output [o] ': r) a -> Sem (Output [o] ': r) a` is the right move, since people can always do some sort of `reinterpret $ \case Output o -> output [o]` as an interop.

Infinisil:
> I think probably `Int -> Sem (Output [o] ': r) a -> Sem (Output [o] ': r) a` is the right move, since people can always do some sort of `reinterpret $ \case Output o -> output [o]` as an interop.

In the same spirit, one can also do `reinterpret $ \case Output os -> traverse output os`. With this I don't think there's a reason to choose lists over single elements, list lengths also don't have any influence on anything anyways.

PR #158, opened 2019-07-04
Improve runBatchOutput

Infinisil:
Addresses #112 

This function now transforms an `Output o` into `Output [o]`, which only outputs elements once the specified number of `o`'s have been collected. The only exception is the last element which contains the rest (if any).


isovector:
Excellent, thanks!

Issue #156, opened 2019-07-04
Change ordering of type variable application where appropriate


TheMatten:
Based on #146 (and #122). See comments for possible candidates.

TheMatten:
https://github.com/polysemy-research/polysemy/blob/8bbd9dc7d6ff1d55c9f8ea5f7f160ecf337d0e6e/src/Polysemy/Error.hs#L58-L63
`e r m a` -> `e m r a`

TheMatten:
https://github.com/polysemy-research/polysemy/blob/8bbd9dc7d6ff1d55c9f8ea5f7f160ecf337d0e6e/src/Polysemy/Error.hs#L99-L103
`e2 r e1 a` -> `e1 e2 r a`

TheMatten:
https://github.com/polysemy-research/polysemy/blob/d9dab9fa6acf5fba5c871c9a0e9db55bd4417f4a/src/Polysemy/IO.hs#L56-L62
`r m a` -> `m r a`

TheMatten:
https://github.com/polysemy-research/polysemy/blob/8bbd9dc7d6ff1d55c9f8ea5f7f160ecf337d0e6e/src/Polysemy/Reader.hs#L32
`i r j` -> `i j r`

PR #161, opened 2019-07-05
UX around Lift

### LINKED ISSUE #142: UX around Lift
Let's move `Lift` into `Polysemy.Lift.Type`, and then put `runLift` into `Polysemy.Lift`.

https://github.com/polysemy-research/polysemy-zoo/pull/33#issuecomment-506958946
https://github.com/polysemy-research/polysemy-zoo/pull/33#issuecomment-506968371

googleson78:
Closes #142 

Not sure if I should:

0. Commit the new `.cabal`
1. Bump package version
2. Add a "since" haddock to `runLift`


src/Polysemy/Lift/Type.hs
```diff
@@ -2,7 +2,10 @@
 
 {-# OPTIONS_HADDOCK not-home #-}
```
googleson78: Ah, this isn't supposed to be here anymore, I think!

googleson78: Nevermind, it actually should, I guess, because of the reexports. I need to improve my haddock-fu.

src/Polysemy/Lift.hs
```diff
@@ -0,0 +1,27 @@
...
+
+import           Polysemy
+import           Polysemy.Lift.Type (Lift(..))
+
+makeSem ''Lift
+
+------------------------------------------------------------------------------
+-- | Given a natural transform from @m1@ to @m2@
+-- run a @Lift m1@ effect by transforming it into a @Lift m2@ effect.
+runLift :: Member (Lift m2) r => (forall x. m1 x -> m2 x) -> Sem (Lift m1 ': r) a -> Sem r a
```
isovector: Please add a `forall m1 m2 r a.` before this type; we want to make sure all of our types are easily type-apply-able. See #156 for more details.

src/Polysemy/Lift.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Polysemy.Lift
+  ( -- * Effect
+    Lift (..)
+
+    -- * Actions
+  , lift
```
isovector: Traditionally this has been called `sendM` but admittedly that is a crap name. I'm a little concerned about the overlap with `MonadTrans.trans` --- but I have to say this is nice. Thoughts?

isovector: If we go this way, let's remove `sendM` across the codebase and reexport `lift` from `Polysemy`.

src/Polysemy/Lift.hs
```diff
@@ -0,0 +1,27 @@
...
+import           Polysemy
+import           Polysemy.Lift.Type (Lift(..))
+
+makeSem ''Lift
+
+------------------------------------------------------------------------------
+-- | Given a natural transform from @m1@ to @m2@
+-- run a @Lift m1@ effect by transforming it into a @Lift m2@ effect.
+runLift :: Member (Lift m2) r => (forall x. m1 x -> m2 x) -> Sem (Lift m1 ': r) a -> Sem r a
+runLift f = interpret $ sendM . f . unLift
```
isovector: We can implement `Polysemy.IO.runIO` in terms of this, right? I think we should make that change in this PR too.

src/Polysemy/Lift.hs
googleson78: Ah, I just did this because I was following convention (and was sleepy).

I like `lift` more, because there is no non-`Internal` `send` for `sendM` to be "based off". (like `map` -> `mapM`).

It has the same idea as `MonadTrans.lift` so I don't think that's an issue. (?)

googleson78: I went ahead and did the rename, if you want to use it.

isovector: Thanks for the rename! Upon looking at the changes though, I'm already confused by the name. I think overloading `lift` here is only going to lead to confusion in the library. Can we come up with some other name (possibly also for the type itself?)

googleson78: I don't have any particularly bright ideas:

* `Name` - because you're giving a "name" (syntax) to a thing that already has an interpretation (inspired by logic, where you do the same when you want to refer to an element of your universe with a constant)

* `Concrete/Unambiguous` (or some nice sounding synonym) - because you're not leaving anything "open to interpretation" (in the Syntax vs Semantics sense in which the other "native" effects are open to "interpretation")

isovector:
How about we name the effect `Embed`, with action `embed`?

src/Polysemy.hs
```diff
@@ -10,8 +10,8 @@ module Polysemy
   , runM
 
   -- * Interoperating With Other Monads
-  , Lift (..)
-  , sendM
+  , Embed (..)
+  , embed
 
     -- * Lifting
```
googleson78: Perhaps this also need a better section name (?)

googleson78:
Renamed to `Embed`.

src/Polysemy/Internal.hs
```diff
@@ -300,18 +300,20 @@ raiseUnder3 = hoistSem $ hoist raiseUnder3 . weakenUnder3
...
 
 
 ------------------------------------------------------------------------------
--- | Lift a monadic action @m@ into 'Sem'.
-sendM :: Member (Lift m) r => m a -> Sem r a
-sendM = send . Lift
-{-# INLINE sendM #-}
+-- | Embed a monadic action @m@ in 'Sem'.
+--
+-- TODO(sandy): @since
```
googleson78: Not sure if this actually needs a `@since`, added just in case.

src/Polysemy.hs
isovector: I think it's ok! But keen eye!

src/Polysemy/Embed.hs
```diff
@@ -0,0 +1,29 @@
...
+
+import Polysemy
+import Polysemy.Embed.Type (Embed (..))
+
+------------------------------------------------------------------------------
+-- | Given a natural transform from @m1@ to @m2@
+-- run a @Embed m1@ effect by transforming it into a @Embed m2@ effect.
+--
+-- TODO(sandy): @since
+runEmbed
```
isovector: Let's go with `runEmbedded`

isovector: Is this the same thing that exists in `Polysemy.IO`? This is admittedly a better place for it.

src/Polysemy/IO.hs
```diff
@@ -37,16 +38,16 @@ import Polysemy.Internal.Union
...
+    => Sem (Embed IO ': r) a
     -> Sem r a
-runIO = interpret $ sendM . liftIO @m . unLift
+runIO = runEmbed $ liftIO @m
 {-# INLINE runIO #-}
 
 
 ------------------------------------------------------------------------------
--- | Given some @'MonadIO' m@, interpret all @'Lift' m@ actions in that monad
+-- | Given some @'MonadIO' m@, interpret all @'Embed' m@ actions in that monad
```
isovector: Let's rename this to `runEmbeddedInIO`

src/Polysemy/Internal.hs
isovector: This thing has been around since v0, but I guess it doesn't hurt to be explicit about that! Good catch!

googleson78: I was thinking more does it needs a since for whatever version this will be in, because it's renamed.

src/Polysemy/Embed.hs
googleson78: This is the original `runLift` (now renamed to `runEmbed`, because the effect is `Embed`) from #142, which started this pr.

It's not the same thing as `Polysemy.IO.runEmbedded`.

Personally I'm fine with it being `runEmbed` (in accordance with the other effects), but we need to rename the second one if this is going to be `runEmbedded`.

src/Polysemy/IO.hs
googleson78: Ah, I didn't see this before commenting above.

src/Polysemy/Internal.hs
isovector: Yup, you're right! That's what I get for reviewing just before bed.

googleson78:
Is this ready to go?

isovector:
Thanks for the reminder, it slipped off my radar. Let's get it merged today!

isovector:
Excellent work. Thank you so much!

PR #164, opened 2019-07-07
Don't use Loopbreaker until GHC 8.7+

### LINKED ISSUE #160: Loopbreaker dependency breaks haddock generation
Haddock can't run plugins in 8.6 (https://github.com/haskell/haddock/issues/900 again). We can fix this by doing a 

```haskell
#ifndef __HADDOCK__
{-# OPTIONS_GHC -fplugin=Loopbreaker #-}
#endif
```

on a per module basis, but it'd be nice if we could figure out a way to conditionally enable it on a project-wide basis iff haddock isn't being run.

isovector:
Turns out haddocks have been broken forever, but because of https://github.com/haskell/cabal/issues/5977 we never noticed. This PR fixes CI so it breaks (see 6f915b6), and then fixes it by changing the GHC versions under which we're allowed to use loopbreaker. Fixes #160

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ script:
  - cabal v2-update
  - cabal v2-build
  - cabal v2-test --enable-test
- - cabal new-haddock 2>&1 >/dev/null | ! grep 'Failed to build documentation'
+ - "! cabal new-haddock 2>&1 >/dev/null | grep 'Failed to build documentation'; echo $?"
  - cabal check
  - cabal sdist   # tests that a source-distribution can be generated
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -20,7 +20,7 @@ script:
  - cabal v2-update
  - cabal v2-build
  - cabal v2-test --enable-test
- - cabal new-haddock
+ - "! cabal new-haddock 2>&1 >/dev/null | grep 'Failed to build documentation'"
```
TheMatten: Maybe we can add comment here about haskell/cabal#5977 - we probably want to revert this after it gets fixed.

PR #169, opened 2019-07-08
Add Pass to Writer, replacing Censor. Fix semantics of listen

### LINKED ISSUE #166: Is Writer wrong?
IIRC @adamConnerSax was saying our Writer instance doesn't line up with MTL. Is this true, and if so, let's fix it.

KingoftheHomeless:
This closes #166.

### DIFF
--- a/src/Polysemy/Writer.hs
+++ b/src/Polysemy/Writer.hs
@@ -28,7 +28,7 @@ import Polysemy.State
 data Writer o m a where
   Tell   :: o -> Writer o m ()
   Listen :: ∀ o m a. m a -> Writer o m (o, a)
-  Pass   :: m (a, o -> o) -> Writer o m a
+  Pass   :: m (o -> o, a) -> Writer o m a
 
 makeSem ''Writer
 
@@ -36,7 +36,7 @@ censor :: Member (Writer o) r
        => (o -> o)
        -> Sem r a
        -> Sem r a
-censor f m = pass (fmap (\a -> (a, f)) m)
+censor f m = pass (fmap ((,) f) m)
 
 ------------------------------------------------------------------------------
 -- | Transform an 'Output' effect into a 'Writer' effect.
--- a/src/Polysemy/Writer.hs
+++ b/src/Polysemy/Writer.hs
@@ -67,7 +67,7 @@ runWriter = runState mempty . reinterpretH
         mm <- runT m
         (o, t) <- raise $ runWriter mm
         ins <- getInspectorT
-        let f = maybe id snd (inspect ins t)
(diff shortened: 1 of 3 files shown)


src/Polysemy/Writer.hs
```diff
@@ -27,10 +28,15 @@ import Polysemy.State
 data Writer o m a where
   Tell   :: o -> Writer o m ()
   Listen :: ∀ o m a. m a -> Writer o m (o, a)
-  Censor :: (o -> o) -> m a -> Writer o m a
+  Pass   :: m (o -> o, a) -> Writer o m a
```
KingoftheHomeless: I flipped the elements in the tuple compared to MTL, since polysemy does so for `listen`, `runWriter`, etc. I have to assume that's reasonable, since nobody's going to use `pass` anyway.

KingoftheHomeless:
Note: this will require some refactoring in polysemy-zoo.

adamConnerSax:
Yeah.  I'm happy to fix the absorber stuff.  I should wait until this is merged, right?

src/Polysemy/Writer.hs
isovector: :+1: the original order in mtl is a bug, imo.

isovector:
@adamConnerSax yeah, the zoo is going to need some love. It's going to be a little tricky, as there are a bunch of breaking changes in HEAD right now. If it works for you, one solution would be to send a PR based on `polysemy` HEAD, and I'll figure out how to merge everything on the next release.

adamConnerSax:
Happy to give it all a shot.  But zoo has also had some changes, right?  Should I fork HEAD from zoo (master) and then use cabal.project to specify a specific commit of polysemy itself?

isovector:
Actually, I don't think the other breaking changes will affect the zoo. I'm
not sure of the right way to do this, but I can just push a polysemy
release if it makes everyone's life easier.

adamConnerSax:
I don't mind using cabal.project to use a github version instead of a release.  But I'm not sure how much that complicates your life when merging and releasing.

isovector:
I don't know either. I'll just cut a release!

adamConnerSax:
Re-implemented the ```pass``` function directly in terms of ```pass``` in Polysemy.Writer.  Compiles and tests all pass.  Happy to submit PR.  But before I do, should any of the convo above get turned into more tests that check that the semantics match?
Reassuringly, when I first pulled the zoo, it compiled *but* the absorbWriter test (of the whole stack) failed because my implementation of ```pass``` in terms of ```censor``` did not line up.  So we're catching some things but I'm not sure if anything else ought to be added while I'm here.

PR #170, opened 2019-07-08
Add inlining to Polysemy.Writer

KingoftheHomeless:
I forgot to add those wacky `INLINE` pragmas...

### DIFF
--- a/src/Polysemy/Writer.hs
+++ b/src/Polysemy/Writer.hs
@@ -36,7 +36,8 @@ censor :: Member (Writer o) r
        => (o -> o)
        -> Sem r a
        -> Sem r a
-censor f m = pass (fmap ((,) f) m)
+censor f m = pass (fmap (f ,) m)
+{-# INLINE censor #-}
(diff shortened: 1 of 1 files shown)


isovector:
I wonder if we could convince `Loopbreaker` to insert these automatically. @TheMatten any ideas?

TheMatten:
You mean something like `INLINE` to all functions that manipulate `Sem` type, or all functions in module?

isovector:
Just things that are wrappers around `send`.

isovector:
I think it's a good idea to manually mark your interpreters as `INLINE`, but the actions should technically be `INLINEABLE CONLIKE`, which is more of a mouthful than I want to ask from users. The TH correctly generates the right stuff.

TheMatten:
Generally (and this should be compatible with reimplementation in the future), I would like to have `{-# ANN foo Loopbreaker #-}`, that would just "do it's magic" and add required annotations and stuff.

PR #171, opened 2019-07-08
Fix ambiguously looking variables in smart constructors, refactor Polysemy.Internal.TH.Common

### LINKED ISSUE #157: TH generates signatures with ambiguously-looking names
See https://www.stackage.org/haddock/nightly-2019-07-04/polysemy-0.5.1.0/Polysemy-Resource.html#v:bracket - it looks like `a` appears twice in `forall` - in reality, these are different variables (https://www.stackage.org/nightly-2019-07-04/hoogle?q=bracket+%2Bpolysemy) - `th-abstraction` defaults to names from type constructors instead of data constructors - so `b` in:
```hs
data Resource m a where
  Bracket :: m a -> (a -> m c) -> (a -> m b) -> Resource m b
```
is provided under name `a` - but `a` from `Resource m a`, not from `Bracket`!

While this shouldn't affect any functionality, it's puzzling to the user = bad UX.

TheMatten:
Fixes #157. Plus with changes to how `mkCLInfo` (now `makeCLInfo` for consistency) works, it felt convenient to restructure module a little bit (v1 is coming, so let's make it nice :slightly_smiling_face:).

Current behaviour is a little bit different - instead of aggresively unifying all equalities, we now follow "shape" of user-specified data constructor's type and all type variables now follow names from this signature, instead of type constructor (this is reflected in changes to documentation) - it is result of switching to TH's `reify` for getting needed information, instead of using `th-abstraction`.


test/ThEffectSpec.hs
```diff
@@ -68,24 +68,25 @@ newtype Newtype2 m a where
 makeSem ''Newtype2
 
 -- Data families -------------------------------------------------------------
+-- NOTE: Not supported currently, but may be reconsidered in the future.
```
isovector: Let's just remove this; source control exists for a reason!

src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -101,16 +99,19 @@ makeSem = genFreer True
 -- These restrictions may be removed in the future, depending on changes to
 -- the compiler.
 --
+-- Change in (TODO(Sandy): version): in case of GADTs, signatures now only use
```
isovector: This isn't a breaking change, right? It should only affect haddock, so I don't think we need to do any serious sem versioning.

TheMatten: It's a breaking change for users that quantified variables previously chosen by type constructor - example where explicit signature would be a problem is `get`, which changed from:
```hs
get :: forall a r. MemberWithError (State a) r => Sem r a 
```
to
```hs
get :: forall s r. MemberWithError (State s) r => Sem r s 
```
because data constructor uses `s` instead.

TheMatten:
:slightly_smiling_face:
BTW, regarding comment in code - I will look into it once more later to check if data families would be viable.

TheMatten:
Ok, they should be safe with current code.

src/Polysemy/Internal/TH/Common.hs
```diff
@@ -97,11 +97,6 @@ makeCLInfo cliConName = do
     DataConI _ t p -> pure (t, p)
     _              -> notDataCon cliConName
 
-  -- We do not support type families (polysemy-research/polysemy#157).
```
isovector: Do we now support type families???

test/ThEffectSpec.hs
```diff
@@ -68,25 +68,30 @@ newtype Newtype2 m a where
 makeSem ''Newtype2
 
 -- Data families -------------------------------------------------------------
--- NOTE: Not supported currently, but may be reconsidered in the future.
 
--- data Instance = ADTI | GADTI | NTI
+data Instance = ADTI | GADTI | NTI | MMI
```
isovector: What do these constructors mean?

src/Polysemy/Internal/TH/Common.hs
isovector: Not just _data families_ ??

test/ThEffectSpec.hs
TheMatten: Simple **ADT**, **GADT**, **n**ewtype and "`m` **m**atching" **I**nstances - I just needed some names to put in there - they don't have any deeper meanings

TheMatten: Should I name them simply `I1`, `I2`, etc.?

isovector: Ahh, no, it's fine. Makes sense now!

Issue #168, opened 2019-07-08
Come up with a consistent naming scheme


isovector:
I'm bad at naming things. Consider that these things all reference `IO`, but all have completely different semantics going on.

* `runIO`
* `runTraceIO`
* `runAsyncInIO`
* `runStateInIORef`

And then consider the `Resource` functions:

* `runResourceInIO`
* `runResource`
* `runResourceBase`

??????

and then `Output` is all over the place too:

* `runOutputAsList`
* `runFoldMapOutput`
* `runBatchOutput`

isovector:
I think maybe `interpret`s should be "`run*`", but then `intercept`s should be something else. Maybe "`transform`" or something?

isovector:
It would be good if the operators that work with `.@` all have their own naming scheme. Currently this is `*InIO` but that isn't very descriptive.

isovector:
It seems like there are a few distinct categories of sem transformations here:

1. Things that implement one effect in terms of another, eg `runTraceIO`, `runOutputAsWriter`
2. Things that lower directly to IO, eg `runResourceInIO`, `runAsyncInIO`
3. Things that are functors over their effects, eg `runErrorAsAnther`, `runStateAsAnother`
4. Pure implementations, eg. `runResource` , `runOutputAsList`
5. Void implementations, eg. `runIgnoringTrace`, `runIgnoringOutput`
6. Transforming intercepts, eg `runBatchOutput` (I can't think of any others, but probably there are lots to be written!)

Presumably each group should have a naming convention

TheMatten:
Some ideas:
1. `run_As_`
2. `run_In_`
3. `convert_`/`transform_`/`map_` or maybe even typeclass with e.g. `emap`?
4. They're both pure, but second is somewhat specific in return value - maybe `run_` and `run_To_`
5. `ignore_`
6. `runOutputAsBatching`?

googleson78:
Regarding 5.: 

Is it possible to implement
```
ignore :: Sem (_ ': r) a -> Sem r a
```
🤔

TheMatten:
We would somehow need to "catch" `x` in effect `e m x` - some sort of `e m x -> x` or proof that `x` is `()` (as is technically case with `Output`).

TheMatten:
Basically, you can't come up with result value of action out of thin air, unless it's a singleton - `()`s are fine, but what about:
```hs
do val <- input @Int
   pure $ val + 1
```
What should happen when we ignore `Input`?...

[...]

Issue #165, opened 2019-07-07
Missing interpreters


isovector:
Still need:

* `runMonadicOutput`
* `runOuputAsAnother` -- maybe more generally we want variance relations over our effects??

googleson78:
What are the desired type signatures for these?

isovector:
* `Member (Lift m) r => (o -> m ()) -> Sem (Output o ': r) a -> Sem r a`
* `Member (Output o2) r => (o1 -> o2) -> Sem (Output o1 ': r) a -> Sem r a`

googleson78:
I'll add these around this coming weekend, if nobody else wants to before then.

W.r.t. the variance relations - most of the transformations should be something like (contra-)functors, but I don't immediately see how effects can fit the default `Functor` and `Contravariant`. Perhaps a special version for effects of these? Or just straight up writing all the `runXAsAnother` functions, without being too fancy (avoids increasing library surface area).

isovector:
Back in the long long ago, @shlevy and I wrote this https://github.com/IxpertaSolutions/freer-effects/pull/32 --- essentially new type classes for co/contra/invariant functors over effects. The idea didn't get any use in reality, but I still think it was a nice solution.

isovector:
Let's coordinate against #168 before putting in the engineering legwork --- I'd feel bad if I gave you the renaming runaround on two PRs in a row!

PR #173, opened 2019-07-11
Functors over effect values

googleson78:
I played around with this a bit to see how it would look.

What's peoples opinion on whether this actually brings any benefit  (i.e. some functions that use these classes? - I want to be able to write `ignore` which works over `Effunctors`, but alas, I don't think that is going to be possible :sob: )
vs
just writing them out (this has a clear benefit in my eyes - less surface area for new users)

Related to #165 and #168.

### DIFF
--- a/src/Polysemy/Internal/Effunctor.hs
+++ b/src/Polysemy/Internal/Effunctor.hs
@@ -1,0 +1,16 @@
+module Polysemy.Internal.Effunctor where
+
+import Data.Kind
+import Polysemy.Internal
+import Polysemy.Internal.Kind
+
+-- | Effects that are \"functors\" over their argument.
+--
+-- TODO: add better docs
+class Effunctor (e1 :: Type -> Effect) where
+  effmap
+    :: forall r a b c
+     . Member (e1 b) r
+    => (a -> b)
+    -> Sem (e1 a ': r) c
+    -> Sem r c
--- a/src/Polysemy/Internal/Effunctor.hs
+++ b/src/Polysemy/Internal/Effunctor.hs
@@ -14,3 +14,14 @@ class Effunctor (e1 :: Type -> Effect) where
     => (a -> b)
     -> Sem (e1 a ': r) c
     -> Sem r c
+
+-- | Effects that are \"contravariant functors\" over their argument.
+--
+-- TODO: add better docs
+class Effecontra (e1 :: Type -> Effect) where
+  effcontra
+    :: forall r a b c
+     . Member (e1 a) r
+    => (a -> b)
+    -> Sem (e1 b ': r) c
+    -> Sem r c
--- a/polysemy.cabal
+++ b/polysemy.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 7e0e9242c36e377756ae6ef007d19a16cdaf3f244855b0bfc97e267f4df143ea
+-- hash: 704abe4bacb43ce46510310f8b44aa453854ee080bf3853642c42cc900b58d4f
 
 name:           polysemy
 version:        0.7.0.0
@@ -50,6 +50,7 @@ library
       Polysemy.Internal.Combinators
       Polysemy.Internal.CustomErrors
       Polysemy.Internal.CustomErrors.Redefined
+      Polysemy.Internal.Effunctor
       Polysemy.Internal.Fixpoint
       Polysemy.Internal.Forklift
       Polysemy.Internal.Kind
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -118,6 +118,7 @@ module Polysemy
   , getInspectorT
   , Inspector (..)
   , Effunctor (..)
+  , Effecontra (..)
   ) where
 
 import Polysemy.Internal
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -117,10 +117,12 @@ module Polysemy
   , bindT
   , getInspectorT
   , Inspector (..)
+  , Effunctor (..)
   ) where
 
 import Polysemy.Internal
 import Polysemy.Internal.Combinators
+import Polysemy.Internal.Effunctor
 import Polysemy.Internal.Forklift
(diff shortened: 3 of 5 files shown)


src/Polysemy/Internal/Effunctor.hs
```diff
@@ -0,0 +1,27 @@
...
+-- | Effects that are \"functors\" over their argument.
+--
+-- TODO: add better docs
+class Effunctor (e1 :: Type -> Effect) where
+  effmap
+    :: forall r a b c
+     . Member (e1 b) r
+    => (a -> b)
+    -> Sem (e1 a ': r) c
+    -> Sem r c
```
isovector: I would expect this to be `Sem (e1 b ': r) c`

isovector:
I'm in favor of this. Especially if we can TH-generate the functor instances!

isovector:
@totbwf can your tactics stuff write these automatically for us yet? <3

src/Polysemy/Internal/Effunctor.hs
googleson78: I mirrored
```haskell
runOuputAsAnother :: Member (Output o2) r => (o1 -> o2) -> Sem (Output o1 ': r) a -> Sem r a
```
from [165](https://github.com/polysemy-research/polysemy/issues/165#issuecomment-509036665) for the type. What are the pros/cons of explicitly putting it in the front? (type inference/error messages/internal machinery/performance issues?)

isovector: Good point. That's a very motivating example! Explicitly putting it at the front would mean we can't solve the [trouble with typed errors](https://www.parsonsmatt.org/2018/11/03/trouble_with_typed_errors.html)

googleson78:
Do you have any ideas for other "functors" that would be common?

I will try to look into generating instances for these two with TH but:

0. I'm not 100% sure what the necessary and sufficient conditions for these are. (so I would appreciate any pointers/links)
1. I'm not experienced with TH, so it might take me a while to get it rolling when I'm done with 0.

src/Polysemy/Internal/Effunctor.hs
Infinisil: I seem to often write types like

```haskell
data Action r = forall s . Action
  { startingValue :: s
  , action        :: Int -> Sem (State s ': r) ()
  }

printAction :: Member Trace r => Action r
interactiveSetAction :: Member (Lift IO) r => Action r
```

And then I want `Action r` to be a `Semigroup` which runs both actions with their respective state:

```haskell
instance Semigroup (Action r) where
  (Action s a) <> (Action s' a') = Action (s, s') $ \value -> do
    mapStateFirst' (a value)
    mapStateSecond' (a' value)

mapStateFirst' :: Sem (State a ': r) x -> Sem (State (a, b) ': r) x
mapStateSecond' :: Sem (State b ': r) x -> Sem (State (a, b) ': r) x
```

For this case, these functions won't work:
```haskell
mapStateFirst :: Member (State (a, b)) r => Sem (State a ': r) x -> Sem r x
mapStateSecond :: Member (State (a, b)) r => Sem (State b ': r) x -> Sem r x
```

Failing with
<details>

```
    • Occurs check: cannot construct the infinite type:
        r ~ State (s, s1) : r
      Expected type: Sem (State (s, s1) : r) ()
        Actual type: Sem r ()
    • In a stmt of a 'do' block: mapStateFirst (a value)
      In the expression:
        do mapStateFirst (a value)
           mapStateSecond (a' value)
      In the second argument of ‘($)’, namely
        ‘\ value
           -> do mapStateFirst (a value)
                 mapStateSecond (a' value)’
    • Relevant bindings include
        a' :: Int -> Sem (State s1 : r) () (bound at tests/Main.hs:29:30)
        s' :: s1 (bound at tests/Main.hs:29:27)
        a :: Int -> Sem (State s : r) () (bound at tests/Main.hs:29:13)
        s :: s (bound at tests/Main.hs:29:11)
        (<>) :: Action r -> Action r -> Action r
          (bound at tests/Main.hs:29:16)
   |
30 |     mapStateFirst (a value)
   |     ^^^^^^^^^^^^^^^^^^^^^^^
```
</details>

I'm not sure if I'm designing something wrong here (should I not use `Semigroup`?), but if not, both of these function variants seem to have their use.

In my code I currently even use both variants. Usage is [here](https://github.com/Infinisil/arvy/blob/64f966cbd3afe0b7b87f480c94987f001c9a1be2/eval/Evaluation.hs#L23-L67), implementation of them is [here](https://github.com/Infinisil/arvy/blob/64f966cbd3afe0b7b87f480c94987f001c9a1be2/eval/Utils.hs#L38-L89).

isovector:
For any GADT of the form `E i m a` we know that if `i` appears only in positive position, we should generate a `Effunctor E` instance. If `i` is only in negative position, then `Contraeffunctor`. Otherwise we don't have any instances for it.

TheMatten:
@googleson78 I could do this as a separate PR if you want :slightly_smiling_face:
@isovector What about `InvEffunctor` in style of `Invariant` - could it have some practical use?

googleson78:
> @googleson78 I could do this as a separate PR if you want

Let me have some fun too! (unless I'm taking too long)

src/Polysemy/Internal/Effunctor.hs
googleson78: I haven't gone too deep into your example, but why not write `action` as `Member (State s) r => Int -> Sem r ()` instead?

TheMatten: He later tries to map `State` in this action, which requires him to remove original and add a modified one - `Member` constraint is not enough for that.

isovector:
What's the state of this?

googleson78:
0. There needs to be a decision on which (if not both) form of `effmap` should be in the type class (return-type polymorphic or not). - Don't know.
1. TH autogeneration needs implementing. - I haven't had the time recently. If anyone else wants to, go ahead!

TheMatten:
@googleson78 I may find some time for 1.

isovector:
@TheMatten and I spent some time thinking about this, but it unfortunately doesn't play very nicely with higher order effects.

PR #174, opened 2019-07-12
Make NonDet Higher-Order

KingoftheHomeless:
`Alternative` is typically used for simple exception handling a la "try-otherwise". However, `NonDet` is too weak in its current form to write interpreters for it that actually skip effects of the second branch. The solution to this is to make `NonDet` higher-order.

I also chose to represent `Choose` through embedding the two branches directly, as `Bool -> m a` is isomorphic to `(m a, m a)`, and the latter formulation is _probably_ more efficient.

This does not actually break anything - you can still write `runNonDet` with the same semantics as before (although my implementation could possibly be less efficient).
But now you can also write interpreters like `runNonDetMaybe`, which skips local effects (i.e. effects whose interpreters have already been run) of the second branch if the first branch succeeds.

Over at `polysemy-zoo`, this also allows me to write the following:
```haskell
runNonDetFinal :: (Member (Final m) r, Monad m, Alternative m)
               => Sem (NonDet ': r) a
               -> Sem r a
```
which interprets the uses of `<|>` and `empty` in terms of a final `Alternative`.

### DIFF
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -211,10 +211,7 @@ instance Monad (Sem f) where
 instance (Member NonDet r) => Alternative (Sem r) where
   empty = send Empty
   {-# INLINE empty #-}
-  a <|> b = do
-    send (Choose id) >>= \case
-      False -> a
-      True  -> b
+  a <|> b = send (Choose a b)
   {-# INLINE (<|>) #-}
 
 -- | @since 0.2.1.0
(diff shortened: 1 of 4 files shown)


KingoftheHomeless:
Actually, I can do better than the current semantics of `runNonDetMaybe` by simply reinterpreting the effect in terms of `Error`. That way, it will skip all effects of the second option, not just local ones.

KingoftheHomeless:
Turns out the "bad semantics" of `runNonDetMaybe` was just me being stupid when writing the tests.

I make too many of such mistakes.

isovector:
Excellent, thanks!

PR #179, opened 2019-07-16
Add nonDetToError

KingoftheHomeless:
Now that `NonDet` is higher-order, this strikes me as an interpreter that may be useful in practice, for people that like to use `<|>` rather than the clunkier `catch`.

### DIFF
--- a/src/Polysemy/NonDet.hs
+++ b/src/Polysemy/NonDet.hs
@@ -8,11 +8,14 @@ module Polysemy.NonDet
     -- * Interpretations
   , runNonDet
   , runNonDetMaybe
+  , nonDetToError
   ) where
 
 import Control.Applicative
 import Control.Monad.Trans.Maybe
 import Data.Maybe
+import Polysemy
+import Polysemy.Error
 import Polysemy.Internal
 import Polysemy.Internal.NonDet
 import Polysemy.Internal.Union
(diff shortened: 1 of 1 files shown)


KingoftheHomeless:
Would you also like me to add a version that takes an explicit lowering function to a base monad that's `Alternative`? (Although I still prefer `Final` interpreters rather than interpreters that work like that.)

isovector:
This is super cool. I'm not sure I understand the comment in https://github.com/polysemy-research/polysemy/pull/179#issuecomment-511981139; do we need to do this for a base monad, or is it sufficient to just have one for `Alternative f => ... -> Sem r (f a)` like nondet used to be?

KingoftheHomeless:
I'm talking about an additional interpreter which is like this:
```haskell
runNonDetInBase :: (Member (Lift m) r, Alternative m)
               => (forall x. Sem r x -> m x)
               -> Sem (NonDet ': r) a
               -> Sem r a
```

Or, via `Final`:
```haskell
runNonDetFinal :: (Member (Final m) r, Alternative m)
               -> Sem (NonDet ': r) a
              -> Sem r a
```

On an unrelated note, I have a radical idea which will allow `Final` interpreters to respect local/global state semantics, and therefore allow us to write a good interpreter for `Async` without any of the `Forklift` machinery. If it pans out, I'll create an issue about that soon-ish.

isovector:
> Edit: Didn't pan out. It relied on Weaving e m a being invariant in m, but that breaks raise and friends (but nothing else from what I can tell.)
 
@KingoftheHomeless please shoot it my way anyway! `raise` shouldn't be a real problem!

KingoftheHomeless:
Actually, it is. The more I'm experimenting with this, the more I realize this isn't viable.

To explain:
The idea was to introduce another field to `Weaving e n a`. Given that the existential monad is `m` and the existential functor `f`, the field would be `weaveRestore :: forall x. n (f x) -> m x`. That is, a way to "rebuild" the existentially hidden monad after a distribution has taken place. This kind of power is inherent to all interpreters that simply reduce - even including interpreters that work via an intermediary monad, like `runState`, etc. because it's easy to do `forall x. Sem r (s, x) -> Sem (State s ': r) x`.

With the addition of `weaveRestore`, `WithWeaving` of `Final` can also be expanded to include it. This grants enough power to actually give `runAsyncFinal` local/global state semantics depending on the order interpreters are run. It's hard to explain why over text, but if you are interested in a call, I could more easily break it down.

Anyway, the problem with `withRestore` is that it makes `withWeaving` invariant, and that breaks all interpreters that don't simply reduce a `Sem`. I've already mention `raise`, but other casualties include all flavours of `reinterpret`, so this DEFINITELY isn't worth it.

I experimented with a more restricted, non-invariant version like this: `weaveRestore :: forall x. f x -> m x`, but you can't make that compose.

KingoftheHomeless:
Anyway, we should discuss that some other place. Do you have any suggestion where? I don't feel an issue is appropriate for something like this.

To get back to the pull request: should I add `runNonDetInBase`?

isovector:
I'm happy to merge this PR. I'm less sold on the lift/final versions, but could be convinced by a motivating use-case. Regarding the other discussion, we could set up a call! Feel free to email me at sandy@sandymaguire.me and we can set something up.

PR #181, opened 2019-07-17
Document change in semantics based on order of interpreters

### LINKED ISSUE #127: Documentation for the difference in semantics on when you run your effects
What is the difference between `runState . runError` and `runError . runState`? When is `runResourceBase` safe? The answers to all of these questions and more. COMING THIS SUMMER

TheMatten:
Fixes #127.

Does `resourceToIO` need any more documentation?

### DIFF
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -100,8 +100,7 @@ import Polysemy.Internal.Union
 -- main = example
 --     'Data.Function.&' 'Polysemy.Error.runError'
 --     'Data.Function.&' fmap (either id id)
---     'Data.Function.&' 'Polysemy.State.runState' ""
---     'Data.Function.&' fmap snd
+--     'Data.Function.&' 'Polysemy.State.evalState' ""
 --     'Data.Function.&' 'runM'
 --     'Data.Function.&' (print =<<)
 -- @
@@ -114,8 +113,7 @@ import Polysemy.Internal.Union
 -- @
 -- main :: IO ()
 -- main = example
---     'Data.Function.&' 'Polysemy.State.runState' ""
---     'Data.Function.&' fmap snd
+--     'Data.Function.&' 'Polysemy.State.evalState' ""
 --     'Data.Function.&' 'Polysemy.Error.runError'
 --     'Data.Function.&' fmap (either id id)
 --     'Data.Function.&' 'runM'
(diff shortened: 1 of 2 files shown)


src/Polysemy/Internal.hs
```diff
@@ -72,6 +72,60 @@ import Polysemy.Internal.Union
 -- is the order in which you call the interpreters that determines the
 -- monomorphic representation of the @r@ parameter.
 --
+-- While order of effects in effect stack by itself is not important, order of
```
googleson78: Doesn't the order of interpreters *determine* the order of the effect stack? (what I mean is that the effect stack is important - it would be more accurate to say that you determine its order via the order of interpreters) Sorry if I'm just misunderstanding!

isovector: Yeah, @googleson78 has the right idea here!

TheMatten: Yes - I was thinking about mappings, ignoring and stuff that changes stack in a little bit different ways and didn't use the right words :sweat_smile:

src/Polysemy/Internal.hs
```diff
@@ -72,6 +77,53 @@ import Polysemy.Internal.Union
...
+--   example
+--     & evalState ""
+--     & runError
+--     & fmap (either id id)
+--     & runM
+--     & (print =<<)
+-- :}
+-- "start-catch"
+--
+-- Good rule of thumb is to handle effect which should have "global" behaviour
```
googleson78: Am I right in understanding you mean handle them first, before everything else?

googleson78: I think you need to escape the double quotes here, or haddock might consider `global` to be a module name? Not sure though

(`"global"` -> `\"global\"`)

src/Polysemy/Internal.hs
```diff
@@ -72,6 +77,53 @@ import Polysemy.Internal.Union
...
+-- over other effect later in chain.
```
googleson78: Nits:
`Good rule of thumb` -> `A good rule of thumb`
`effect` (here and perhaps also in line 124) should be plural.
`later in chain` -> `later in the chain`

src/Polysemy/Internal.hs
TheMatten: We don't really care e.g. about `Reader`-like effects when it comes to ordering as far as I know

TheMatten: You're probably right, thanks :slightly_smiling_face:

src/Polysemy/Internal.hs
TheMatten: I meant singular, but plural is technically right and sounds better :smile:

TheMatten:
@isovector what do you think?

isovector:
love it. thanks for tackling this. lets ship v1!

PR #182, opened 2019-07-19
[#180] Use type-errors-pretty package

### LINKED ISSUE #180: Would it be a good idea to use `type-errors-pretty` to write text of error messages?
I've [announced the `type-errors-pretty` library](https://twitter.com/ChShersh/status/1150707729658286080) recently. And I wonder, would it improve code readability and maintainability if it will be used instead of default `ErrorMessage`? I see that `polysemy` has a lot of custom type errors, so I wonder, whether the library with a lot of custom error messages can benefit from prettier way of defining the text of those messages.

For example, instead of existing:

```haskell
type AmbigousEffectMessage (rstate :: EffectRowCtor)
                           (r :: EffectRow)
                           (e :: k)
                           (t :: Effect)
                           (vs :: [Type]) =
        ( 'Text "Ambiguous use of effect '"
    ':<>: 'ShowType e
    ':<>: 'Text "'"
    ':$$: 'Text "Possible fix:"
    ':$$: 'Text "  add (Member ("
    ':<>: 'ShowType t
    ':<>: 'Text ") "
    ':<>: ShowRQuoted rstate r
    ':<>: 'Text ") to the context of "
    ':$$: 'Text "    the type signature"
    ':$$: 'Text "If you already have the constraint you want, instead"
    ':$$: 'Text "  add a type application to specify"
    ':$$: 'Text "    "
    ':<>: PrettyPrintList vs
    ':<>: 'Text " directly, or activate polysemy-plugin which"
    ':$$: 'Text "      can usually infer the type correctly."
        )
```

This can look like this using `type-errors-pretty` library:

```haskell
type AmbigousEffectMessage (rstate :: EffectRowCtor)
                           (r :: EffectRow)
                           (e :: k)
                           (t :: Effect)
                           (vs :: [Type])
    = "Ambiguous use of effect '" <> e <> "'"
    % "Possible fix:"
    % "  add (Member (" <> t <> ") " <> ShowRQuoted rstate r <> ") to the context of "
    % "    the type signature"
    % "If you already have the constraint you want, instead"
    % "  add a type application to specify"
    % "    " <> PrettyPrintList vs <> " directly, or activate polysemy-plugin which"
    % "      can usually infer the type correctly."
```

chshersh:
Resolves #180

I hope it makes it easier to write and maintain messages of custom type errors.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -27,6 +27,7 @@ dependencies:
 - unagi-chan >= 0.4.0.0 && < 0.5
 - async >= 2.2 && < 3
 - type-errors >= 0.2.0.0
+- type-errors-pretty >= 0.0.0.0 && < 0.1
 
 default-extensions:
   - DataKinds
--- a/polysemy.cabal
+++ b/polysemy.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 7e0e9242c36e377756ae6ef007d19a16cdaf3f244855b0bfc97e267f4df143ea
+-- hash: 168b7f25b456d0a92b6c245a91e1acc9c81ef40eaa2893f2eb6a238abd74f7f9
 
 name:           polysemy
 version:        0.7.0.0
@@ -83,6 +83,7 @@ library
     , th-abstraction >=0.3.1.0 && <0.4
     , transformers >=0.5.2.0 && <0.6
     , type-errors >=0.2.0.0
+    , type-errors-pretty >=0.0.0.0 && <0.1
     , unagi-chan >=0.4.0.0 && <0.5
   if impl(ghc < 8.6)
     default-extensions: MonadFailDesugaring TypeInType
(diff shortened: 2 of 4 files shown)


isovector:
I like this, thanks! Happy to merge so long as `type-errors-pretty` is already on stackage. Otherwise they'll yell at me :)

package.yaml
```diff
@@ -27,6 +27,7 @@ dependencies:
 - unagi-chan >= 0.4.0.0 && < 0.5
 - async >= 2.2 && < 3
 - type-errors >= 0.2.0.0
+- type-errors-pretty >= 0.0.0.0 && < 0.1
```
symbiont-joseph-kachmar: Is the lower bound here necessary if it's just going to be `0.0.0.0`? 

I would have thought that `type-errors-pretty < 0.1` would suffice.

chshersh: @symbiont-joseph-kachmar Usually I specify both bounds: lower and upper. This helps to have more robust build plans. Yes, I understand that `0.0.0.0` is the smallest possible version, so in this particular corner case it should be enough to specify `< 0.1`. I'm using `^>=` Cabal operator to specify lower and upper bounds at the same time. But I see that `.cabal` file here is generated via `package.yaml` and uses `cabal-version: 1.12` so I can't use this operator. That's why writing bounds explicitly not to forget to update them both in case of breaking changes.

chshersh:
@isovector Feel free to merge :slightly_smiling_face: `type-errors-pretty` is already on Stackage. I've added it immediately after release several days ago :slightly_smiling_face: 

* https://github.com/commercialhaskell/stackage/pull/4715

PR #186, opened 2019-07-24
Simplified Forklift machinery, remove `LastMember`

### LINKED ISSUE #183: 'LastMember' vs 'prj'
`LastMember` is currently used in `runViaForklift` to identify individual uses of `embed` and interpret these in the forked thread rather than shipping them off to the interpreter thread.
However, there already exists a mechanism to identify if an union corresponds to a specific effect: namely, `prj` in `Polysemy.Internal.Union`. Indeed, it's possible to use `prj` in place of `decompLast`.

What benefit does `LastMember` bring compared to `prj`?

KingoftheHomeless:
This simplifies the `Forklift` machinery, and removes the need for `LastMember`.

Note that this is a breaking change, which is a bit of a shame seeing how we just released v1...

This closes #183

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -3,7 +3,6 @@ module Polysemy
     Sem ()
   , Member
   , Members
-  , LastMember
 
   -- * Running Sem
   , run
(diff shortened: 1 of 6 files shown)


isovector:
It's so much prettier. Probably worth pushing out 1.1.0.0 sooner than later to limit the breakage here.

isovector:
Thanks!

PR #187, opened 2019-07-26
Improve semantics of Fixpoint

KingoftheHomeless:
I stumbled across this interesting improvement to the semantics of `Fixpoint` interpreters.

The big problem with the current interpreters of fixpoint is that they diverge if any distribution law `forall x. f (m x) -> Sem r (f x)` of any `Weaving` is strict in the functorial state, which they pretty much _always are_ if you have run any return-value changing interpreter such as `runState` and `runError`.

However, it's possible to improve upon this by taking cues from the `MonadFix` implementations of `StateT` and `ExceptT`. Instead of using the recursive functorial state, we could use the initial state. Because of this, we need to extract the encapsulated value of the recursive state in order to have something to provide to the function. We can do that using the inspector, inserting a deferred bomb if the inspector fails. The idea here is that if the inspector failed, then the resulting functorial state can't contain any values; thus, embedding the bomb doesn't pose any issue, as it won't be observable.

Through this mechanism, `Fixpoint` becomes a lot more powerful than it used to be. Previously, it required that any effects with result-value changing interpreters are interpreted in the final monad, but now, you can use `runFixpoint` safely together with `runError`, and `runState`.

This isn't a perfect solution, however; unfortunately, the interaction with `runNonDet` doesn't reflect that of the list monad's `MonadFix` instance. (Which is a possible argument for why you'd want a `lowerNonDet` interpreter to solve that)

Another flaw of this solution is that it hinges on the implication that if the inspector fails, then the functorial state doesn't contain any values. This is true for all interpreters in polysemy, but it's easy to violate it if you decide to write your own return-value changing interpreter and just provide `const Nothing` as the inspector when you `weave`.

What's your opinion of this @isovector ? I'll return in a while to write some proper tests.

### DIFF
--- a/test/FixpointSpec.hs
+++ b/test/FixpointSpec.hs
@@ -46,7 +46,7 @@ test2 =
   . runFixpoint run
   . runError
   $ mdo
-  a <- throw (1 : a) `catch` (\e -> return e)
+  a <- throw (2 : a) `catch` (\e -> return (1 : e))
   return a
 
 test3 :: Either () (Int, Int)
@@ -78,7 +78,7 @@ spec = parallel $ describe "runFixpoint" $ do
     test1 `shouldBe` ("12",  (2, ()))
   it "should work with runError" $ do
     let res = fmap (take 10) test2
-    res `shouldBe` Right (replicate 10 1)
+    res `shouldBe` Right (take 10 $ cycle [1,2])
   it "should not trigger the bomb" $ do
     test3 `shouldBe` Left ()
   it "should trigger the bomb" $ do
--- a/test/FixpointSpec.hs
+++ b/test/FixpointSpec.hs
@@ -1,0 +1,97 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE RecursiveDo #-}
+module FixpointSpec where
+
+import Control.Exception (try, evaluate)
+import Control.Monad.Fix
+
+import Polysemy
+import Polysemy.Fixpoint
+import Polysemy.Error
+import Polysemy.State
+import Polysemy.Output
+
+import Test.Hspec
+
+data FinalState s m a where
+  GetEventualState :: FinalState s m s
+
+makeSem ''FinalState
+
+runFinalState :: Member Fixpoint r
+              => s
+              -> Sem (State s ': FinalState s ': r) a
+              -> Sem r (s, a)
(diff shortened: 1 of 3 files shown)


isovector:
So cool. I often want to write `getEventualState :: Members '[Fixpoint, State s] r => Sem r s`, but this thing would always diverge. 

I'm feeling spoiled by all of the fantastic PRs you're sending my way!

KingoftheHomeless:
Thanks. I'll try to keep them up!

KingoftheHomeless:
Hmm. It IS actually possible for the bomb to trigger, under the same circumstances that bombs would trigger under MTL.

```haskell
test :: (String, Either () Int)
test =
    run
  . runFixpoint run
  . runOutputMonoid (show @Int)
  . runError
  $ do
  mfix $ \a -> do
    output a
    throw ()
```

Compare to:
```haskell
import Control.Monad.Trans.Except
import Control.Monad.Trans.Writer.Strict

test' :: (Either () Int, String)
test' =
    runIdentity
  . runWriterT
  . runExceptT
  $ do
  mfix $ \a -> do
    lift $ tell (show a)
    throwE ()
```

Both of these spits out bombs (`test'` spits out the bomb in `ExceptT`'s `MonadFix` instance).

Funnily enough, if you use `Control.Monad.Trans.CPS`, then `test'` will hang forever. This is due to a discrepancy between polysemy and MTL; MTL forces the result of a tell, polysemy doesn't (which means the space leak isn't actually completely gone).

I'll change the bomb message to reflect the fact that it IS possible to see it, without it being the interpreter's fault.

Other than that, it all seems to work as intended (except for the wrong interaction with `runNonDet`).

KingoftheHomeless:
Tests and fixed docs are up now. Ready to be merged if you're okay with how horribly verbose the documentation is.

test/FixpointSpec.hs
```diff
@@ -0,0 +1,97 @@
...
+test1 =
+    run
+  . runFixpoint run
+  . runOutputMonoid (show @Int)
+  . runFinalState 1
+  $ do
+  s  <- get @Int
+  s' <- getEventualState @Int
+  output @Int s
+  output @Int s'
```
isovector: so cooooool!

KingoftheHomeless: :)

PR #194, opened 2019-07-28
Fix strictness semantics of State and Writer

### LINKED ISSUE #192: runState rewrite rules change strictness semantics
I ran into this while trying to make output effects (`Trace`, `Writer`, `Output`) more strict (which is a separate issue I'll open tomorrow). 

The easiest way to see this is to test with the following:
```haskell
runOutput :: Monoid o
          => Sem (Output o ': r) a
          -> Sem r (o, a)
runOutput =
  let lower = runState mempty
  in lower . reinterpret (\(Output a) -> do
        s <- get
        put $! s <> a
    )

test :: ()
test =
    snd
  . run
  . runOutput @String
  $ do
  output @String undefined
  return ()
```

When `lower` is marked `NOINLINE`, `test` throws an exception; if `lower` is marked `INLINE`, the `runState/reinterpret` rule of `Polysemy.State` will fire, and `test` will not throw an exception. You can also see it by removing the rule entirely, in which case `test` will throw an exception whether or not `lower` is marked `NO`/`INLINE`.

This is an issue because it's surprising behaviour, and it makes it difficult to write strict, stateful interpreters (like `runOutput` above is supposed to be).

I'm not entirely sure what makes the rewrite rule behave this way.

### LINKED ISSUE #193: Monoidal interpreters for writer effects aren't optimally strict
Unlike `Control.Monad.Trans.Writer.CPS.tell` (which is built upon [`writer`](https://hackage.haskell.org/package/transformers-0.5.6.2/docs/src/Control.Monad.Trans.Writer.CPS.html#writer)), the monoidal interpreters for `Writer`/`Output`/`Trace` don't force the result of a `tell`/`output`/`trace`, leading to a build-up of thunks in the state a la bad uses of `foldl`.

Ideally, we would be able to fix this by replacing the uses of `modify` in the relevant interpreters with the following:
```haskell
modify' :: Member (State s) r => (s -> s) -> Sem r ()
modify' f = do
  s <- get
  put $! f s
```
But unfortunately, #192 causes this not to work properly.

KingoftheHomeless:
This closes #192 and closes #193

### DIFF
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -22,6 +22,7 @@ module Polysemy.Internal.Combinators
   , lazilyStateful
   ) where
 
+import           Control.Monad
 import qualified Control.Monad.Trans.State.Lazy as LS
 import qualified Control.Monad.Trans.State.Strict as S
 import qualified Data.Tuple as S (swap)
@@ -89,10 +90,11 @@ interpretInStateT
     -> Sem (e ': r) a
     -> Sem r (s, a)
 interpretInStateT f s (Sem m) = Sem $ \k ->
-  fmap S.swap $ flip S.runStateT s $ m $ \u ->
+  (S.swap <$!>) $ flip S.runStateT s $ m $ \u ->
     case decomp u of
         Left x -> S.StateT $ \s' ->
-          k . fmap S.swap
+              (S.swap <$!>)
+            . k
             . weave (s', ())
                     (uncurry $ interpretInStateT f)
                     (Just . snd)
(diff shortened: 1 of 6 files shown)


src/Polysemy/Internal/Combinators.hs
```diff
@@ -24,6 +24,7 @@ module Polysemy.Internal.Combinators
 
 import qualified Control.Monad.Trans.State.Lazy as LS
 import qualified Control.Monad.Trans.State.Strict as S
+import qualified Data.Tuple as S (swap)
```
isovector: The swap defined in this module was my attemp to get around runFixpoint diverging. I don't think it's necessary anymore.

KingoftheHomeless: It's not, since the fixpoint tests I added still succeed. Still, maybe we want the lazy swap for `runLazyState`?

PR #197, opened 2019-08-01
Have AsyncSpec use locks instead of sleeps

### LINKED ISSUE #195: polysemy-1.0.0.0 flaky test?
Just wanted to report that I saw the following test failure, but it did not fail again when I reran the test suite.

```
      test/AsyncSpec.hs:43:5: 
      1) Async.async should thread state and not lock
           ["1> hello","2> olleh","1> olleh"] does not contain ["1> hello","2> olleh","1> pong"]
```

This was on the stackage build server while building `nightly-2019-07-31`.

KingoftheHomeless:
This presumably fixes #195, but that issue warrants some further investigation

### DIFF
--- a/test/AsyncSpec.hs
+++ b/test/AsyncSpec.hs
@@ -3,6 +3,7 @@
 module AsyncSpec where
 
 import Control.Concurrent
+import Control.Concurrent.MVar
 import Control.Monad
 import Polysemy
 import Polysemy.Async
@@ -21,22 +22,24 @@ spec = describe "async" $ do
       let message :: Member Trace r => Int -> String -> Sem r ()
           message n msg = trace $ mconcat
             [ show n, "> ", msg ]
-
+      ~[lock1, lock2] <- embed $
+        replicateM 2 newEmptyMVar
       a1 <- async $ do
           v <- get @String
           message 1 v
           put $ reverse v
 
-          embed $ threadDelay 1e5
+          embed $ putMVar lock1 ()
+          embed $ takeMVar lock2
           get >>= message 1
 
-          embed $ threadDelay 1e5
           get @String
 
       void $ async $ do
-          embed $ threadDelay 5e4
+          embed $ takeMVar lock1
           get >>= message 2
           put "pong"
+          embed $ putMVar lock2 ()
 
       await a1 <* put "final"
 


test/AsyncSpec.hs
```diff
@@ -21,22 +22,24 @@ spec = describe "async" $ do
...
           v <- get @String
           message 1 v
           put $ reverse v
 
-          embed $ threadDelay 1e5
+          embed $ putMVar lock1 ()
+          embed $ takeMVar lock2
           get >>= message 1
 
-          embed $ threadDelay 1e5
```
KingoftheHomeless: I don't get what this `threadDelay` is for, so I simply removed it.

isovector: This was so that the final `get` in this block would read the `pong` from the other thread.

test/AsyncSpec.hs
```diff
@@ -21,22 +22,24 @@ spec = describe "async" $ do
       let message :: Member Trace r => Int -> String -> Sem r ()
           message n msg = trace $ mconcat
             [ show n, "> ", msg ]
-
+      ~[lock1, lock2] <- embed $
```
isovector: ooc, why the lazy match here?

KingoftheHomeless: Irrefutable pattern match; doesn't require `Sem r` to be `MonadFail`.

test/AsyncSpec.hs
KingoftheHomeless: But this is already guaranteed (or should be, rather) from the previous `threadDelay`, isn't it?

test/AsyncSpec.hs
isovector: Right -- makes sense! Thanks for the explanation.

test/AsyncSpec.hs
isovector: The logic I intended (perhaps not what I implemented) was this interleaving of threads:

thread1
thread2
thread1

KingoftheHomeless: Yeah, that is what happens:
1. thread2 sleeps. thread1 gets, messages, puts, and sleeps
2. thread2 wakes, gets, messages, puts, and terminates
3. thread1 wakes, gets, messages, puts, and sleeps.
4. thread1 wakes, gets, and terminates.
5. main thread completes its `await` as thread1 has terminated, puts, and terminates.

The `threadDelay` I removed was the one in time slot 3, as it's seemingly redundant.

isovector:
Thanks!

PR #199, opened 2019-08-01
Add Fail effect

KingoftheHomeless:
Currently, `Sem`'s `MonadFail` instance is bound to `NonDet`, and invokes `fail` through `const empty`. This is too specialized for my tastes, so this introduces a `Fail` effect a la `fused-effects`. 

This has the downside of being a breaking change for anyone who has ever used the `MonadFail` instance. It's possible that cost is too heavy to be worth it.

### DIFF
--- a/src/Polysemy/Internal/Fail.hs
+++ b/src/Polysemy/Internal/Fail.hs
@@ -1,0 +1,3 @@
+module Polysemy.Internal.Fail where
+
+data Fail m a = Fail String
--- a/src/Polysemy/Internal/Fail.hs
+++ b/src/Polysemy/Internal/Fail.hs
@@ -1,3 +1,0 @@
-module Polysemy.Internal.Fail where
-
-newtype Fail m a = Fail String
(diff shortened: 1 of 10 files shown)


isovector:
I like this. I'm not sure where I got the idea for a nondet fail, but I feel like it was fused-effects in the first place!

isovector:
Is this good to merge?

KingoftheHomeless:
If you're fine with it, yes.

src/Polysemy/Internal/Fail.hs
```diff
@@ -0,0 +1,3 @@
+module Polysemy.Internal.Fail where
```
isovector: Let's put this in `Polysemy.Fail.Type` instead

KingoftheHomeless: Ok.

KingoftheHomeless:
I've added a `TODO` for the `@since` under the `MonadFail` instance since I figured the new definition was important enough to update the `@since`. You'll have to update that when you release v1.1

PR #200, opened 2019-08-03
Add `state` action to `State`

TheMatten:
Similar function can be found e.g. in `transformers` - but this one follows swapped order of elements in tuple already present in `polysemy`'s functions (like `runState` or `stateful`).

- [ ] Should we test it?

### DIFF
--- a/src/Polysemy/State.hs
+++ b/src/Polysemy/State.hs
@@ -8,6 +8,7 @@ module Polysemy.State
   , get
   , gets
   , put
+  , state
   , modify
   , modify'
 
@@ -40,8 +41,9 @@ import           Polysemy.Internal.Union
 -- Interpreters which require statefulness can 'Polysemy.reinterpret'
 -- themselves in terms of 'State', and subsequently call 'runState'.
 data State s m a where
-  Get :: State s m s
-  Put :: s -> State s m ()
+  Get   :: State s m s
+  Put   :: s -> State s m ()
+  State :: (s -> (s, a)) -> State s m a
 
 makeSem ''State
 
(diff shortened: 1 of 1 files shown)


isovector:
I'd prefer not to do this; now everyone will need to handle it in their interpreters, and it doesn't buy us anything semantically.

TheMatten:
Turns out I don't actually want to use it in specific places as I thought previously - okay, let's close it for now then.

KingoftheHomeless:
A `state` action actually _does_ buy us something semantically new: atomic modification.

TheMatten:
@KingoftheHomeless Hmm, it's a nice property to carry in types - what about having:
```
data Atomic s m a where
  Atomic :: (s -> (s, a)) -> Atomic s m a
```
separately - and then we could have:
```
runStateAtomic :: Member (Atomic s) r => Sem (State s : r) a -> Sem r a
runStateAtomic = interpret $ \case
  Get   -> atomic $ \s -> (s, s)
  Put s -> atomic $ const (s, ())
```
to provide `get` and `put` over same state if needed.

TheMatten:
@isovector What do you think about adding `Atomic` to zoo?

KingoftheHomeless:
I'd probably call it `AtomicState`, and the action `atomicModify`.

The seperation between `State` and `AtomicState` is somewhat unfortunate if you want my opinion. I'd prefer if it were up to the interpreter, but the breakage of adding `state` to `State` would most likely be significant.

ocharles:
In my work I have an effect 'Atomic m', which shows how to atomically embed any 'm a' into another monad. So in this, you'd have Atomic (Sem [State s])

TheMatten:
@ocharles sound cool! I'm just worried it would be a little bit of an overkill in this case :smile:

PR #202, opened 2019-08-05
Hide visibility of Polysemy.Fixpoint.bomb, fix changelog

KingoftheHomeless:
I accidentally exported the `bomb` in `Polysemy.Fixpoint`, so this moves it to the internal module.

I also fixed some minor errors in the changelog of v1.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -30,13 +30,13 @@
 - Renamed `runTraceAsOutput` to `traceToOutput`
 - Renamed `runTraceIO` to `traceToIO`
 - Renamed `sendM` to `embed` (thanks to @googleson78)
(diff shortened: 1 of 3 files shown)


ChangeLog.md
```diff
@@ -30,13 +30,13 @@
 - Renamed `runTraceAsOutput` to `traceToOutput`
 - Renamed `runTraceIO` to `traceToIO`
 - Renamed `sendM` to `embed` (thanks to @googleson78)
-- The `NonDet` effect will no longer perform effects in untaken branches (thanks to @KingoftheHomeless)
```
KingoftheHomeless: To explain, that behaviour is specific to `runNonDetMaybe` and `nonDetToError`. `runNonDet` still function as it did before.

isovector: makes sense

PR #203, opened 2019-08-06
Add right-associative variants of runOutputMonoid and runWriter

KingoftheHomeless:
`runOutputMonoid` and `runWriter` combine the monoidal values left-associatively, which results in terrible performance when used together with linked lists such as `String`. This can be addressed by instead using `DList a`/`Endo [a]` as the monoid, which, through some CPS magic, effectively "right-associates" uses of `<>`, making lists behave asymptotically better.

I figured we could add interpreters that automate the conversions between a monoid and its `Endo` variant, and therefore do all this wrangling under the hood, so that's what this pull request is about.

Right now, the interpreters are called `runOutputMonoidAssocR` and `runWriterAssocR`, which I don't particularly like. Does anyone have suggestions for better names?

### DIFF
--- a/src/Polysemy/Output.hs
+++ b/src/Polysemy/Output.hs
@@ -10,11 +10,13 @@ module Polysemy.Output
     -- * Interpretations
   , runOutputList
   , runOutputMonoid
+  , runOutputMonoidAssocR
   , ignoreOutput
   , runOutputBatched
   , runOutputSem
   ) where
 
+import Data.Semigroup (Endo(..))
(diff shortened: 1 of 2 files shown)


KingoftheHomeless:
Apprently Travis can't find GHC 8.4.4? Weird.

KingoftheHomeless:
I closed/reopened the pull request to trigger a restart of the Travis build.

isovector:
Excellent, thank you! I've always been worried about this.

PR #204, opened 2019-08-06
add DataKinds to the checkExtensions list

### LINKED ISSUE #201: TH should give error about missing `DataKinds` extension
Currently, use of `makeSem`(`_`) without `DataKinds` results in error:
```
Parsemy.hs:18:1-16: error:
    Illegal type: ‘"Parsemy"’ Perhaps you intended to use DataKinds
   |
18 | makeSem ''Parser
   | ^^^^^^^^^^^^^^^^
```
which is not bad, but does not explicitly say that we depend on that extension.
The only change needed should be adding extension to list of required ones [here](https://github.com/polysemy-research/polysemy/blob/829e3202fc7d09faaf757cddd11a8a741b687a70/src/Polysemy/Internal/TH/Effect.hs#L125).
https://github.com/polysemy-research/polysemy/blob/829e3202fc7d09faaf757cddd11a8a741b687a70/src/Polysemy/Internal/TH/Effect.hs#L125

pepegar:
fixes #201

### DIFF
--- a/src/Polysemy/Internal/TH/Effect.hs
+++ b/src/Polysemy/Internal/TH/Effect.hs
@@ -122,7 +122,7 @@ makeSem_ = genFreer False
 -- constructors into 'Sem' actions.
 genFreer :: Bool -> Name -> Q [Dec]
 genFreer should_mk_sigs type_name = do
-  checkExtensions [ScopedTypeVariables, FlexibleContexts]
+  checkExtensions [ScopedTypeVariables, FlexibleContexts, DataKinds]
   (dt_name, cl_infos) <- getEffectMetadata type_name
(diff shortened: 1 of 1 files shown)


src/Polysemy/Internal/TH/Effect.hs
```diff
@@ -171,4 +171,3 @@ genDec should_mk_sigs cli = do
                  []
         ]
     ]
-
```
isovector: nit: please revert this change

pepegar: You're right, sorry :)

isovector:
thanks!

PR #206, opened 2019-08-15
Add AtomicState and atomic interpreters for Output/Writer

KingoftheHomeless:
I haven't added any tests for `AtomicState` itself, tell me if you think they would be necessary.

### DIFF
--- a/src/Polysemy/Writer.hs
+++ b/src/Polysemy/Writer.hs
@@ -14,17 +14,12 @@ module Polysemy.Writer
     -- * Interpretations
   , runWriter
   , runWriterAssocR
-  , runWriterIORef
-  , runWriterTVar
 
     -- * Interpretations for Other Effects
   , outputToWriter
   ) where
 
-import Control.Concurrent.STM
-
 import Data.Bifunctor (first)
-import Data.IORef
 
 import Polysemy
 import Polysemy.Output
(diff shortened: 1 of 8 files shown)


src/Polysemy/Writer.hs
```diff
@@ -122,3 +127,122 @@ runWriterAssocR =
...
+  s <- readTVar tvar
+  writeTVar tvar $! s <> o
+{-# INLINE runWriterTVar #-}
+
+runWriterTVarAction :: forall o a r
+                     . (Monoid o, Member (Embed IO) r)
+                    => (o -> STM ())
+                    -> Sem (Writer o ': r) a
+                    -> Sem r a
+runWriterTVarAction act = interpretH $ \case
```
KingoftheHomeless: I'm interested to hear what you think about this. The implementation is a mess, but it gives us pretty good semantics.

KingoftheHomeless:
The big problem with `runWriterTVar` - that writes are lost if a `listen`ed/`pass`ed computation fails for whatever reason - can't be fixed using `Embed IO` alone, BUT it can be fixed using `Final IO`, so once we migrate `Final` to polysemy proper I'll add `runWriterTVarFinal`, which will be the optimal interpreter for `Writer` if you always want to preserve writes.

src/Polysemy/Writer.hs
```diff
@@ -122,3 +127,122 @@ runWriterAssocR =
     {-# INLINE go #-}
   in fmap (first ($ mempty)) . go
 {-# INLINE runWriterAssocR #-}
+
+------------------------------------------------------------------------------
+-- | Run a 'Writer' effect by transforming it into atomic
+-- operations over an 'IORef'.
+--
+-- /Note/: All 'tell's in the argument of a 'listen' or 'pass' will temporarily
```
isovector: Yeah, this is a little weird, but I guess it makes sense.

src/Polysemy/Writer.hs
```diff
@@ -122,3 +127,122 @@ runWriterAssocR =
...
+    {-
+       KingoftheHomeless: this switch controls how 'tell's of
+       'mm' are committed. As long as 'mm' is running,
+       it and any async computations spawned by it will write to
+       the local tvar. Once the 'listen' has
+       completed, the switch will flip, and any async computations
+       spawned by 'mm' will write to the
+       global tvar (which is represented by 'act').
+
+       Since the switch flip is part of the same 'atomically' that
```
isovector: I think this comment would be more clear in `act'`, too me a few goes-around to figure out why this is true.

KingoftheHomeless: Ok, I'll move this comment when I have the opportunity.

src/Polysemy/Writer.hs
KingoftheHomeless: It's borderline for me. Perhaps it's better to wait until `Final` drops before I write `runWriterTVar`? I.e. only keep `runOutputMonoidIORef` and `runOutputMonoidTVar`?

KingoftheHomeless:
The more I consider this, the more I feel like that it'd best to drop `runWriterIORef` and `runWriterTVar` for the time being, and implement `runWriterTVar` using `Final IO` once `Final` arrives in polysemy. Better to have only one proper interpreter instead of one proper interpreter and two weird interpreters. Do you agree @isovector ?

isovector:
@KingoftheHomeless yeah, that seems like a better idea to me, rather than introducing it and then immediately issuing a breaking change!

KingoftheHomeless:
This is ready to be merged.

PR #210, opened 2019-08-17
Improve choice of functorial state for runNonDet

KingoftheHomeless:
I discovered while working on polysemy-zoo that `runNonDet` plays really badly with `Cont`, no matter how `Cont` is interpreted.

Consider this example, using the most reliable interpreters available (this uses https://github.com/polysemy-research/polysemy-zoo/pull/45):
```haskell
knot :: Member (Cont ref) r => a -> Sem r (a, a -> Sem r b)
knot a = callCC $ \c -> let b x = c (x, b) in pure (a, b)

bad :: IO [Int]
bad = do
  ref <- newIORef []
  _ <-  (`runContT` pure)
      . runFinal
      . embedToFinal
      . embedToMonadIO
      . contToFinal
      . runOutputMonoidIORef ref (:[])
      . runReader ()
      . runNonDet @[]
      $ do
        (end, (i, loop)) <- local id $ ((,) 10 <$> knot 1) <|> ((,) 8 <$> knot 2)
        output i
        when (i < end) $
          loop $! i + 1
  readIORef ref
```
Without the `local id`, this does what you'd expect: return `[1,2,3,4,5,6,7,8,9,10,2,3,4,5,6,7,8]`.
With the `local id`, however, this returns `[1,2,3,4,5,6,7,8,9,10,2,10,3,10,4,10,5,10,6,10,7,10,8]`!

I tracked this down to the choice of functorial state that is used when weaving through other effects in `runNonDet`; `[]`. So I took a cue from the instance of `MonadBaseControl` for `ListT` in the `list-t` package, and replaced the functorial state with one that fixes this bug!

Here's the thing, however; the change in functorial state makes the weaving of other higher-order actions in `runNonDet` extremely less efficient if the higher-order action is heavy in `<|>`. I haven't been able to figure out if the choice of `[]` as the functorial state has any other adverse effects except for the wrong interaction with `Cont`, so the question is if this change is worth it.

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -23,8 +23,8 @@ module Polysemy
     --
     -- @
     -- data Console m a where
-    --   WriteLine :: 'String' -> Console m ()
-    --   ReadLine  :: Console m 'String'
+    --   WriteLine :: String -> Console m ()
+    --   ReadLine  :: Console m String
     -- @
     --
     -- Notice that the @a@ parameter gets instantiated at the /desired return/
@@ -41,7 +41,7 @@ module Polysemy
     --
     -- @
     -- writeLine :: 'Member' Console r => String -> 'Sem' r ()
-    -- readLine  :: 'Member' Console r => 'Sem' r 'String'
+    -- readLine  :: 'Member' Console r => 'Sem' r String
     -- @
     --
     -- Effects which don't make use of the @m@ parameter are known as
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -23,12 +23,12 @@ module Polysemy
     --
     -- @
     -- data Console m a where
-    --   WriteLine :: String -> Console m ()
-    --   ReadLine  :: Console m String
+    --   WriteLine :: 'String' -> Console m ()
+    --   ReadLine  :: Console m 'String'
     -- @
     --
-    -- Notice that the @a@ parameter gets instataniated at the /desired return
-    -- type/ of the actions. Writing a line returns a '()', but reading one
+    -- Notice that the @a@ parameter gets instantiated at the /desired return/
+    -- /type/ of the actions. Writing a line returns a @()@, but reading one
     -- returns 'String'.
     --
     -- By enabling @-XTemplateHaskell@, we can use the 'makeSem' function
@@ -41,7 +41,7 @@ module Polysemy
     --
     -- @
     -- writeLine :: 'Member' Console r => String -> 'Sem' r ()
-    -- readLine  :: 'Member' Console r => 'Sem' r String
+    -- readLine  :: 'Member' Console r => 'Sem' r 'String'
     -- @
     --
     -- Effects which don't make use of the @m@ parameter are known as
--- a/src/Polysemy/NonDet.hs
+++ b/src/Polysemy/NonDet.hs
@@ -13,70 +13,18 @@ module Polysemy.NonDet
 
 import Control.Applicative
 import Control.Monad.Trans.Maybe
-import Data.Maybe
+
 import Polysemy
 import Polysemy.Error
 import Polysemy.Internal
 import Polysemy.Internal.NonDet
 import Polysemy.Internal.Union
 
-
---------------------------------------------------------------------------------
--- This stuff is lifted from 'fused-effects'. Thanks guys!
-runNonDetC :: (Alternative f, Applicative m) => NonDetC m a -> m (f a)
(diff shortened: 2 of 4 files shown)


src/Polysemy.hs
```diff
@@ -41,7 +41,7 @@ module Polysemy
     --
     -- @
     -- writeLine :: 'Member' Console r => String -> 'Sem' r ()
-    -- readLine  :: 'Member' Console r => 'Sem' r String
+    -- readLine  :: 'Member' Console r => 'Sem' r 'String'
```
isovector: This one is intentionally omitted; i'd rather just keep the links here for polysemy concepts.

isovector:
There's nothing special about `[]` here; iirc I just chose it because it was convenient. Maybe we could put our functorial state in a balanced binary tree instead of this `ListT` stuff --- would that improve the asymptotics? I'm not entirely sure what's going on with `unconsC`!

KingoftheHomeless:
`unconcC` is the `NonDetC`/[`LogicT`](https://hackage.haskell.org/package/logict-0.7.0.2/docs/Control-Monad-Logic.html#t:LogicT) equivalent of `ListT`'s `uncons` (https://hackage.haskell.org/package/list-t-1.0.3.1/docs/ListT.html#v:uncons). We need `unconsC` in order to transform `NonDetC (Sem r) a` to `Sem r (NonDetState r a)`. Somehow representing `NonDetState` using a binary tree doesn't solve that; we still need to split a `NonDetC` into a concrete head, and concrete tail.

`unconcC` is horribly inefficient for the same reason the `uncons` operation is horribly inefficient for the church encoding of lists (i.e. `newtype List a = List (forall r. (a -> r -> r) -> r -> r)`; i.e. simple limitations of the encoding; we always need to run through the entire stack in order to remove items from it. Compare this to `ListT`, where it's as easy as pattern matching on a newtype constructor.
The reason why we use `LogicT` instead of `ListT` is because `LogicT` has O(1) `<|>`, compared to O(n) for `ListT`, and that's even more important.

The solution is to use [Reflection without Remorse](http://okmij.org/ftp/Haskell/zseq.pdf). This offers a way to get a `ListT` monad transformer with both O(1) `<|>` and O(1) `uncons`, solving the problem.

Although now that I think about it more, perhaps the performance isn't degraded compared to when we use `[]`? Because the `runNonDet` in `fmap concat . traverse runNonDet` is also `O(n)`, so if the provided state is `n` items long,  `fmap concat . traverse runNonDet`  also displays quadratic behaviour.

KingoftheHomeless:
I did a small, informal benchmark of this. Under `-O2`, `NonDetState` seems to perform exactly the same as `[]`, except for consuming a tiny bit more memory (0.6%).

I'm slightly dubious towards the result however. For one, use of `<|>` inside a higher-order action (which causes the distribution to be used) doesn't seem to be significantly less efficient compared to using it outside a higher-order action. For another, I tried my hand at making a reflection-without-remorse `ListT`, and performance when using it as the carrier was absolute garbage.

I'll look into doing a more thorough benchmark later.

isovector:
The reflection without remorse approach has always had terrible performance in my experience; the problem is not the asymptotics but the huge constant factors associated with all of the allocation and collection of the intermediate data structures. I'm not too concerned with the perf here, just want to make sure its implications. If you're happy with the PR I'll do a more thorough review today.

KingoftheHomeless:
Seeing how the performance switch from `[]` to `NonDetState` doesn't seem terrible at all, as I first thought, and reflection without remorse doesn't seem like the way to go, I'll remove some of the comments, and then I'm happy with it.

KingoftheHomeless:
I'm happy with this now.

src/Polysemy/NonDet.hs
```diff
@@ -56,63 +101,48 @@ instance Monad (NonDetC m) where
...
-      z <- liftSem $ weave [()]
-                     (fmap concat . traverse runNonDet)
-                     -- TODO(sandy): Is this the right semantics?
-                     listToMaybe
-                     x
-      foldr cons nil z
+    Left x  -> consC $ fmap getNonDetState $
+      liftSem $ weave (NonDetState (Just ((), empty)))
+                  distribNonDetC
+                  -- TODO(KingoftheHomeless): Is THIS the right semantics?
```
isovector: :joy:

KingoftheHomeless:
Doh, just realized I forgot to squash and merge. I'll remember to do that in the future.

PR #215, opened 2019-08-26
Add subsume operation, expose raiseUnder/2/3

### LINKED ISSUE #212: Recursively using 'reinterpretH'
'reinterpretH' has the issue that it can't be used recursively (which the `Tactics` environment requires`) unless you also interpret the effect which it introduces.

This is fine if you immediately consume the new effect together with the use of `reinterpretH`, but can be problematic if you want to separate the reinterpretation and the consumption.

A motivating example would be a `writerToEndoWriter` reinterpreter; this could be used as a combinator for building `-AssocR` variants of `Writer` interpreters; use `writerToEndoWriter` to right-associate `tell`s, and then use the proper interpreter.

If you try to write this, you run into this problem:
```haskell
writerToEndoWriter
    :: Monoid o
    => Sem (Writer o ': r) a
    -> Sem (Writer (Endo o) ': r) a
writerToEndoWriter = reinterpretH $ \case
      Listen m -> do
        m' <- writerToEndoWriter <$> runT m
        raise $ _hole $ do
          (o, fa) <- listen m'
          return $ (,) (appEndo o mempty) <$> fa
   -- ...
```

where
```haskell
_hole :: Sem (Writer (Endo o) : Writer (Endo o) : r) (f (o, a1))
         -> Sem (Writer (Endo o) : r) (f (o, a1))
```

A solution to this would be the addition of a new primitive:
```haskell
subsume :: Member e r => Sem (e ': r) a -> Sem r a
```
and exposing it in `Polysemy` (I can implement this).

Another alternative would be not using `reinterpretH` at all, but instead use `interpretH`:

```haskell
writerToEndoWriter
    :: (Monoid o, Member (Writer (Endo o)) r) 
    => Sem (Writer o ': r) a
    -> Sem r a
```
but then, any interpreter that wants to use this, and then immediately consume the `Writer (Endo o)` effect, would need to use `raiseUnder` in order to introduce the `Writer (Endo o)` effect to the stack. `raiseUnder` is currently hidden in `Polysemy.Internal`. IIRC I also think `consumeNewEffect . reinterpretH _` is more efficient than `consumeNewEffect . interpretH _ . raiseUnder` because of optimizations `reinterpretH` does?

Personally, I'm in favor of both adding `subsume` and exposing `raiseUnder` in `Polysemy`; `subsume` so people who use `reinterpretH` don't get stuck if they run into this problem and don't know about the `interpretH` + `raiseUnder` trick; and `raiseUnder` because it's more useful in general, as it can be used together with other transformative interpreters.

If it comes down to one or the other, I prefer `raiseUnder`.

KingoftheHomeless:
This fixes #212

### DIFF
--- a/src/Polysemy/Internal/Union.hs
+++ b/src/Polysemy/Internal/Union.hs
@@ -20,6 +20,7 @@ module Polysemy.Internal.Union
   , hoist
   -- * Building Unions
   , inj
+  , injWeaving
   , weaken
   -- * Using Unions
   , decomp
@@ -227,14 +228,19 @@ weaken (Union n a) = Union (SS n) a
 
 ------------------------------------------------------------------------------
 -- | Lift an effect @e@ into a 'Union' capable of holding it.
-inj :: forall r e a m. (Functor m , Member e r) => e m a -> Union r m a
-inj e = Union (finder @_ @r @e) $
+inj :: forall e r m a. (Functor m , Member e r) => e m a -> Union r m a
+inj e = injWeaving $
   Weaving e (Identity ())
             (fmap Identity . runIdentity)
             runIdentity
             (Just . runIdentity)
 {-# INLINE inj #-}
 
+------------------------------------------------------------------------------
+-- | Lift a @Weaving e@ into a 'Union' capable of holding it.
+injWeaving :: forall e r m a. Member e r => Weaving e m a -> Union r m a
+injWeaving = Union (finder @_ @r @e)
+{-# INLINE injWeaving #-}
 
 ------------------------------------------------------------------------------
 -- | Attempt to take an @e@ effect out of a 'Union'.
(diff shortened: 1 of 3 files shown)


src/Polysemy/Internal/Union.hs
```diff
@@ -227,14 +228,19 @@ weaken (Union n a) = Union (SS n) a
 
 ------------------------------------------------------------------------------
 -- | Lift an effect @e@ into a 'Union' capable of holding it.
-inj :: forall r e a m. (Functor m , Member e r) => e m a -> Union r m a
-inj e = Union (finder @_ @r @e) $
+inj :: forall e r m a. (Functor m , Member e r) => e m a -> Union r m a
```
KingoftheHomeless: Hmm, I rearranged this to fit the other functions in `Polysemy.Internal`, but under further consideration, perhaps this breaks something in `polysemy-plugin` or some of the TH machinery? I don't think this breaks anything inside TH, since I got everything to compile and the tests to complete without issue.

KingoftheHomeless: and when I mean other functions, I apparently only mean `weaken`; `prj` has `e r a m` instead.

KingoftheHomeless: I've now also rearranged the type vars in `prj` to fit the `e r m a` scheme, if you're ok with it.

Issue #205, opened 2019-08-15
'asyncToIO' deadlocks when 'await' is used in certain scenarios


KingoftheHomeless:
`asyncToIO` may deadlock when `await` is used inside a higher-order action on a thread that is still running, and is querying the interpreter thread.

```haskell
-- This deadlocks
deadlocks1 :: IO (String, Maybe ())
deadlocks1 =
    runM
  . runWriter
  . asyncToIO
  $ do
    lock <- embed $ newEmptyMVar
    a <- async $ do
      embed $ takeMVar lock
      tell "blocked"
    censor @String id $ do
      embed $ putMVar lock ()
      await a

-- This does not deadlock
runs1 :: IO (String, Maybe ())
runs1 =
    runM
  . runWriter
  . asyncToIO
  $ do
    lock <- embed $ newEmptyMVar
    a <- async $ do
      embed $ takeMVar lock
      tell "blocked"
    embed $ putMVar lock ()
    await a
```

This is because the interpreter thread is currently interpreting the higher-order action (in this case, `censor id`), which `awaits` on the `async`:d thread, which, in turn, waits on the interpreter thread to interpret its `tell "blocked"`.

In addition, `asyncToIO` also deadlocks when `awaiting` on a thread that has died due to an exception.
```haskell
deadlocks2 :: IO ()
deadlocks2 =
      runM
    . asyncToIO
    $ do
      a <- async $ embed $ throwIO (userError "Oh dear")
      _ <- await a
      return ()
```

This is because the exception is propagated to the "main thread" (not the interpreter thread), which therefore dies, and fails to signal the interpreter thread that it has finished.

The latter issue is pretty easy to fix (simply use `mask` and `finally` in the thread you spawn in `withLowerToIO`). The former issue is much more difficult, since you basically need the interpreter thread to interpret requests as it's already interpreting a request in order to resolve the deadlock. And even if you do that, you would have the issue that these requests produce different pieces of state.

KingoftheHomeless:
`deadlocks2` doesn't need the await. Any `IO` exception is good enough.

```
deadlocks3 :: IO ()
deadlocks3 =
      runM
    . asyncToIO
    $ do
      throwIO (userError "Oh dear")
```

Which means that this deadlocks, too:
```haskell
deadlocks4 :: IO (Either () ())
deadlocks4 =
      (runM .@ lowerError)
    . asyncToIO
    $ do
      throw ()
```

KingoftheHomeless:
[...]

PR #217, opened 2019-08-28
Add Final Effect

KingoftheHomeless:
There are a few things I want to hash out with you before I declare this ready.


src/Polysemy/Async.hs
```diff
@@ -57,9 +67,35 @@ asyncToIO m = withLowerToIO $ \lower _ -> lower $
...
+-- will have local state semantics in regards to 'Async' effects
+-- interpreted this way. See 'Final'.
+--
+-- Notably, unlike 'asyncToIO', this is not consistent with
+-- 'Polysemy.State.State' unless 'Polysemy.State.runStateIORef' is used.
+-- State that seems like it should be threaded globally throughout 'Async'
+-- /will not be./
+--
+-- Use 'asyncToIO' instead if you need to interpret some stateful effect
+-- after 'Async' in other terms than 'IO'.
```
KingoftheHomeless: Slightly modified wording compared to `polysemy-zoo`. Is this too unclear?

src/Polysemy/Final.hs
```diff
@@ -0,0 +1,170 @@
+{-# LANGUAGE TemplateHaskell #-}
+module Polysemy.Final
```
KingoftheHomeless: How much of this module should be exposed in `Polysemy`?

src/Polysemy/Fixpoint.hs
```diff
@@ -12,11 +12,16 @@ import Control.Monad.Fix
...
 import Polysemy
+import Polysemy.Final
 import Polysemy.Internal.Fixpoint
 
 ------------------------------------------------------------------------------
 -- | Run a 'Fixpoint' effect purely.
 --
+-- Consider using 'fixpointToFinal' together with
+-- @'Final' 'Data.Functor.Identity'@ instead, which doesn't need to be
+-- provided a function.
```
KingoftheHomeless: Should this statement be upgraded to a full-on deprecation of `runFixpoint`?

src/Polysemy/Internal/Final.hs
```diff
@@ -0,0 +1,191 @@
...
+--     this issue altogether, as long as the initialization work is performed
+--     outside of a 'interpretH' / 'Polysemy.Final.interpretFinal'.
+--
+--   * Instead of having access to a natural transformation
+--     @forall x. 'Sem' r x -> m x@, @'Final' m@ provides a
+--     distribution through the effectful state:
+--     @forall x. f ('Sem' r x) -> m (f x)@,
+--     together with the initial state @f ()@, and an @'Inspector' f@.
+--     (See 'ThroughWeavingToFinal')
+--     This is powerful enough for most purposes.
```
KingoftheHomeless: Should this gigantic pros vs. cons compared to `lower-` interpreters stay? It made sense in `polysemy-zoo`, but does it in `polysemy` proper?

src/Polysemy/Internal/Forklift.hs
```diff
@@ -66,7 +67,7 @@ withLowerToIO action = do
   res <- embed $ A.async $ do
     a <- action (runViaForklift inchan)
                 (putMVar signal ())
-    putMVar signal ()
+          `finally` (putMVar signal ())
```
KingoftheHomeless: Pointing this out so you won't miss it; this fixes the "exception" part of #205 .

src/Polysemy/Internal/Writer.hs
```diff
@@ -0,0 +1,156 @@
...
+-- TODO(KingoftheHomeless): Research if this is more or less efficient than
+-- using 'reinterpretH' + 'subsume'
+
+-----------------------------------------------------------------------------
+-- | Transform a @'Writer' o@ effect into a  @'Writer' ('Endo' o)@ effect,
+-- right-associating all uses of '<>' for @o@.
+--
+-- This can be used together with 'raiseUnder' in order to create
+-- @-AssocR@ variants out of regular 'Writer' interpreters.
+writerToEndoWriter
```
KingoftheHomeless: Should this be exposed in `Polysemy.Writer`?

polysemy-plugin/test/ExampleSpec.hs
```diff
@@ -33,7 +34,13 @@ program = catch @CustomException work $ \e -> writeTTY ("Caught " ++ show e)
...
 
 foo :: IO (Either CustomException ())
-foo = (runM .@ lowerResource .@@ lowerError @CustomException) $ teletypeToIO program
+foo =
+    runFinal
+  . embedToFinal @IO
+  . resourceToIOFinal
+  . errorToIOFinal @CustomException
+  . teletypeToIO
+  $ program
```
KingoftheHomeless: This stings. The tests for `polysemy-plugin` use the now deprecated `lower-` interpreters, so they'd generate warnings without this, but this also requires a dependency bump for polysemy-plugin from `polysemy>=0.6` to `polysemy>=1.2`.

I'm not sure what the best solution to this is.

KingoftheHomeless:
Should I add `nonDetToFinal` from `polysemy-zoo`? I feel it's borderline.

src/Polysemy/Output.hs
```diff
@@ -116,6 +118,44 @@ runOutputMonoidTVar tvar f = interpret $ \case
...
 
+
+--------------------------------------------------------------------
+-- | Run an 'Output' effect in terms of atomic operations
+-- in 'IO'.
+--
+-- Internally, this simply creates a new 'IORef', passes it to
+-- 'runOutputMonoidIORef', and then returns the result and the final value
+-- of the 'IORef'.
+outputToIOMonoid
```
KingoftheHomeless: Since `outputToIOMonoid` uses an `IORef`, it means it doesn't respect local/global state semantics depending on the order of interpreters run. Should a _Beware_ be added for that?

src/Polysemy/Final.hs
KingoftheHomeless: My thinking is that I expose only the datatype, `runFinal`, `embedToFinal`, and `embedFinal`, together with adding a tip about the `Polysemy.Final` module for more advanced uses of `Final`.

src/Polysemy/Async.hs
KingoftheHomeless: I'll improve this. It may make it a bit more verbose, but it'll be worth it.

src/Polysemy/Fixpoint.hs
KingoftheHomeless: Ok

src/Polysemy/Internal/Final.hs
KingoftheHomeless: Ok

polysemy-plugin/test/ExampleSpec.hs
KingoftheHomeless: Oh yeah, there are separate dependencies for the tests. Then there's no issue.

src/Polysemy/Internal/Final.hs
```diff
@@ -0,0 +1,191 @@
...
+
+-----------------------------------------------------------------------------
+-- | An effect for embedding higher-order actions in the final target monad
+-- of the effect stack.
+--
+-- This is very useful for writing interpreters that interpret higher-order
+-- effects in terms of the final monad.
+--
+-- 'Final' is slightly weaker, but using it is otherwise more preferable
+-- than using a provided lowering function (a la 'Polysemy.Async.lowerAsync'):
```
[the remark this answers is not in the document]
KingoftheHomeless: Probably not. `Final` _is_ slightly weaker than a lowering function, so it's half-imaginable that people would create an interpreter that would need the extra power of a lowering function rather than `Final`.
Thing is, the restrictions placed by `Final` are already placed by the need to thread functorial state from previous interpreters through (for example, you can't do a `UnliftIO` interpreter even with a lowering function). I can't think of one example where an explicit lowering function would give you the power to do something `Final` can't.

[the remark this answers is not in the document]
KingoftheHomeless: I removed the comment.

src/Polysemy/Output.hs
KingoftheHomeless: Done

package.yaml
```diff
@@ -1,5 +1,5 @@
 name:                polysemy
-version:             1.1.0.0
+version:             1.2.0.0
```
KingoftheHomeless: Version bump in order to update dependencies for polysemy-plugin-test.

src/Polysemy/Async.hs
```diff
@@ -57,9 +69,37 @@ asyncToIO m = withLowerToIO $ \lower _ -> lower $
...
+--
+-- Notably, unlike 'asyncToIO', this is not consistent with
+-- 'Polysemy.State.State' unless 'Polysemy.State.runStateIORef' is used.
+-- State that seems like it should be threaded globally throughout 'Async'
+-- /will not be./
+--
+-- Use 'asyncToIO' instead if you need to run
+-- pure, stateful interpreters after the interpreter for 'Async'.
+-- (Pure interpreters are interpreters that aren't expressed in terms of
+-- another effect or monad; for example, 'Polysemy.State.runState'.)
```
KingoftheHomeless: Is this any better? It's the wording from `polysemy-zoo`, but with clarification on what is meant by "pure".

KingoftheHomeless: Good!

ChangeLog.md
```diff
@@ -1,5 +1,34 @@
 # Changelog for polysemy
 
+
+## 1.2.0.0 (TODO)
```
KingoftheHomeless: Marked the date as TODO here since I figure we might want to add more stuff before we publish this to Hackage.

KingoftheHomeless:
Pretty much ready now. It's a bit silly; I figured it'd be a bit longer before we were ready to merge this, which is why I first imported the final rework to `polysemy-zoo`.  Now it turns out perhaps that version won't even reach `polysemy-zoo` on Hackage.

KingoftheHomeless:
I'm gonna stew on this for a day to make sure I'm satisfied with it before I merge it in.

KingoftheHomeless:
A question in my mind is if we shouldn't just have `runM` process both `Embed` and `Final`, like the old version of `runFinal`:
```haskell
runM :: Monad m => Sem '[Embed m, Final m] a -> m a
```
It's backwards-compatible (except when people explicitly write out their effect stack instead of using `Members`), and it means users won't have to juggle between `runM` and `runFinal . embedToFinal`.

KingoftheHomeless:
Oh, I should ping you too about that. It's the only unresolved question I have about this, with the exception of potential performance issues. @isovector

KingoftheHomeless:
For the record: the performance issue I'm worried about is how `loopbreaker` interacts with `Final`; `Final` works by using the distribution function provided by a weaving, the one that is generated by the
`hoist runInterpreter g` and `weave ... g` that interpreters do. The problem is that `loopbreaker` will, instead of injecting `runInterpreter` into the weaving, instead inject the loop-breaker `runInterpreter_b`, which will mean that when the distribution function is used, it isn't optimized.

It that's indeed the case, `errorToIOFinal` would get hit hard, as it takes all runs the distribution function on it in order to attempt it's `try` at the end. `lower-` interpreters don't have this issue since the user-supplied function will use the proper interpreters, and not the loop-breakers.

KingoftheHomeless:
@TheMatten Does that sound right to you? (As in, is it plausible that's indeed what would happen?)

isovector:
I'd prefer to keep `runM` as it is today, at least until 8.10 lands and we can better judge performance without needing to use our brains.

KingoftheHomeless:
Oh, the performance question was separate from the question about `runM`; `runM` was entirely about user convenience.

isovector:
I guess I'm still not sure what `Final` buys us that `interpret`/`embed` doesn't, except for continuations?

src/Polysemy/Final.hs
```diff
@@ -174,24 +169,24 @@ withStrategicToFinal strat = withWeavingToFinal $ \s wv ins ->
...
-                  (n e)
-      Left g -> liftSem (hoist go g)
+        injWeaving $
+          Weaving
+            (WithWeavingToFinal (runStrategy (n e)))
+            s
+            (go . wv)
+            ex
+            ins
+      Left g -> hoist go g
```
KingoftheHomeless: Found this neat optimisation made possible from the new `injWeaving`, which cleans up the code to boot!

TheMatten:
@KingoftheHomeless having some simple example that we can bench and `-ddump-simpl` would be nice. But honestly, `loopbreaker` is going to be rewritten to instead do Core pass for better reliability (it just wasn't needed for now and I forgot about it :slightly_smiling_face:) - this will probably change how it behaves.

KingoftheHomeless:
`Final` let's us not need lowering functions for interpreting higher-order effects in terms of the final monad! That's really the main point of it all.

isovector:
Ahh, right. Sorry, not sure why I can't seem to remember that. I'm definitely sympathetic to the idea, though why do we need the `Embed` in `runM :: Monad m => Sem '[Embed m, Final m] a -> m a`. Just for backwards compatibility?

TheMatten:
@isovector As I understand it, this would let us use `runM` for both `Embed` and `Final`

KingoftheHomeless:
Partly backwards compatibility, partly user convenience. Since `Embed` is more flexible to interpret than `Final`, we don't want to simply replace `Embed` with `Final`; in fact, we want interpreters to use `Embed` instead when they don't need the extra power. So this means that at the end, you'll end up with `Embed m` and/or `Final m`. `runM` can be used under any circumstance here; even if you haven't made use of `Embed m`, or haven't made use of `Final m`. So it's a nice one-interpreter-fits-all.

isovector:
That's a pretty good argument. I'm on board, but let's hold off for a 2.0 release.

KingoftheHomeless:
Ok! That's reasonable to me.

I'll go ahead and merge this, then.

PR #218, opened 2019-08-31
Fix mistakes in Final, add atomic/StateToIO

KingoftheHomeless:
I won't merge this yet in case I find something more (which I doubt).

@isovector Don't hesitate to merge this when you're ready to publish v1.2.


KingoftheHomeless:
This turned into more than fixes, now: I remembered that I intended to add `stateToIO` and `atomicStateToIO` since with `Final` we no longer have to worry about work duplication. Of course, I then promptly forgot about it in the `Final` pull request.

KingoftheHomeless:
@isovector If you're fine with this I'll merge it

PR #221, opened 2019-09-01
polysemy-plugin: Reject ununifiable effect candidates early

KingoftheHomeless:
This is the improvement I described in #219. 

Any other improvements may be a long way off, if I even find a way to do them, so I figured I'd open this pull request separately for the easy fix.

### DIFF
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -116,12 +116,12 @@ mkWantedForce
   -> Type
   -> TcPluginM (Unification, Ct)
 mkWantedForce fc given = do
-    (ev, _) <- unsafeTcPluginTcM
-             . runTcSDeriveds
-             $ newWantedEq (fcLoc fc) Nominal wanted given
-    pure ( Unification (OrdType wanted) (OrdType given)
-         , CNonCanonical ev
-         )
+  (ev, _) <- unsafeTcPluginTcM
+           . runTcSDeriveds
+           $ newWantedEq (fcLoc fc) Nominal wanted given
+  pure ( Unification (OrdType wanted) (OrdType given)
+       , CNonCanonical ev
+       )
   where
     wanted = fcEffect fc
 
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -88,16 +88,17 @@ getFindConstraints (findClass -> cls) cts = do
 
 ------------------------------------------------------------------------------
 -- | If there's only a single @Member@ in the same @r@ whose effect name
--- matches, return its effect (including tyvars.)
+-- matches and could possibly unify, return its effect (including tyvars.)
 findMatchingEffectIfSingular
     :: FindConstraint
     -> [FindConstraint]
     -> Maybe Type
-findMatchingEffectIfSingular (FindConstraint _ eff_name _ r) ts =
+findMatchingEffectIfSingular (FindConstraint _ eff_name wanted r) ts =
   singleListToJust $ do
     FindConstraint _ eff_name' eff' r' <- ts
     guard $ eqType eff_name eff_name'
     guard $ eqType r r'
+    guard $ canUnifyRecursive FunctionDef wanted eff'
     pure eff'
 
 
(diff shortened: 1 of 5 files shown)


polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
```diff
@@ -202,8 +215,9 @@ solveFundep (ref, stuff) given _ wanted = do
     let r  = fcRow fc
     case findMatchingEffectIfSingular fc given_finds of
```
KingoftheHomeless: Potential bug source? Previously, if `findMatchingEffectIfSingular` finds an single, ununifiable effect, then the `Just` branch would get taken, and then `mkWanted` would reject making an equality constraint as the effect is ununifiable.
With my modifications, the ununifiable effect would get rejected in `findMatchingEffectIfSingular`, and the `Nothing` path will get taken instead.

KingoftheHomeless: Or perhaps that's actually an improvement?

polysemy-plugin/test/MultipleVarsSpec.hs
```diff
@@ -0,0 +1,28 @@
...
+        ] r
+      => Sem r ()
+test = taggedPut @Bool 10
+
+
+spec :: Spec
+spec = describe "Using multiple, but ununifiable instances\
+               \ of the same effect" $ do
+  it "should get disambiguated and compile!" $ do
+    True `shouldBe` True
```
isovector: Can you add a test that ensures it also works without an explicit `Given`? ie

```haskell
flip shouldBe blah . run . runTagged @Bool . runTagged @Char $ do
  taggedPut @Bool 10
```

isovector: Also I'd like to see actually running `test`, to ensure the tagged effects end up in the right place. Right now we just give evidence by fiat, which will happily compile even if it _does the wrong thing_ :scream:

polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
```diff
@@ -202,8 +215,9 @@ solveFundep (ref, stuff) given _ wanted = do
...
       -- We found a real given, therefore we are in the context of a function
-      -- with an explicit @Member e r@ constraint.
-      Just eff' -> mkWanted fc FunctionDef eff'
+      -- with an explicit @Member e r@ constraint. We also know it can
+      -- be unified (although it may generate unsatisfiable constraints).
```
isovector: What do those unsatisfiable constraints look like? Can we add some tests for multiple states and see what happens in that case? I'm particularly curious about two concrete `Num`s.

KingoftheHomeless: What I mean by unsatisfiable constraints in this context is `Num String`; in particular, I don't mean obviously unsatisfiable constraints like `String ~ Int`.

For example, `polysemy-plugin` will generate a `Num String` constraint for code like this:

```haskell
ex :: Member (State String) r => Sem r ()
ex = put 10
```

With or without my modification, because it sees the type variable created by the literal, and `canUnifyRecursive` says "oh hey, we can unify string with that".

All my modification does is use `canUnifyRecursive` to get rid of cases we know unification won't work out, even without knowing what constraints we are given.

I'm not sure what you mean about two concrete `Num`s.

polysemy-plugin/test/MultipleVarsSpec.hs
KingoftheHomeless: Just tested it. Works fine; I'll commit the update to the test together with any other changes you'd like.

isovector:
What's the result of the following?

```
foo :: Members '[State Int, State Bool] r => Sem r ()
foo = do
  put 5
  put True
```

What about this?

```
foo :: Members '[State Int, State Double] r => Sem r ()
foo = do
  put 5
  put 6.0
```

KingoftheHomeless:
Both would fail to disambiguate.
In the first case, the type variable generated by `put 5` _could_ unify with either `Int` or `Bool` (despite the fact that a `Num Bool` constraint/instance isn't given).

Similarly, either type variable generated in the second case could be unified with either `Int` or `Double` (despite the fact that a `Fractional Int` constraint/instance isn't given).

So my improvement doesn't fix #219. But it does allow this:
```
foo :: Members '[TaggedState "one" Int, TaggedState "two" Double] r => Sem r ()
foo = do
  taggedPut @"one" 5
  taggedPut @"two" 6.0
```
Which wasn't possible before.

That's because `TaggedState "one" ambiguous_variable` generated by `taggedPut @"one" 5` could only unify with `TaggedState "one" Int`, and not `TaggedState "two" Double`.

isovector:
Yeah, the tagged stuff makes sense. But I'm curious about the actual behavior of the plugin; does it simply not do any work in my example, or does it do something silly like try to unify `State Int ~ State Bool`?

KingoftheHomeless:
`findMatchingEffectIfSingular` will find that it has two candidates when trying to disambiguate `put 5` even with pruning using `canUnifyRecursive`, so it gives up; the resulting error message is:
```
 * Ambiguous use of effect 'State'
      Possible fix:
        add (Member (State Integer) r) to the context of
          the type signature
```

So the behaviour in that case is the same as it used to be.

PR #224, opened 2019-09-06
Add type synonym 'InterpreterFor' and documentation

bolt12:
Hi,

This PR tackles issue #220 . I've put the type synonym on the Polysemy.Internal module and documented it.

Please let me know if I should improve anything.

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -136,11 +136,11 @@ module Polysemy
   , Inspector (..)
   ) where
 
-import           Polysemy.Final
-import           Polysemy.Internal
-import           Polysemy.Internal.Combinators
-import           Polysemy.Internal.Forklift
-import           Polysemy.Internal.Kind
-import           Polysemy.Internal.Tactics
-import           Polysemy.Internal.TH.Effect
+import Polysemy.Final
+import Polysemy.Internal
+import Polysemy.Internal.Combinators
+import Polysemy.Internal.Forklift
+import Polysemy.Internal.Kind
+import Polysemy.Internal.Tactics
+import Polysemy.Internal.TH.Effect
 
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -9,15 +9,15 @@ module Polysemy
   , runM
   , runFinal
 
+  -- * Type synonyms for user convenience
+  , InterpreterFor
+
   -- * Interoperating With Other Monads
   -- ** Embed
   , Embed (..)
   , embed
   , embedToFinal
 
-  -- * Type synonyms for user convenience
-  -- ** InterpreterFor
-  , InterpreterFor
   -- ** Final
   -- | For advanced uses of 'Final', including creating your own interpreters
   -- that make use of it, see "Polysemy.Final"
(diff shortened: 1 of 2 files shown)


src/Polysemy/Internal.hs
```diff
@@ -434,6 +435,18 @@ runM (Sem m) = m $ \z ->
...
+-- | Type synonym for interpreters that consume an effect without changing the 
+-- return value. Offered for user convenience. 
+--
+-- 'r' Is kept polymorphic so it's possible to place constraints upon it:
+--
+-- @
+-- teletypeToIO
+-- :: 'Member' (Embed IO) r
+-- => 'InterpreterFor' Teletype r
+-- @
```
KingoftheHomeless: I _think_ you need another comment line here to make the code block render properly. I don't quite get how Haddock works on that point. Also, the indentation is a bit wrong here.

To illustrate what I mean, here's how I'd do it:

```haskell
-- @r@ Is kept polymorphic so it's possible to place constraints upon it:
--
-- @
-- teletypeToIO
--   :: 'Member' (Embed IO) r
--   => 'InterpreterFor' Teletype r
-- @
--
```

src/Polysemy.hs
```diff
@@ -14,6 +14,10 @@ module Polysemy
   , Embed (..)
   , embed
   , embedToFinal
+
+  -- * Type synonyms for user convenience
+  -- ** InterpreterFor
+  , InterpreterFor
```
KingoftheHomeless: This is exported in the wrong place; in the middle of `Interoperating with other monads`, which will make the documentation look weird. Instead, place this after the export of `runFinal`.

Also, you don't need the `InterpreterFor` subheading.

src/Polysemy.hs
```diff
@@ -132,11 +136,11 @@ module Polysemy
...
-import Polysemy.Internal.TH.Effect
-import Polysemy.Internal.Tactics
-import Polysemy.Final
+import           Polysemy.Final
+import           Polysemy.Internal
+import           Polysemy.Internal.Combinators
+import           Polysemy.Internal.Forklift
+import           Polysemy.Internal.Kind
+import           Polysemy.Internal.Tactics
+import           Polysemy.Internal.TH.Effect
```
KingoftheHomeless: I'd say the extra indentation isn't needed here; I doubt we'll ever use quantified modules in `Polysemy`. The reordering is good, though.

bolt12: Oops! My setup indents imports by default, I forgot to turn this off. I will revert this changes!

src/Polysemy.hs
bolt12: Okay Thanks! I wasn't sure where to export it. I will correct this!

src/Polysemy/Internal.hs
bolt12: I have generated the docs and everything looked alright! I'll make this change and see how it looks!

KingoftheHomeless:
@isovector When it comes to simple pull requests like these, do you also want to take a look at them, or can I merge them at my discretion?

isovector:
@KingoftheHomeless i trust you! thanks for taking the initiative on this!

KingoftheHomeless:
Thanks for your work @bolt12!

bolt12:
@KingoftheHomeless No problem! My pleasure, I hope to continue contributing :)

PR #241, opened 2019-10-05
Strengthen Strategy

KingoftheHomeless:


isovector:
Cool. So I think we should start thinking about replacing `runM` with `runStrategy` and just get all of this stuff for free.

isovector:
Also, mind giving me a high-level overview of wtf is going on here? What does this buy us?

KingoftheHomeless:
Previously, the `Strategic` environment only allowed us to embed actions of the final monad at the very end; as part of the action you return.

Now, you can embed actions of the final monad whereever in `Strategic`. This means you can use `Strategic` in places where it previously wasn't possible. For example, I was now able to rewrite `runWriterSTMAction` to use `withStrategicToFinal` instead of `withWeavingToFinal`, making that code a lot more prettier.

By "start thinking about replacing `runM` with `runStrategy`", do you mean having `runM :: Monad m => Sem '[Embed m, Final m] a -> m a` like I suggested a bit back?

src/Polysemy/Internal/Strategy.hs
```diff
@@ -67,8 +67,8 @@ runStrategy sem = \s wv ins ->
 -- | Embed a value into 'Strategic'.
 --
 -- @since 1.2.0.0
-pureS :: forall m n a. Applicative m => a -> Strategic m n a
-pureS a = (a <$) <$> getInitialStateS
+pureS :: a -> Strategic m n a
+pureS = pureT
```
isovector: :open_mouth:

KingoftheHomeless:
Also, note that this is into `no-more-strategy`. Considering that `no-more-strategy` is already for the v2.0 release, the breakage stemming from the change to `Strategic` is acceptable. I'll merge this into that branch if you have no issue with this.

isovector:
love it, thanks!

PR #242, opened 2019-10-05
Delete deprecated

isovector:
fuck deprecated code!!

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -113,10 +113,6 @@ module Polysemy
   , Effect
   , EffectRow
 
-    -- * Composing IO-based Interpreters
-  , (.@)
-  , (.@@)
-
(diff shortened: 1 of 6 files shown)


KingoftheHomeless:
`runFixpoint` should go too, right?

isovector:
Yeah, it should, but the comment scared me about how we need the `runIdentity` bit afterwards.

isovector:
Can we write a `runPure` with that is like the new `runM` in #239 , except specialised to `Identity`? Then this thing would just work?

KingoftheHomeless:
We had a previous discussion about this when I did the `Final` rework at `polysemy-zoo`.
https://github.com/polysemy-research/polysemy-zoo/pull/45#discussion_r316719585

PR #250, opened 2019-10-10
Add a convenience atomicGets

googleson78:
Is there something wrong with having this?

Notably right now the "get" operation is atomic, but the function application isn't. I think this is the expected/desirable behaviour? Or should we go the other way around and have the function also be "atomic" by doing `snd <$> atomicState (id &&& f)`?

### DIFF
--- a/src/Polysemy/AtomicState.hs
+++ b/src/Polysemy/AtomicState.hs
@@ -49,8 +49,6 @@ atomicGet :: forall s r
            . Member (AtomicState s) r
           => Sem r s
 
--- | Only the \"get\" part of this operation is atomic - 
--- the function call doesn't block other consumers of the state.
 atomicGets :: forall s s' r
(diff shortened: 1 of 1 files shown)


KingoftheHomeless:
This is good!
You're right; there's no need for the function call to be blocking.

KingoftheHomeless:
I don't think the documentation is needed; I feel it will confuse people more than it helps them ("what do they mean that only the `get` is atomic? Is this unsafe to use?")

src/Polysemy/AtomicState.hs
```diff
@@ -48,6 +49,15 @@ atomicGet :: forall s r
            . Member (AtomicState s) r
           => Sem r s
 
+-- | Only the \"get\" part of this operation is atomic -
```
KingoftheHomeless: The documentation is unneeded

PR #254, opened 2019-10-17
Added a type synonym Before to constrain order of effect interpretation

SamuelSchlesinger:
It is wonderful that we can separate our syntax from our semantics, and generalize over effect membership in a stack, but there are a number of cases in which this such extreme generality is undesirable. For instance, depending on where you put an `Error CustomException` in your stack, your code will have very different behavior. In fact, there are codebases you could ruin with a 2 line diff, by swapping 2 interpreters and running the same polymorphic code.

To fix this, I made the constraint `Before e1 e2 r` to enforce the fact that `e1` is interpreted in your effect stack `r` before `e2` is. Some other things one might want to add before merging this is `Firstly e r`, `Lastly e r`, `BeforeAll e1 es r`, `After e1 e2 r`, `AfterAll e1 es r`, but I have a work obligation super early and I don't feel like implementing, documenting, and testing all of that at the moment. Feel free to let me know if you are interested in adding this into the library and I can add these things and whatever else we want before merging it in.

Note: It may be interesting to see what you can do with QuantifiedConstraints in combination with this... Maybe nothing? I dunno, could be fun.

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -4,6 +4,9 @@ module Polysemy
   , Member
   , Members
   , Before
+  , After
+  , Firstly
+  , Lastly
 
   -- * Running Sem
   , run
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -3,6 +3,7 @@ module Polysemy
     Sem ()
   , Member
   , Members
+  , Before
 
   -- * Running Sem
   , run
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -11,6 +11,9 @@ module Polysemy.Internal
   , Member
   , Members
   , Before
+  , After
+  , Firstly
+  , Lastly
   , send
   , embed
   , run
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -10,6 +10,7 @@ module Polysemy.Internal
   ( Sem (..)
   , Member
   , Members
+  , Before
   , send
   , embed
   , run
--- a/src/Polysemy/Internal/Union.hs
+++ b/src/Polysemy/Internal/Union.hs
@@ -16,6 +16,7 @@ module Polysemy.Internal.Union
   , Weaving (..)
   , Member
   , MemberWithError
+  , Before
   , weave
   , hoist
   -- * Building Unions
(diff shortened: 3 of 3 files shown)


SamuelSchlesinger:
Another interesting bit to add might be `r \`Can't\` e`, to say that the context `r` cannot have the effect `e`. Might constrain some idiot like me from using the `FireMissiles` effect more often than he should in some production code.

SamuelSchlesinger:
> Where do you imagine introducing such constraints would be most valuable? My guess is on the interpreters, but I can't wrap my head around how this would behave.

I think I might use it in the original effectful program description to specify more semantically accurate constraints on the effect stack. I more or less never want someone to run code with my `Logic` backtracking effect and my `State Stuff` effect in the wrong order, if I need my state to get reset when we backtrack, or if I need to keep it. 

It will force whoever is interpreting that program to do it in a way that actually respects my meaning. The type errors could be improved for this, at the moment it tells you that there is no instance for `Compare Logic (State Stuff) LT`, and `Compare` is not exposed so that is not a great situation.

SamuelSchlesinger:
Looking at some of the interpreters, it seems that there are many cases in which you would like to use it in the interpreters as well. I am thinking that there could be a something like `AtomicState` which allows rollbacks that you could actually implement in terms of a `Logic` effect and a `State` effect, but that only makes sense in a particular order.

SamuelSchlesinger:
I added `Firstly`, `Lastly`, and `After`. I defined `After` as it's own thing instead of in terms of `Before`, because it might be nice to eventually have custom error messages for each.

KingoftheHomeless:
Although I like this, here's where the problem of transformative interpreters rears its ugly head.
The order of effects is only a soft constraint on how effects interact, since you can always interpret an effect in terms of a later effect; given the stack `[A,B,C]`, `A` could still have global state semantics with regards to `B` if `A` is interpreted in terms of  `C`.
The typical scenario where this happens are effects interpreted using `Embed IO` and `Final IO`, which is why such transformers have _Beware_s attached to them. I've been trying to figure out a way to restrict when you use such interpreters so you can't fool yourself into a "gotcha", without sacrificing composability. I haven't found a perfect solution yet.

SamuelSchlesinger:
Yeah, I was actually dealing with this last night, and I came on to comment that I am not sure this works as I naively thought. I implemented a version of "first class monad transformer stacks" insteas which might give me what I want, but I am not sure the story on type inference and class lifting in the style of mtl will work out there as it does in polysemy. 

I would say that what I have defined here will behave in ways that are so counterintuitive that the PR in it's current form should probably be closed. Agreed?

KingoftheHomeless:
It hurts, but yeah. The ideal **is** that effect ordering should dictate semantics between effects; but we haven't managed to reach that ideal yet, and thus you can't use these constraints confidently.

If we figure out how to restrict the placement of transformative interpreters, then I'd love to have this in!

SamuelSchlesinger:
> If we figure out how to restrict the placement of transformative interpreters, then I'd love to have this in!

At least I now know where to place my effort to get this thing. I have been spending some time trying to hack together a first class Monad transformer stack with an explicit base monad, trying to get what I want there. If you are interested in seeing that, I could post some of it here. I have gotten some stuff to compile, now its off to see how the ergonomics work out.

SamuelSchlesinger:
Here are the language extensions required for getting what I am working on to compile 😆 

```
1   {-# LANGUAGE MultiParamTypeClasses #-}
  1 {-# LANGUAGE AllowAmbiguousTypes #-}
  2 {-# LANGUAGE UndecidableInstances #-}
  3 {-# LANGUAGE DeriveFunctor #-}
  4 {-# LANGUAGE ScopedTypeVariables #-}
  5 {-# LANGUAGE RankNTypes #-}
  6 {-# LANGUAGE PolyKinds #-}
  7 {-# LANGUAGE DataKinds #-}
  8 {-# LANGUAGE FlexibleContexts #-}
  9 {-# LANGUAGE FlexibleInstances #-}
 10 {-# LANGUAGE ConstraintKinds #-}
 11 {-# LANGUAGE TypeApplications #-}
 12 {-# LANGUAGE TypeFamilies #-}
 13 {-# LANGUAGE QuantifiedConstraints #-}
 14 {-# LANGUAGE DataKinds #-}
 15 {-# LANGUAGE TypeOperators #-}
```

SamuelSchlesinger:
Urgh, some horrid side effects (har) of the n^2 instance problem neuters the member constraints... I guess I will wait for this thing to be worked out in polysemy.

Issue #235, opened 2019-10-01
Let's build polysemy with latest stackage and ghc 8.8


googleson78:
Not sure if I'm acting out of place or there are some other considerations which I'm missing (in which cases, sorry and feel free to close!) but it would be nice to have `polysemy` in an lts/nightly with ghc 8.8.

I tried to build `polysemy` with `nightly-2019-09-30` and found the following things to do:

* [x] get `loopbreaker` to work with ghc 8.8 - https://github.com/polysemy-research/loopbreaker/issues/6
* [x] depend on `first-class-families` `0.6.0.0`
We can either bump the version in `polysemy`'s `stack.yaml`, or we can depend on the nightly stackage's `first-class-families` (which coincidentally is exactly `0.6.0.0` in `nightly-2019-09-30`.
* [x] depend on a version `>=  0.0.1.0` of `type-errors-pretty`, because it has a `base` restriction before that
It hasn't been uploaded to Hackage yet, however. https://github.com/chshersh/type-errors-pretty/issues/6
* [x] depend on a later `unagi-chan` - `0.4.1.0` has an upper bound on `primitive`, which is later removed
a `base` restriction of `< 4.13` is however added in [Revision 1](https://hackage.haskell.org/package/unagi-chan-0.4.1.2/revisions/) - I don't really understand this, but we can either depend on `0.4.1.2:rev0`, which doesn't have the restriction, or wait for the situation there to be resolved (?)

TheMatten:
@isovector Do we just want to jump on current (at the moment nightly) snapshots from now on, or wait for LTS? Or create separate configuration?

Tehnix:
Don't know if this is helpful at all, but it seems both `unagi-chan` and `type-errors-pretty` are working if you use them from GitHub HEAD.

The following stack.yaml currently works for me with GHC 8.8.1:

```
resolver: nightly-2019-10-14
packages:
- .
extra-deps:
- polysemy-1.2.1.0
- polysemy-plugin-0.2.3.0
- loopbreaker-0.1.1.1
- github: jberryman/unagi-chan
  commit: 671ee5dc4c6894a51bc216bdb18fa21e42e4cdba
- github: chshersh/type-errors-pretty
  commit: 93f7f90c6b956bd5ab59645746c4a78e3f8d1676
```

isovector:
excellent, thanks! i'd love to get this merged.

chshersh:
[...]

PR #258, opened 2019-10-18
Try to pass travis with 8.8.1

googleson78:
Am I doing this right?
(progress toward #235 (hopefully))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
   directories:
     - $HOME/.cabal/store
 
-cabal: "2.4"
+cabal: "3.0"
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ script:
 # Check that the resulting source distribution can be built & installed.
 # If there are no other `.tar.gz` files in `dist`, this can be even simpler:
 # `cabal install --force-reinstalls dist/*-*.tar.gz`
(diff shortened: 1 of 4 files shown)


isovector:
[And now we wait](https://travis-ci.org/polysemy-research/polysemy/jobs/599545192) and find out!

googleson78:
```
[__1] trying: unagi-chan-0.4.1.2 (dependency of polysemy)
```
Why won't it use `unagi-chan-0.4.1.3` :thinking:

googleson78:
Well.. the build is passing now, after the bump to cabal 3, but it seems the doctest that depends on the plugin is broken? Help?

isovector:
I'm not sure offhand. I'll try digging in later today!

googleson78:
I think `type-error` also needs a bump, currently looking into that.

googleson78:
Ok, I think `doctest` is broken with `cabal 3.0`/`ghc 8.8.1`

isovector:
If this is the only thing preventing 8.8.1, I'd be happy to CPP those tests out and we can figure out a solution later.

googleson78:
Well I'm obviously just flailing about right now:

Entirely disabling the two doctests (by commenting them out!) work. I created an issue at `doctest` too - https://github.com/sol/doctest/issues/245

But then the `cabal install polysemy*.tar.gz` fails. Perhaps newer `cabal install` invocations don't support `.tar.gz`s?

isovector:
Thanks for your work on this, @googleson78 . I just merged #267 which I think fixes it!

PR #265, opened 2019-10-25
Use cabal-doctest

sevanspowell:
- Haskell build tools run in slightly different environments (meaning different
  package databases are available).
- The nixpkgs build for polysemy-plugin is failing due to a missing package
  database, which causes the doctest to fail (more information here:
  https://github.com/NixOS/nixpkgs/issues/71164).
- By using cabal-doctest we can expose the Haskell packages required to the
  doctests no matter the build tool we're using (claim not tested on everything).

### DIFF
--- a/polysemy-plugin/Setup.hs
+++ b/polysemy-plugin/Setup.hs
@@ -1,2 +1,3 @@
-import Distribution.Simple
-main = defaultMain
+import Distribution.Extra.Doctest (defaultMainWithDoctests)
+
+main = defaultMainWithDoctests "polysemy-plugin-test"
--- a/polysemy-plugin/polysemy-plugin.cabal
+++ b/polysemy-plugin/polysemy-plugin.cabal
@@ -18,7 +18,7 @@ maintainer:     sandy@sandymaguire.me
 copyright:      2019 Sandy Maguire
 license:        BSD3
 license-file:   LICENSE
-build-type:     Simple
+build-type:     Custom
 extra-source-files:
     README.md
     ChangeLog.md
@@ -26,6 +26,12 @@ extra-source-files:
 source-repository head
   type: git
   location: https://github.com/isovector/polysemy
+
+custom-setup
+    setup-depends:
+        base
+      , Cabal
+      , cabal-doctest >=1.0.6 && <1.1
 
 library
   exposed-modules:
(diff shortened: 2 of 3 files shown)


googleson78:
When trying to build with cabal 3.0/ghc 8.8.1 the doctests also fail with in a similar (to me) vein with a

```
  test/DoctestSpec.hs:9:47:
  1) Doctest, Error messages, should pass the doctest
       uncaught exception: GhcException
       <command line>: cannot satisfy -package type-errors
           (use -v for more information)
```

Is it actually a similar issue (that could perhaps be remedied with `cabal-doctest`)?

sevanspowell:
@googleson78 Sounds like it could be! I’ll try to reproduce your failure, but do you get the same failure with my modifications?

sevanspowell:
@googleson78 I'm just having a play around with your issue here: https://github.com/polysemy-research/polysemy/pull/267

sevanspowell:
Current issue is that it doesn't seem the new cabal version supports install ".tar.gz" files in the same way.

googleson78:
Yeah that's what I hit also in #258 , but before that I had to entirely comment out the doctests.

If you want an easy(ier?) repro the `type-errors` package also doesn't work with cabal 3/ghc881, the way it is currently set up when you run `cabal v2-test`.

googleson78:
I did indeed try to use your modifications, and I was still getting the same failures though!
Sorry for not commenting earlier!

sevanspowell:
I was able to reproduce the doctest issue locally. The changes I made were only to polysemy-plugin, so I've expanded them to apply to polysemy too (https://github.com/polysemy-research/polysemy/pull/267). I've found that the new source dist. is located in ("dist-newstyle/sdist"), but cabal install doesn't seem to support ".tar.gz" files anymore, as you discovered.

sevanspowell:
@googleson78 Aha! Got it.

If we just use "cabal v1-install dist-newstyle/sdist/..." it'll pass!

googleson78:
Cool! Thanks a lot! I'm still not sure about the intent behind this check though - if cabal no longer uses `v1-install` by default, shouldn't the check be different/gone? :thinking:  @isovector

sevanspowell:
@googleson78 Agreed, I have no idea behind it's intent.

I've got to head off now, feel free to take the baton <3

sevanspowell:
I need to apply these changes to the `polysemy` target too, currently it's only applied to `polysemy-plugin`. I've already done this here https://github.com/polysemy-research/polysemy/pull/267 so I should consolidate these PRs.

isovector:
I merged #267  . Thank you both for your hard work here!

PR #269, opened 2019-10-31
Law Machinery

### LINKED ISSUE #266: Add machinery for laws for effects
I've been thinking about how to prove interpreters are correct. Maybe this:

```haskell
class HasLaws (e :: Effect) where
  laws :: Member e r => [exists a. Eq a => (Gen (Sem r a, Sem r a))]
```

which states that for each element of the list, the two programs should evaluate to the same `a` under any interpreter (obviously using the impredicativty trick for `laws`)

For example:

```haskell
instance Eq s => HasLaws (State s) where
  laws = 
    [ do  -- overwriting of put
        s <- arbitrary
        s' <- arbitrary
        pure (put s >> put s' >> get, put s' >> get)

    , -- more
    ]
```

isovector:
This PR adds machinery for writing easy laws that should hold for an effect. It gives an example of what they'd look like for `State s`.

Fixes #266

### DIFF
--- a/src/Polysemy/Law.hs
+++ b/src/Polysemy/Law.hs
@@ -9,6 +9,7 @@ module Polysemy.Law
   , runLaw
   , MakeLaw (..)
   , Citizen (..)
+  , printf
   , module Test.QuickCheck
   ) where
 
@@ -18,14 +19,19 @@ import Polysemy
 import Test.QuickCheck
 
 
+------------------------------------------------------------------------------
+-- | Associates the name @r@ with the eventual type @a@. For example,
+-- @'Citizen' (String -> Bool) Bool@ can produce arbitrary @Bool@s by calling
+-- the given function with arbitrary @String@s.
 class Citizen r a | r -> a where
+  -- | Generate two @a@s via two @r@s. Additionally, produce a list of strings
+  -- corresponding to any arbitrary arguments we needed to build.
   getCitizen :: r -> r -> Gen ([String], (a, a))
-
 
 instance {-# OVERLAPPING #-} Citizen (Sem r a -> b) (Sem r a -> b) where
   getCitizen r1 r2 = pure ([], (r1, r2))
 
-instance {-# OVERLAPPING #-} Citizen (Sem r a) (Sem r a) where
+instance Citizen (Sem r a) (Sem r a) where
   getCitizen r1 r2 = pure ([], (r1, r2))
 
 instance (Arbitrary a, Show a, Citizen b r) => Citizen (a -> b) r where
@@ -34,7 +40,11 @@ instance (Arbitrary a, Show a, Citizen b r) => Citizen (a -> b) r where
     first (show a :) <$> getCitizen (f1 a) (f2 a)
 
 
+------------------------------------------------------------------------------
+-- | A law that effect @e@ must satisfy whenever it is in environment @r@. You
+-- can use 'runLaw' to transform these 'Law's into QuickCheck-able 'Property's.
 data Law e r where
+  -- | A pure 'Law', that doesn't require any access to 'IO'.
   Law
       :: ( Eq a
          , Show a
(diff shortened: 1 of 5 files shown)


googleson78:
Sorry if I'm missing something (including the point entirely), but can't this live in tests (to avoid depending on `QuickCheck` from the library)?

isovector:
I wish! But `Law` needs to live in the library proper (or some other library) so that any other interpretations of `State` can also prove that they're lawful.

ocharles:
`lens` deals with that by having an external library: https://hackage.haskell.org/package/lens-properties

ocharles:
I guess it's not quite the same though, because each effect has its own laws, so they would be a bit weird being in `polysemy-laws`...

isovector:
Yeah --- I can't think of a good solution here. By putting it directly in the library, at least we can set a good example. Maybe these can be hidden behind a cabal flag to disable them if you really don't want the dependency,

spacekitteh:
Alternative approach: Liquid Haskell

spacekitteh:
Also, this CLEARLY cries out for a Generator effect :P

src/Polysemy/Law.hs
```diff
@@ -0,0 +1,131 @@
...
+    ]
+
+printf :: String -> [String] -> String
+printf str args = splitArgs str
+  where
+    splitArgs :: String -> String
+    splitArgs s =
+      case break (== '%') s of
+        (as, "") -> as
+        (as, drop 1 -> (b : bs))
```
TheMatten: Just a stylistic question - do we need `ViewPatterns` here?

`drop 1 -> (b : bs)` -> `_ : b : bs`

src/Polysemy/Law.hs
```diff
@@ -0,0 +1,131 @@
...
+  , module Test.QuickCheck
+  ) where
+
+import Control.Arrow (first)
+import Data.Char
+import Polysemy
+import Test.QuickCheck
+
+
+class Citizen r a | r -> a where
```
TheMatten: We'll probably need to have some comment here, or maybe some different name - personally I had to read through the code to get the idea of what `Citizen` means :slightly_smiling_face:

src/Polysemy/Law.hs
```diff
@@ -0,0 +1,131 @@
...
+printf :: String -> [String] -> String
+printf str args = splitArgs str
+  where
+    splitArgs :: String -> String
+    splitArgs s =
+      case break (== '%') s of
+        (as, "") -> as
+        (as, drop 1 -> (b : bs))
+          | isDigit b
+          , let d = read [b] - 1
```
TheMatten: Just a nitpick - this code seems to work with one-digit indexes only - is this a problem?

src/Polysemy/State/Law.hs
```diff
@@ -0,0 +1,60 @@
...
+
+
+------------------------------------------------------------------------------
+-- | A collection of laws that show a `State` interpreter is correct.
+prop_lawfulState
+    :: forall r s
+     . (Eq s, Show s, Arbitrary s, MakeLaw (State s) r)
+    => InterpreterFor (State s) r
+    -> Property
+prop_lawfulState i12n = conjoin
```
TheMatten: Nitpick - not sure I understand name `i12n` :smile:

src/Polysemy/Law.hs
isovector: You're right, it does. I think if your law requires more than 9 moving parts, it's probably too complicated :)

src/Polysemy/State/Law.hs
isovector: **i** nterpretation **n**, but yeah, fair enough

src/Polysemy/Law.hs
isovector: spoilsport ;)

KingoftheHomeless:
hey! I just started working on a review.

isovector:
sorry! i get too excited. i'll respond to your comments  when you're done

src/Polysemy/Law.hs
```diff
@@ -0,0 +1,197 @@
...
+      -> String
+      -> res
+      -> Law e r
+
+instance MakeLaw e '[] where
+  mkLaw = Law run
+
+instance MakeLaw e '[Embed IO] where
+  mkLaw = LawIO runM
+
```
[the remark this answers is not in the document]
isovector: sure! i'm just adding crap here as i need it.

isovector:
@KingoftheHomeless good questions. My current plan is to make it as widely-useful batteries-included as possible. Having more packages just makes my job harder, and doesn't really buy anyone anything on this front.

isovector:
That being said, this isn't release yet, so I could still be talked into moving it if yall are convinced it doesn't belong here.

PR #270, opened 2019-11-01
Add fromException and fromExceptionVia

isovector:
Been thinking about how to write good real life code with polysemy --- realized this thing was conspicuously missing.

### DIFF
--- a/polysemy.cabal
+++ b/polysemy.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.24
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 5fb909edb732407e798db0fa3e2e0a62dbeb807e960b77564b99b34a3235fdb3
+-- hash: 8a6be6da10631778dfdd107f8abaf266ea185e5b22d4a9ba78283ec5eff648b2
 
 name:           polysemy
 version:        1.2.3.0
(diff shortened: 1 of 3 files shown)


KingoftheHomeless:
Should we provide a `Final`-generalized variant too?

```haskell
fromExceptionSemVia
    :: ( X.Exception exc
       , Member (Error err) r
       , Member (Final IO) r
       )
    => (exc -> err)
    -> Sem r a
    -> Sem r a
fromExceptionFinalVia f m = do
  r <- withStrategicToFinal $ do
    m' <- runS m
    s  <- getInitialStateS
    pure $ (fmap . fmap) Right m' `X.catch` \e -> pure (Left e <$ s)
  case r of
    Left e -> throw $ f e
    Right a -> pure a
```

isovector:
I think if you're throwing exceptions in polysemy you deserve to get burned! Would `fromExceptionSemVia` buy us anything I'm not seeing?

KingoftheHomeless:
@incertia [had a use-case where it would've been useful](https://github.com/polysemy-research/polysemy/issues/243#issue-503009538). Although from what I can tell, it was mostly for convenience; instead of doing `fromExceptionVia` on every IO action which may throw an exception, you instead apply `fromExceptionSemVia` on a big batch of actions converting any IO exception that may occur.

isovector:
That sounds reasonable to me. I'd be happy to merge a PR.

PR #272, opened 2019-11-01
WIP: History effect

isovector:
Get undo/redo over a state effect for freeeee

### DIFF
--- a/src/Polysemy/History.hs
+++ b/src/Polysemy/History.hs
@@ -45,7 +45,7 @@ runHistory m = do
           put s'
         PutAndForget s' ->
           put s'
-                )
+                             )
     . intercept @(State s) ( \case
         Get -> get
         Put s' -> do
--- a/src/Polysemy/History.hs
+++ b/src/Polysemy/History.hs
@@ -1,0 +1,74 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE TemplateHaskell     #-}
+
+module Polysemy.History
+  ( -- * Effect
+    History (..)
+
+    -- * Actions
+  , undo
+  , redo
+  , putAndForget
+
+    -- * Interpretations
+  , runHistory
+  ) where
+
+import Polysemy
+import Polysemy.State
+import Polysemy.Tagged
+
+
+data History s m a where
+  Undo :: History s m ()
+  Redo :: History s m ()
+  PutAndForget :: s -> History s m ()
+
+makeSem ''History
+
+
+runHistory
+    :: forall s r a
+     . Member (State s) r
+    => Sem (History s ': r) a
+    -> Sem r a
+runHistory m = do
+  s <- get
+  evalState (Zipper [] s [])
+    . untag @"history" @(State (Zipper s))
+    . interpret @(History s) (\case
+        Undo -> do
+          s' <- tagged @"history" $ modify (goBack @s) >> gets (focusZ @s)
+          put s'
+        Redo -> do
+          s' <- tagged @"history" $ modify (goForward @s) >> gets (focusZ @s)
+          put s'
+        PutAndForget s' ->
+          put s'
+                )
+    . intercept @(State s) ( \case
+        Get -> get
+        Put s' -> do
+          raise $ tagged @"history" $ modify $ modifyZ $ const s'
+          put s'
+                           )
+    . raiseUnder
+    $ m
+
+
+data Zipper a = Zipper [a] a [a]
+
+focusZ :: Zipper a -> a
+focusZ (Zipper _ a _) = a
+
+modifyZ :: (a -> a) -> Zipper a -> Zipper a
+modifyZ f (Zipper past a _) = Zipper (a : past) (f a) []
+
+goBack :: Zipper a -> Zipper a
+goBack z@(Zipper [] _ _) = z
+goBack (Zipper (p : ps) a fs) = Zipper ps p (a : fs)
+
+goForward :: Zipper a -> Zipper a
+goForward z@(Zipper _ _ []) = z
+goForward (Zipper ps a (f : fs)) = Zipper (a : ps) f fs
+
--- a/src/Polysemy/History/Law.hs
+++ b/src/Polysemy/History/Law.hs
@@ -7,7 +7,7 @@ import Polysemy.History
 
 
 ------------------------------------------------------------------------------
--- | A collection of laws that show a `State` interpreter is correct.
+-- | A collection of laws that show a 'History' interpreter is correct.
 prop_lawfulHistory
     :: (Eq s, Show s, Arbitrary s)
     => InterpreterFor (History s) '[State s]
@@ -62,7 +62,7 @@ law_forgotten
     => Law (History s) '[State s]
 law_forgotten =
   Law (\s -> run . evalState @s s)
-    "put %1 >> undo >> redo >> get"
+    "put %1 >> putAndForget %2 >> undo >> redo >> get"
     (\s s' -> put @s s >> putAndForget @s s' >> undo @s >> redo @s >> get @s)
     "put %1 >> get"
     (\s _ -> put @s s >> get @s)
--- a/src/Polysemy/History/Law.hs
+++ b/src/Polysemy/History/Law.hs
@@ -1,0 +1,56 @@
+module Polysemy.History.Law where
+
+import Polysemy
+import Polysemy.Law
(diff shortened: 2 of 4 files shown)


src/Polysemy/History.hs
```diff
@@ -0,0 +1,74 @@
...
+
+makeSem ''History
+
+
+runHistory
+    :: forall s r a
+     . Member (State s) r
+    => Sem (History s ': r) a
+    -> Sem r a
+runHistory m = do
```
KingoftheHomeless: Is `historyToIO` possible? Feels like it should be.

src/Polysemy/History/Law.hs
```diff
@@ -0,0 +1,69 @@
+module Polysemy.History.Law where
+
+import Polysemy
+import Polysemy.Law
+import Polysemy.State
+import Polysemy.History
+
+
+------------------------------------------------------------------------------
+-- | A collection of laws that show a `State` interpreter is correct.
```
KingoftheHomeless:
```
-- | A collection of laws that show a `History` interpreter is correct.
```

test/LawsSpec.hs
```diff
@@ -18,3 +20,7 @@ spec = parallel $ do
     it "stateToIO should pass the laws" $
       property $ prop_lawfulState @'[Embed IO] $ fmap snd . stateToIO @Int 0
 
+  describe "History effects" $ do
+    it "runHistory should pass the laws" $
+      property $ prop_lawfulHistory $ runHistory @Int
+
```
KingoftheHomeless: If we're going to add more laws for testing purposes, then I feel `LawsSpec` should be removed, and all laws should be placed in their respective effect's test spec.

src/Polysemy/History/Law.hs
```diff
@@ -0,0 +1,69 @@
...
+    :: forall s
+     . (Eq s, Arbitrary s, Show s)
+    => Law (History s) '[State s]
+law_forgotten =
+  Law (\s -> run . evalState @s s)
+    "put %1 >> undo >> redo >> get"
+    (\s s' -> put @s s >> putAndForget @s s' >> undo @s >> redo @s >> get @s)
+    "put %1 >> get"
+    (\s _ -> put @s s >> get @s)
+
```
KingoftheHomeless: `put a >> undo >> put b >> redo >> get == put b >> get`, right? If so, that law also belongs here.

test/LawsSpec.hs
isovector: yeah, that's reasonable. the whole directory could use some cleaning love

src/Polysemy/History/Law.hs
isovector: I think that is just the composition of `undoes` and `redoNop`

src/Polysemy/History.hs
isovector: with what semantics?

KingoftheHomeless: Global-to-everything semantics; i.e. `State (Zipper s)` is interpreted via `stateToIO`. It's the only way we get consistency if the targeted `State s` is interpreted via `stateToIO` (since `get`/`put` of `State (Zipper s)` can get reverted by IO exceptions, but not of the `State s`).

Then again, `historyToIO` then becomes inconsistent if `State s` is interpreted via `runState`. Maybe it'd be too messy.

src/Polysemy/History.hs
```diff
@@ -0,0 +1,74 @@
...
+  s <- get
+  evalState (Zipper [] s [])
```
KingoftheHomeless: If an error effect comes between `History` and `State s`, then it's possible to throw exceptions that will revert the record-keeping of `History`, but not affect the targeted `State s`, sabotaging consistency.

src/Polysemy/History/Law.hs
KingoftheHomeless: You're right! Missed `redoNop`.

src/Polysemy/History.hs
isovector: Maybe there's a more general solution here which is to be able to steal an interpretation?

KingoftheHomeless: I don't think it's possible to interpret `History` solely using `State s`.

However, I've previously entertained the notion of a "StateManager" effect, like this:

```haskell
data StateManager k m a where
  StateCarrier :: k a -> StateManager k m a
  ManageState :: Member (StateManager k) r => s -> Sem (State s ': r) a -> StateManager k m (s, a)

makeSem ''StateManager

runStateManager :: Sem (StateManager (Const Void) ': r) a -> Sem r a
runStateManager = interpretH $ \case
  StateCarrier (Cont m) -> absurd m
  ManageState s m -> runT (runState s m) >>= raise . runStateManager

stateManagerToIO :: Member (Embed IO) r => Sem (StateManager IO ': r) a -> Sem r a
stateManagerToIO = interpretH $ \case
  StateCarrier m -> embed m
  ManageState s m -> do
    ref <- embed $ newIORef s
    m'<- runT $ ($ m) $ interpret $ \case
      Put s' -> stateCarrier $ writeIORef ref s'
      Get -> stateCarrier $ readIORef ref
    res <- raise $ stateManagerToIO m'
    end <- embed $ readIORef ref
    return $ fmap (end,) res
```

It's just that I've been looking for a formulation of it that isn't, y'know, disgusting.

src/Polysemy/History.hs
isovector: haddock: "Make sure to run this immediately before running your state!!!"?

KingoftheHomeless: If we want that restriction, we could do:
```haskell
runHistory :: s -> Sem (History s ': State s ': r) a -> Sem r (s, a)
runHistory s = runState s . ...
```

isovector:
I consider the zoo to be experimental and gross and should be avoided if possible. This effect, on the other hand, is a drop-in replacement for most applications' entire undo/redo management. 

Possibly a `polysemy-extra` package would be a better place for it, but I cbf to do all the maintenance to make that happen, and this doesn't acquire any new dependencies.

I'm not tied to landing it here, but I think `polysemy` would be much easier to understand if it had lots of examples of useful-out-of-the-box effects, rather than just "low level things you should never use in real applications."

isovector:
Although a downside: the implementations here are more convoluted than they need to be due to not having plugin support in the core lib.

src/Polysemy/History.hs
KingoftheHomeless: The plus side to that is that `historyToIO` would lose its problems too.
```haskell
historyToIO :: s -> Sem (History s ': State s ': r) a -> Sem r (s, a)
historyToIO s = stateToIO s . ...
```

isovector: but then you can't mix and match your state :(

KingoftheHomeless: Oh yeah. Damn.

bolt12: > haddock: "Make sure to run this immediately before running your state!!!"?
How is this different than forcing `History` to come before `State` in the stack?

Maybe I didn't understand what you mean by "Mix and match the state"

KingoftheHomeless: One example is that it's possible to interpret a `State s` in terms of `State t` if you have a `Lens' t s` (although it's not actually offered in the library):

```haskell
mapState :: Member (State t) r => Lens' t s -> Sem (State s ': r) a -> Sem r a
mapState lens = interpret $ \case
  Put s -> do
    t <- get
    put $ set lens s t
  Get -> view lens <$> get
```
Hence, you can mix and match smaller pieces of state with larger pieces of state. These kinds of transformations are impossible if we force one part of the state to be interpreted immediately as part of interpreting `History`. 
Admittedly, perhaps @isovector meant something different.

isovector: more generally, i meant "there are lots of state interpretations, why should we privilege `runState` and `stateToIO`"?

bolt12: Oh okay, so basically there are other useful cases where you might want to add an interpretation between `History` and `State`, right?

isovector: my design principle is "people in the future are going to be smarter than i am, so don't make arbitrary restrictions in what they can do just because i don't see it"

bolt12: I see, that's actually a pretty good design principle!

TheMatten:
Do we want to merge this? Personally I like the idea of `polysemy-extra`.

Issue #246, opened 2019-10-08
Choice of functorial state for runNonDet


KingoftheHomeless:
Turns out choosing a functorial state for `runNonDet` is rife with trouble.

We previously used to use `[]` as the functorial state for `runNonDet`. However, this made the semantics of `<|>` differ depending on if it's used inside or outside a higher-order action of an effect interpreted after it.

```haskell
data HO m a where
  HO :: m a -> HO m a

makeSem HO

runHO :: Sem (HO ': r) a -> Sem r a
runHO = interpretH $ \(Ho m) -> runT m >>= raise . runHO

-- prints "abrahadabra" with both the new and old choice of functorial state
abrahadabra :: IO [()]
abrahadabra =
    runM
  . runHO
  . runNonDet @[]
  $ do
    _ <- putStr "ab" <|> putStr "hadab"
    putStr "ra"

-- prints "abhadabrara" with the old choice of functorial state.
--- prints "abrahadabra" with the new choice of functorial state.
abhadabrara :: IO [()]
abhadabrara =
    runM
  . runHO
  . runNonDet @[]
  $ do
    _ <- hO $ putStr "ab" <|> putStr "ha"
    putStr "ra"
```

This also means that `listen` and `censor` will hear all `tell`s of all branches collectively within the argument to a `listen`/`censor`; for example, if you do `listen $ tell "ab" <|> tell "ra"`, then you will hear `"abra"` in both resulting branches, rather than first `"ab"`, and then `"ra"`.

My solution in #210 was to change the choice of functorial state to one similar to one `ListT` uses for its `MonadBaseControl` instance in the `list-t` package. That fixes this, but it has a unfortunate side-effect; **only the first branch is visible to any higher-order action**, since all other branches are hidden in the functorial state.

For example:
```haskell
-- Prints out "onetwo" under the old choice of functorial state.
-- Prints out "one" under the new choice of functorial state.
asyncTest :: IO [()]
asyncTest =
    runFinal
  . asyncToIOFinal
  . runNonDet
  $ do
    a <- async (putStr "one" <|> putStr "two")
    _ <- await a
    return ()
```
Similarly, `censor` only affects the first branch, `listen` only hears what the first branch has to `tell`, and most egregiously, **`catch` can only catch exceptions of the first branch**.

So the two different choices of functorial state of `runNonDet` have their own share of troubles, and neither choice is ideal. I doubt there's a perfect solution to this problem.

[...]

PR #273, opened 2019-11-01
Revert functorial state of runNonDet to []

KingoftheHomeless:
See #246 

all around me are familiar faces

### DIFF
--- a/test/AlternativeSpec.hs
+++ b/test/AlternativeSpec.hs
@@ -51,5 +51,7 @@ spec = parallel $ do
        \ a higher-order action of a later effect (See Issue #246.)" $ do
       (fst . run . runTraceList . runReader () . runNonDet @[]) inHigherOrder
         `shouldBe` ["1","2","3","3"]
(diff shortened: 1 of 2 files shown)


isovector:
i find this hard to take :(

test/AlternativeSpec.hs
```diff
@@ -38,3 +45,11 @@ spec = parallel $ do
         `shouldBe` Just (["salabim"], ())
       (run . runTraceList . runNonDetMaybe) failtrace'
         `shouldBe` (["sim", "salabim"], Just ())
+
+  describe "runNonDet" $ do
+    it "should have terrible semantics when <|> is used within\
```
isovector: excellent test name

isovector: could we also write an "expect failure" for what it should be, just to document that?

KingoftheHomeless: Good idea. Simply via `shouldNotBe`, or is there something better?

isovector: There's nothing better :( but god would it be nice if there were.

PR #278, opened 2019-11-11
Race inputs

### LINKED ISSUE #277: Interpreters don't compose nicely
I am trying to race two `Input` interpreters; the result is a nightmare and took me an hour to work out.

```haskell

raceInput
    :: Member (Final IO) r
    => InterpreterFor (Input i) r
    -> InterpreterFor (Input i) r
    -> InterpreterFor (Input i) r
raceInput i1 i2 m = withWeavingToFinal $ \st lower ins -> do
  let i1' = lower $ (<$ st) $ i1 input
  let i2' = lower $ (<$ st) $ i2 input
  runSem (fmap (<$ st) m) $ \u ->
    case decomp u of
      Left x -> do
        z <- lower $ (<$ st) $ liftSem $ hoist (raceInput i1 i2) x
        pure $ fromJust $ ins z
      Right (Weaving Input st' _ y _) -> do
        z <- race i1' i2'
        pure $ y $ (<$ st') $ fromJust $ ins $ either id id $  z
```

Disgusting, and probably doesn't work very well.

isovector:
Closes #277

### DIFF
--- a/polysemy.cabal
+++ b/polysemy.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.24
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: aac2227e71e3a98458fa5746be7fcbb87da0dc342cdd53913041c95c128871fe
+-- hash: 01c430b0f9f4849729cb3f4ace0bfac8750f4009d11fe06ba8c1b898518be3a2
 
 name:           polysemy
 version:        1.2.3.0
@@ -132,6 +132,7 @@ test-suite polysemy-test
       FixpointSpec
       FusionSpec
       HigherOrderSpec
+      InputSpec
       InspectorSpec
       InterceptSpec
       LawsSpec
--- a/src/Polysemy/Input.hs
+++ b/src/Polysemy/Input.hs
@@ -14,10 +14,15 @@ module Polysemy.Input
   , raceInput
   ) where
 
+import Control.Concurrent.Async (race)
 import Data.Foldable (for_)
 import Data.List (uncons)
 import Polysemy
+import Polysemy.Internal
+import Polysemy.Final
 import Polysemy.State
+import Polysemy.Tagged
+import Data.Coerce
 
 ------------------------------------------------------------------------------
 -- | An effect which can provide input to an application. Useful for dealing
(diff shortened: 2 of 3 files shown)


isovector:
Actually I want to test this before merging it!

isovector:
You're always two steps ahead of me!

isovector:
Added a test, and cleaned up the tags from the implementation. I'm not yet sold on a `Race` effect but keen to be.

KingoftheHomeless:
I'll try to sell you on it in a few hours; busy right now.

isovector:
No rush, though I'd like to get this merged tomorrow-ish.

KingoftheHomeless:
Actually, now that I think about it, the only reasons to tie this interpreter to a `Race` effects is:

1. You can use a mock interpreter you've defined for `Race` to determine the logic of how `raceInput` works.
2. If we do decide to have a `Race` effect (for the reasons I outline in https://github.com/polysemy-research/polysemy/issues/252#issuecomment-552452397) it'd look weird to not have this interpreter be in terms of `Race`.

Of course, 1 is more compelling, since you can imagine that mock interpreters for `Race` could have pretty advanced logic (e.g. the branch that wins alternates between invocations to `race`). You could write your own variant `raceInput` that reimplements that logic, but that would be duplicated code of _advanced_ code.

Nonetheless, none of these 2 reasons are deal-breakers. I'm ok with having `raceInput` be defined in terms of `Final IO`.

isovector:
I've been playing with this quite a lot, and don't like the direction it's heading in. For example, here are some combinators I've written:

```haskell
rewriteInput
    :: (i2 -> i1)
    -> Sem (Input i1 ': r) a
    -> Sem (Input i2 ': r) a
rewriteInput f = reinterpret $ \Input -> do
  i <- input
  pure $ f i


filterInput
    :: Sem (Input i ': r) a
    -> Sem (Input (Maybe i) ': r) a
filterInput = reinterpret $ \Input ->
  fix $ \me -> do
    input >>= \case
      Just i -> pure i
      Nothing -> me
```

which get used via:

```haskell
tuiInput
    :: forall r a
     . ( Member (Final IO) r
       , Member (Embed IO) r
       )
    => Sem (Input Command ': r) a
    -> Sem r a
tuiInput =
  raceInput
    (inputToFileWatch "."
        . rewriteInput eventToCommand
        . filterInput)
    (runInputSem $ embed readLn)
```

At this point it feels like I'm doing a bad job of reimplementing a streaming library. I think a better way forward is to give good interop with `streaming`:

```haskell
runInputViaStream
    :: S.Stream (S.Of i) (Sem r) ()
    -> InterpreterFor (Input i) r
```

Of course, you can always `lift input >>= yield :: Member (Input i) r => S.Stream (S.Of i) (Sem r) ()` to get `Input` actions _into_ the stream.

This would allow us to compose `Input` streams with a nice interface, and gets rid of all the jank.

isovector:
@masaeedu tagging you because you were expressing interest in streaming with polysemy --- looks like we can in fact do it, but _probably shouldn't_

isovector:
Implemented:

```haskell
module Sem.StreamingInput where

import           Data.Functor.Of
import           Polysemy
import           Polysemy.Input
import           Polysemy.State
import qualified Streaming as S
import qualified Streaming.Prelude as S


runInputViaStream
    :: S.Stream (Of i) (Sem r) ()
    -> InterpreterFor (Input (Maybe i)) r
runInputViaStream stream
  = evalState (Just stream)
  . reinterpret ( \Input ->
      get >>= \case
        Nothing -> pure Nothing
        Just s ->
          raise (S.inspect s) >>= \case
            Left () -> pure Nothing
            Right (i :> s') -> do
              put $ Just s'
              pure $ Just i
  )


yieldInput :: Member (Input i) r => S.Stream (Of i) (Sem r) ()
yieldInput = S.lift input >>= S.yield
```

Seems like a good candidate for the zoo.

FongHou:
Does it make sense for an Output interpreter also?

```
runOutputViaStream
  :: forall o r a. Sem (Output o ': r) a -> Sem r (S.Stream (Of o) (Sem r) (), a)
runOutputViaStream = runState (pure ())
  . reinterpret (\case Output o -> modify @(Stream (Of o) (Sem r) ()) (S.cons o))
```

However, the return type of above signature looks strange. I wish it's,

```
runOutputViaStream'
  :: forall o r a. Sem (Output o ': r) a -> S.Stream (Of o) (Sem r) a
```

But not sure it's possible to write it or not.

Issue #286, opened 2019-11-26
outputToTrace has a bad type


isovector:
`outputToTrace :: (Show w, Member Trace r) => Sem (Output w ': r) a -> Sem r a `

but should be

`outputToTrace :: Member Trace r => (w -> String) => Sem (Output w ': r) a -> Sem r a `

juanpaucar:
Hi @isovector, I'd love to take this ticket. But,  I'm just getting familiar with the library so I might have a PR for the weekend. Would that be okay with you? If so, could I get this issue assigned to me?

isovector:
Hi @juanpaucar --- feel free to tackle this, thanks! No rush on it :)

juanpaucar:
@isovector I just wanted to confirm that the desired function signature is

```hs
outputToTrace :: Member Trace r => (w -> String) -> Sem (Output w ': r) a -> Sem r a

```

Replacing the second `=>` with a `->` on the signature you posted above.

Disclaimer: polisemy is so interesting and I've seen so many new things by reading the code that I don't know what is valid Haskell anymore :sweat_smile:

KingoftheHomeless:
Addressed by #289.

isovector:
Thanks for this, @juanpaucar !

PR #289, opened 2019-12-02
Update outputToTrace definition

juanpaucar:
Solves #286

### DIFF
--- a/src/Polysemy/Trace.hs
+++ b/src/Polysemy/Trace.hs
@@ -77,16 +77,16 @@ runTraceList = runOutputList . reinterpret (
 
 
 ------------------------------------------------------------------------------
--- | Transform an 'Output' 'String' effect into a 'Trace' effect.
+-- | Transform an @'Output' w@ effect into a 'Trace' effect given a function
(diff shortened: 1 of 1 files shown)


KingoftheHomeless:
Could you update the documentation too? E.g. `Transforms an @'Output' w@ effect into a 'Trace' effect given a function to transform each @w@ to a 'String'.`

juanpaucar:
@KingoftheHomeless Sorry about that, I totally forgot about updating the documentation. I just pushed the updated documentation too

KingoftheHomeless:
Thanks!

isovector:
I think this should have an explicit `forall w r a.` on it, so that you can type apply the effect; for example, today you'd need to write `outputToTrace @_ @MyEffect show`, when instead we could make it `outputToTrace @MyEffect show`

juanpaucar:
Hi @isovector, Would this change be enough? https://github.com/juanpaucar/polysemy/commit/68307b671139bcf74fa74ef51fddf4fb4a3b1605

I  think I will need to open another PR for that if you want it

isovector:
Yup, that would be enough!

PR #291, opened 2019-12-03
Add explicit quantifier for parameters on outputToTrace

juanpaucar:
Adding an explicit quantifier as requested on this comment https://github.com/polysemy-research/polysemy/pull/289#issuecomment-560953740 by @isovector

### DIFF
--- a/src/Polysemy/Trace.hs
+++ b/src/Polysemy/Trace.hs
@@ -82,7 +82,8 @@ runTraceList = runOutputList . reinterpret (
 --
 -- @since 1.0.0.0
 outputToTrace
-  :: Member Trace r
+  :: forall w r a
+   . Member Trace r
   => (w -> String)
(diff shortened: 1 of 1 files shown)


src/Polysemy/Trace.hs
```diff
@@ -82,7 +82,7 @@ runTraceList = runOutputList . reinterpret (
 --
 -- @since 1.0.0.0
 outputToTrace
-  :: Member Trace r
+  :: forall w r a. Member Trace r
```
isovector:
```suggestion
  :: forall w r a
   . Member Trace r
```

isovector: Happy to merge after this. Thanks again, @juanpaucar

juanpaucar: Done :)

Issue #294, opened 2019-12-10
Support `ST` in `State`


incertia:
Since we can run `State` in `IO` with `IORef`s it makes sense that we should also be able to run `State` in `ST` with `STRef`s unless I am mistaken somehow?

incertia:
I suspect that this may be difficult to do, as the type of `runST` is
```
runST :: (forall s. ST s a) -> a
```
and we are only able to stick
```
Embed (ST s)
```
into the type level list

KingoftheHomeless:
It's possible. Just use `runST $ runM . ... $ m`

KingoftheHomeless:
This feels like a good first issue, actually:

**We want the following interpretations in `Polysemy.State`:**
```haskell
runStateSTRef
  :: forall s st r a
   . Embed (ST st) r
  => STRef st s
  ->Sem (State s ': r) a
  -> Sem r a
```
and
```haskell
stateToST
  :: forall s st r a
   . Embed (ST st) r
  => s
  -> Sem (State s ': r) a
  -> Sem r (s, a)
```

These should be easy to implement by using `runStateIORef` and `stateToIO` as references, adapting them to instead use `ST`/`Ref`. Don't forget to adapt the documentation too! (You can skip the warnings about it not being atomic).

incertia:
Perhaps I should've elaborated a bit more. I did try the implementation you suggested but ended up with this:

```
> m = pure () :: Sem '[State Int, Embed (ST st)] ()
> :t runST $ runM $ stateToST 0 $ m

<interactive>:1:16: error:
    • Couldn't match type ‘Polysemy.Internal.Union.IndexOf
                             '[Embed (ST s)]
                             (Polysemy.Internal.Union.Found '[Embed (ST s)] (Embed (ST st0)))’
                     with ‘Embed (ST st0)’
        arising from a use of ‘stateToST’
      The type variable ‘st0’ is ambiguous
    • In the expression: stateToST 0
      In the second argument of ‘($)’, namely ‘stateToST 0 $ m’
      In the second argument of ‘($)’, namely ‘runM $ stateToST 0 $ m’
> :t m & stateToST 0 & runM & runST

<interactive>:1:26: error:
    • Couldn't match type ‘ST st0 (Int, ())’ with ‘forall s. ST s b’
      Expected type: ST st0 (Int, ()) -> b
        Actual type: (forall s. ST s b) -> b
    • In the second argument of ‘(&)’, namely ‘runST’
[...]

PR #295, opened 2019-12-11
Add runStateSTRef and stateToST

incertia:
This addresses #294. One remaining question: is reading and then writing `IORef`/`STRef`s any faster than calling `modifyIORef`/`modifySTRef` or their strict counterparts? If it offers performance speedups we may want to directly interpret `modify`.

### DIFF
--- a/src/Polysemy/State.hs
+++ b/src/Polysemy/State.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE TemplateHaskell #-}
 
 module Polysemy.State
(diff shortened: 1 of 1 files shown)


KingoftheHomeless:
The source code for `modifyIORef`/`modifySTRef` _does_ simply read the ref and then write to it. It uses no special operations, like `atomicModifyIORef`. Assuming the code gets optimized properly, there should be no difference to using Polysemy's `modify` vs. `modifyIORef`.

src/Polysemy/State.hs
```diff
@@ -157,6 +162,59 @@ stateToIO s sem = do
...
+-- /Beware/: As this uses an 'STRef' internally,
+-- all other effects will have local
+-- state semantics in regards to 'State' effects
+-- interpreted this way.
+-- For example, 'Polysemy.Error.throw' and 'Polysemy.Error.catch' will
+-- never revert 'put's, even if 'Polysemy.Error.runError' is used
+-- after 'stateToST'.
+--
+-- Additionally, one has to introduce the @st@ type to 'stateToST' by binding
+-- the type after 'runM' with an explicit forall with @-XScopedTypeVariables@.
```
KingoftheHomeless: Preface this by saying that this isn't necessary if you're using the plugin, like:
`Additionally, if you're not using `polysemy-plugin`, you have to...`

incertia: done

KingoftheHomeless:
Thanks!

PR #299, opened 2019-12-14
Remove extra-deps that are now in resolver

googleson78:
Thanks to #298 there are now some new things in our resolver, so we can remove them from `extra-deps`

Notably this bumps the following libraries:
* aeson: 1.4.3.0 -> 1.4.6.0
* bifunctors: 5.5.4 -> 5.5.6
* inspection-testing: 0.4.2 -> 0.4.2.2
* th-abstraction: 0.3.1.0 -> 0.3.1.0@rev:1

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -5,15 +5,7 @@ packages:
 - polysemy-plugin
 
 extra-deps:
-- aeson-1.4.3.0
-- bifunctors-5.5.4
 - dump-core-0.1.3.2
 - first-class-families-0.6.0.0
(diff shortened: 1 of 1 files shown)


googleson78:
I also deleted `monadLib` as it seems to be unused!

isovector:
iirc, `monadLib` is used for the core-dump stuff

googleson78:
Are you referring to this - [`dump-core`](https://hackage.haskell.org/package/dump-core)?

PR #303, opened 2019-12-19
Add execState to Polysemy.State.

tjweir:


src/Polysemy/State.hs
```diff
@@ -15,6 +15,7 @@ module Polysemy.State
     -- * Interpretations
   , runState
   , evalState
+  , execState
   , runLazyState
```
KingoftheHomeless: Please also provide an `execLazyState` for symmetry.

tjweir: Will do.

KingoftheHomeless:
Thanks!

PR #308, opened 2020-01-18
Add zulip badge

srid:
[Rendered](https://github.com/srid/polysemy/blob/patch-1/README.md#polysemy)

Funprog zulip has a dedicated stream for Polysemy where a number of discussions are taking place. Sandy is participating as well. So why not make it the official chat for the project?

https://github.com/srid/rib also uses it.


KingoftheHomeless:
I don't have any problem with it, but let's see everyone agrees.

PR #311, opened 2020-01-30
Add lazy interpretations for Output and Writer

### LINKED ISSUE #310: Output to lazy list
Take the program `mapM_ output [1..]`. It will of course simply output an infinite list of integers, but accessing this list has proven difficult. Running it with `runOutputList` doesn't work, since that uses `reverse` internally. `runOutputSem (embed . print)` works for printing it, but that's not generalizable. Is there any way to do this?

KingoftheHomeless:
Closes #310.

How would you go about creating tests for this? Something like "`take 100 . fst . run . runLazyOutputList $ forever (output ())` must terminate within 2 seconds?"

### DIFF
--- a/src/Polysemy/Internal/Writer.hs
+++ b/src/Polysemy/Internal/Writer.hs
@@ -5,11 +5,16 @@ module Polysemy.Internal.Writer where
 import Control.Concurrent.STM
 import Control.Exception
 import Control.Monad
+import qualified Control.Monad.Trans.Writer.Lazy as Lazy
 
+import Data.Bifunctor (first)
 import Data.Semigroup
 
 import Polysemy
 import Polysemy.Final
+
+import Polysemy.Internal
+import Polysemy.Internal.Union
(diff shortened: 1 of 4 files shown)


googleson78:
Maybe
```haskell
test = do
  output 3
  output 3
  output 3
  undefined
<run test and see if you can actually take the first 3 members only>
```
Although to be fair I have no idea whether the way `polysemy` internals work will allow for something like this!

KingoftheHomeless:
It does, as long as no strict interpreters or strict monad are used _after_ `runLazyOutputMonoid`/`runLazyWriter`. `listen` and `censor` are weird though; in order for these to work lazily, you can't use any strict interpreters at all.
I've documented this through the tests.

KingoftheHomeless:
About the requested review: I just want you to take a look at the tests and see that they're understandable, in the sense of communicating when `runLazyWriter` is strict or not.

PR #321, opened 2020-02-16
Added ability to cancel running Async actions

aidangilmore:
I thought this may come in handy.


sorki:
I was looking for this the other day, thanks!

googleson78:
Is there anything dangerous about this or other reasons we shouldn't merge it? @KingoftheHomeless

KingoftheHomeless:
Huh. Yeah, I saw `withAsync` and was dubious towards its safety, and thought to look into it later when I had time. Now I see the author figured out that it was unsafe, and removed it. This is good to merge.

Issue #304, opened 2020-01-02
Restrict the existentially quantified monad in a `Weaving` to be `Sem r`


KingoftheHomeless:
From the documentation for `Weaving` in `Polysemy.Internal.Union`:
```haskell
data Weaving e m a where
  Weaving
    :: Functor f
    =>   { weaveEffect :: e m a
         -- ^ The original effect GADT originally lifted via
         -- 'Polysemy.Internal.send'. There is an invariant that @m ~ Sem r0@,
         -- where @r0@ is the effect row that was in scope when this 'Weaving'
         -- was originally created.
             ...
         }
    -> Weaving e n b
```
Turns out that knowing that `m ~ Sem r0` can actually be relevant, since that means `m` is a monad, and that allows you to manipulate actions that you transform using the distribution `weaveDistrib :: forall x. f (m x) -> n (f x)`.

For a toy example, it would allow interpreting effects like these (which is not possible today):

```haskell
data Zip m a where
  Zip :: m a -> m b -> Zip m (a, b)
```

Through the following:
```haskell
runZip :: Sem (Zip ': r) a -> Sem r a
runZip = interpretH $ \case
  Zip ma mb -> do
    mab <- runT (liftA2 (,) ma mb) -- Requires knowing that existential monad is truly a monad
    (raise . runZip) mab
```

For a more convincing example, this would allow us to write the following interpreter:
```haskell
newtype Restoration m = Restoration (forall x. m x -> m x)

maskToFinal
  :: (Member (Final m) r, MonadMask m)
  => Sem (Mask (Restoration m) ': r) a
  -> Sem r a
```

For the following effect, which completely encapsulates the power of the `MonadMask` type-class:
```haskell
data Mask s m a where
  Mask' :: (s -> m a) -> Mask s m a
  UninterruptibleMask' :: (s -> m a) -> Mask s m a
  Restore' :: s -> m a -> Mask s m a
  GeneralBracket ::
       m a
    -> (a -> ExitCase b -> m c)
    -> (a -> m b)
    -> Mask s m (b, c)
```

To be more specific, this change allows for the interpretation of `GeneralBracket` in a similar way as in https://github.com/polysemy-research/polysemy-zoo/issues/61.

We want to change `Weaving` as follows:
```haskell
data Weaving e m a where
  Weaving
    :: Functor f
    =>   { weaveEffect :: e (Sem r0) a
         -- ^ The original effect GADT originally lifted via
[...]

PR #328, opened 2020-03-19
Make interpretH and friends expose Monad on existential m

A1kmm:
The extra strength this provides makes it far easier to write certain types of higher order interpreters (for example, to use bindT with functions of more than one argument).


KingoftheHomeless:
The related issue to this is #304 . I'd approve this, but I'd like to change the definition of `Weaving` so instead of carrying around a monad dictionary, the existential `m` is `Sem r`. #304 details how you'd go about this, I just haven't gotten around to doing it.

A1kmm:
Superceded by #333 which takes the approach suggested above.

Issue #314, opened 2020-02-03
Intercept and re-send


Caagr98:
I have a project in which I want to intercept one particular case, letting the others bubble up the stack. I currently do that with ```
intercept \case
  Case1 a -> …
  Case2 a -> case2 a
  Case3 a -> case3 a
```
Most of those cases are completely trivial, but trying to generalize to `a -> send a` gives a type error — I guess the `m` parameter is different. Is there any way to deal with this or do I have to keep enumerating all the cases explicitly?

**Comment by KingoftheHomeless** [comment]
Missed this issue, sorry for the late reply.
Your guess is on the mark. `send :: Member e r => e (Sem r) a -> Sem r a` , but intercept gives you an `e m x`, where `m` is existentially quantified, so you can't `send` it.
_However_, since you're using `intercept`, `e` is first-order, and thus `m` is a phantom parameter, so you can fix this by using `coerce`.

Here's an example for `State Int`:
```haskell
interceptor :: forall r a. Member (State Int) r => Sem r a -> Sem r a
interceptor = intercept @(State Int) $ \case
  Get -> (+1) <$> get
  a   -> send @(State Int) (coerce a)
```

This is a decent solution to this issue, but it gives rise to some further questions for us to ponder:
1. This won't work if you use `interceptH`. It's possible to implement a solution for `interceptH`, but it would probably require changing `Tactics` a bit.
2. `send` is internal. Seeing how useful it is in this case, I'm convinced that it shouldn't be (unless we find an even prettier solution). There are also certain niche situations you either can't or don't want to use `makeSem`/`_`, but use `send` instead.

PR #333, opened 2020-04-08
Make interpretH and friends expose Monad on existential m

A1kmm:
In the process replace the existentially quantified parameter `m` of the
Weaving with a parameter `r` and use `Sem r`.

See also https://github.com/polysemy-research/polysemy/compare/master...A1kmm:weavingdetails-sem-experiment?expand=1 for this change broken down into smaller commits.

### DIFF
--- a/src/Polysemy/Bundle.hs
+++ b/src/Polysemy/Bundle.hs
@@ -58,7 +58,7 @@ runBundle
   -> Sem (Append r' r) a
 runBundle = hoistSem $ \u -> hoist runBundle $ case decomp u of
   Right (Weaving (WeavingDetails (Bundle pr e) s wv ex ins)) ->
-    Union $ UnionDetails (extendMembership @_ @r pr) $ Weaving $
+    Union (extendMembership @_ @r pr) $ Weaving $
       WeavingDetails e s wv ex ins
   Left g -> weakenList @r' @r g
 {-# INLINE runBundle #-}
@@ -72,7 +72,7 @@ subsumeBundle
   -> Sem r a
 subsumeBundle = hoistSem $ \u -> hoist subsumeBundle $ case decomp u of
   Right (Weaving (WeavingDetails (Bundle pr e) s wv ex ins)) ->
-    Union $ UnionDetails (subsumeMembership pr)
+    Union (subsumeMembership pr)
               (Weaving (WeavingDetails e s wv ex ins))
   Left g -> g
 {-# INLINE subsumeBundle #-}
--- a/src/Polysemy/Bundle.hs
+++ b/src/Polysemy/Bundle.hs
@@ -44,8 +44,8 @@ sendBundle
   -> Sem r a
 sendBundle = hoistSem $ \u -> case decomp u of
   Right (Weaving e s wv ex ins) ->
-    injWeaving $ Weaving
-      (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
+    injWeaving $
+      Weaving (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
   Left g -> hoist (sendBundle @e @r') g
 {-# INLINE sendBundle #-}
 
@@ -71,7 +71,6 @@ subsumeBundle
   -> Sem r a
 subsumeBundle = hoistSem $ \u -> hoist subsumeBundle $ case decomp u of
   Right (Weaving (Bundle pr e) s wv ex ins) ->
-    Union (subsumeMembership pr)
-              (Weaving e s wv ex ins)
+    Union (subsumeMembership pr) (Weaving e s wv ex ins)
   Left g -> g
 {-# INLINE subsumeBundle #-}
--- a/src/Polysemy/Bundle.hs
+++ b/src/Polysemy/Bundle.hs
@@ -43,9 +43,9 @@ sendBundle
   => Sem (e ': r) a
   -> Sem r a
 sendBundle = hoistSem $ \u -> case decomp u of
-  Right (Weaving e s wv ex ins) ->
-    injWeaving $
-      Weaving (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
+  Right (Weaving (WeavingDetails e s wv ex ins)) ->
+    injWeaving $ Weaving $ WeavingDetails
+      (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
   Left g -> hoist (sendBundle @e @r') g
 {-# INLINE sendBundle #-}
 
@@ -57,8 +57,9 @@ runBundle
   => Sem (Bundle r' ': r) a
   -> Sem (Append r' r) a
 runBundle = hoistSem $ \u -> hoist runBundle $ case decomp u of
-  Right (Weaving (Bundle pr e) s wv ex ins) ->
-    Union (extendMembership @_ @r pr) $ Weaving e s wv ex ins
+  Right (Weaving (WeavingDetails (Bundle pr e) s wv ex ins)) ->
+    Union $ UnionDetails (extendMembership @_ @r pr) $ Weaving $
+      WeavingDetails e s wv ex ins
   Left g -> weakenList @r' @r g
 {-# INLINE runBundle #-}
 
@@ -70,7 +71,8 @@ subsumeBundle
   => Sem (Bundle r' ': r) a
   -> Sem r a
 subsumeBundle = hoistSem $ \u -> hoist subsumeBundle $ case decomp u of
-  Right (Weaving (Bundle pr e) s wv ex ins) ->
-    Union (subsumeMembership pr) (Weaving e s wv ex ins)
+  Right (Weaving (WeavingDetails (Bundle pr e) s wv ex ins)) ->
+    Union $ UnionDetails (subsumeMembership pr)
+              (Weaving (WeavingDetails e s wv ex ins))
   Left g -> g
 {-# INLINE subsumeBundle #-}
--- a/src/Polysemy/Bundle.hs
+++ b/src/Polysemy/Bundle.hs
@@ -43,8 +43,8 @@ sendBundle
   => Sem (e ': r) a
   -> Sem r a
 sendBundle = hoistSem $ \u -> case decomp u of
-  Right (Weaving (WeavingDetails e s wv ex ins)) ->
-    injWeaving $ Weaving $ WeavingDetails
+  Right (Weaving e s wv ex ins) ->
+    injWeaving $ Weaving
       (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
   Left g -> hoist (sendBundle @e @r') g
 {-# INLINE sendBundle #-}
@@ -57,9 +57,8 @@ runBundle
   => Sem (Bundle r' ': r) a
   -> Sem (Append r' r) a
 runBundle = hoistSem $ \u -> hoist runBundle $ case decomp u of
-  Right (Weaving (WeavingDetails (Bundle pr e) s wv ex ins)) ->
-    Union (extendMembership @_ @r pr) $ Weaving $
-      WeavingDetails e s wv ex ins
+  Right (Weaving (Bundle pr e) s wv ex ins) ->
+    Union (extendMembership @_ @r pr) $ Weaving e s wv ex ins
   Left g -> weakenList @r' @r g
 {-# INLINE runBundle #-}
 
@@ -71,8 +70,8 @@ subsumeBundle
   => Sem (Bundle r' ': r) a
   -> Sem r a
 subsumeBundle = hoistSem $ \u -> hoist subsumeBundle $ case decomp u of
-  Right (Weaving (WeavingDetails (Bundle pr e) s wv ex ins)) ->
+  Right (Weaving (Bundle pr e) s wv ex ins) ->
     Union (subsumeMembership pr)
-              (Weaving (WeavingDetails e s wv ex ins))
+              (Weaving e s wv ex ins)
   Left g -> g
 {-# INLINE subsumeBundle #-}
--- a/src/Polysemy/Final.hs
+++ b/src/Polysemy/Final.hs
@@ -182,7 +182,7 @@ withStrategicToFinal strat = withWeavingToFinal (runStrategy strat)
 interpretFinal
     :: forall m e r a
      . Member (Final m) r
-    => (forall x n. e n x -> Strategic m n x)
+    => (forall x rInitial. e (Sem rInitial) x -> Strategic m (Sem rInitial) x)
        -- ^ A natural transformation from the handled effect to the final monad.
     -> Sem (e ': r) a
     -> Sem r a
--- a/src/Polysemy/Final.hs
+++ b/src/Polysemy/Final.hs
@@ -190,9 +190,9 @@ interpretFinal n =
   let
     go :: Sem (e ': r) x -> Sem r x
     go = hoistSem $ \u -> case decomp u of
-      Right (Weaving (WeavingDetails e s wv ex ins)) ->
+      Right (Weaving e s wv ex ins) ->
         injWeaving $
-          Weaving $ WeavingDetails
+          Weaving
             (WithWeavingToFinal (runStrategy (n e)))
             s
             (go . wv)
@@ -214,7 +214,7 @@ interpretFinal n =
 -- @since 1.2.0.0
 runFinal :: Monad m => Sem '[Final m] a -> m a
 runFinal = usingSem $ \u -> case extract u of
-  Weaving (WeavingDetails (WithWeavingToFinal wav) s wv ex ins) ->
+  Weaving (WithWeavingToFinal wav) s wv ex ins ->
     ex <$> wav s (runFinal . wv) ins
 {-# INLINE runFinal #-}
 
@@ -233,9 +233,9 @@ finalToFinal to from =
   let
     go :: Sem (Final m1 ': r) x -> Sem r x
     go = hoistSem $ \u -> case decomp u of
-      Right (Weaving (WeavingDetails (WithWeavingToFinal wav) s wv ex ins)) ->
+      Right (Weaving (WithWeavingToFinal wav) s wv ex ins) ->
         injWeaving $
-          Weaving $ WeavingDetails
+          Weaving
             (WithWeavingToFinal $ \s' wv' ins' ->
               to $ wav s' (from . wv') ins'
             )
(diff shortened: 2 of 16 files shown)


A1kmm:
Closes #304

src/Polysemy/Internal/Union.hs
```diff
@@ -56,77 +59,79 @@ import Polysemy.Internal.CustomErrors
...
-    :: (Functor s, Functor m, Functor n)
+    :: (Functor s, Functor mMid, Functor mAfter)
     => s ()
-    -> (∀ x. s (m x) -> n (s x))
+    -> (∀ x. s (mMid x) -> mAfter (s x))
     -> (∀ x. s x -> Maybe x)
-    -> Union r m a
-    -> Union r n (s a)
-weave s' d v' (Union w (Weaving e s nt f v)) = Union w $
-    Weaving e (Compose $ s <$ s')
```
KingoftheHomeless, line 115: While I appreciate the sentiment of having more descriptive variable names, I don't think they're particularly helpful in this instance. They add clutter without providing much additional information.
In particular, I'm not sure whether or not `mMid` helps or hurts. It's true in the sense that it is the middle monad between the initial existential monad and the monad we're `weave`-ing to, but there's never any reason to care about the initial existential monad beyond the fact that it's a monad. Conceptually, `mMid` represents the monad associated to the original Union, rather than something intermediary.

My opinion is that you leave `weave` unchanged.

src/Polysemy/Internal/Union.hs
```diff
@@ -56,77 +59,79 @@ import Polysemy.Internal.CustomErrors
...
-    :: ( Functor m
-       , Functor n
-       )
-    => (∀ x. m x -> n x)
-    -> Union r m a
-    -> Union r n a
-hoist f' (Union w (Weaving e s nt f v)) = Union w $ Weaving e s (f' . nt) f v
+    :: (∀ x. mBefore x -> mAfter x)
+    -> Union r mBefore a
+    -> Union r mAfter a
```
KingoftheHomeless, line 132: Same as above: I don't find `mBefore` and `mAfter` helpful here. They don't convey any extra information.

src/Polysemy/Internal/Union.hs
```diff
@@ -256,51 +261,52 @@ tryMembership = tryMembership' @r @e
...
-weaken (Union pr a) = Union (There pr) a
+weaken (Union (UnionDetails pr a)) = Union $ UnionDetails (There pr) a
 {-# INLINE weaken #-}
 
 
 
 ------------------------------------------------------------------------------
 -- | Lift an effect @e@ into a 'Union' capable of holding it.
-inj :: forall e r m a. (Functor m , Member e r) => e m a -> Union r m a
+inj :: forall e r rWoven a. (Member e r) => e (Sem rWoven) a -> Union r (Sem rWoven) a
```
KingoftheHomeless, line 294: I like this variable naming better. I think I still prefer `r0` instead of `rWoven`, though, as `r0` better communicates that it represents the initial effect row, but I can be swayed.

src/Polysemy/Internal/Combinators.hs
```diff
@@ -160,14 +159,14 @@ lazilyStateful f = interpretInLazyStateT $ \e -> LS.StateT $ fmap swap . f e
 -- See the notes on 'Tactical' for how to use this function.
 reinterpretH
     :: forall e1 e2 r a
-     . (∀ m x. e1 m x -> Tactical e1 m (e2 ': r) x)
+     . (∀ m x. Monad m => e1 m x -> Tactical e1 m (e2 ': r) x)
```
KingoftheHomeless, line 162: This or `∀ r0 x. e1 (Sem r0) x -> Tactical e1 (Sem r0) (e2 ': r) x`? Yours certainly is prettier, in that it exposes less potentially confusing details, but I'm worried about two things:

Performance: This is the big one. I would absolutely hate it if GHC would pass `Sem r0`'s monad dictionary into this function instead of using it directly, and I believe that, unfortunately, that's exactly what it will do. Maybe it will inline and specialize upon the surface call, thanks to `loop-breaker`, but I believe that, no matter what, it will do the dictionary-passing on recursive calls. Specifying `Sem r0` is a sure way to prevent this kind of dictionary-passing from happening.

Flexibility: Even without knowing `r0`, there are some stuff you can do with an `Sem r0` by virtue of it being a free monad, like splitting it into layers of actions, which you could then muck around with. This is very contrived argument however, and I can't imagine it actually mattering. Also, exposing `m ~ Sem r0` could prove necessary to solve #314, but that's speculation.

src/Polysemy/Internal/Combinators.hs
```diff
@@ -352,7 +351,8 @@ rewrite
 rewrite f (Sem m) = Sem $ \k -> m $ \u ->
```
KingoftheHomeless, line 351: Please also change the type signatures for `rewrite` and `transform`. In contrast to `interpret`/`intercept`, this is actually important to do.

src/Polysemy/Internal/Combinators.hs
```diff
@@ -81,12 +81,11 @@ interpretH
 interpretH f (Sem m) = m $ \u ->
```
KingoftheHomeless, line 81: Please change the type signature of `interpretH` as well.

src/Polysemy/Final.hs
```diff
@@ -190,9 +190,9 @@ interpretFinal n =
   let
```
KingoftheHomeless, line 190: Please also change the type signature of `interpretFinal`.

src/Polysemy/Internal/Union.hs
```diff
@@ -56,77 +59,79 @@ import Polysemy.Internal.CustomErrors
...
-         -- ^ A function for attempting to see inside an @f@. This is no
-         -- guarantees that such a thing will succeed (for example,
-         -- 'Polysemy.Error.Error' might have 'Polysemy.Error.throw'n.)
-       }
-    -> Weaving e n b
+    :: forall f e rBefore a resultType mAfter. (Functor f)
+    => !(WeavingDetails f e rBefore a resultType mAfter)
+    -> Weaving e mAfter resultType
+
+data Functor f => WeavingDetails f e rBefore a resultType mAfter =
```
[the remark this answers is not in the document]
A1kmm: I agree that it makes sense to remove the Functor constraint from the datatype. I will add another commit to the PR soon to remove this.

I added the `...Details` to make it easier to debug at the type level. The combination of existential quantification and product types makes it hard to constrain the type inference enough to pinpoint why something is wrong. GHC doesn't have any extensions that allow, for example, type applications to constrain the existential type to a scoped type variable.

By splitting the existential part of formulation into the GADT and having a non-existentially quantified Details type, it becomes possible to name and constrain the type variables using existing syntax.

Now that I have got it to compile, it is probably possible to reverse the change to add the Details types; however, given it shouldn't have any performance impact due to it being strict, and given it will likely be helpful to other people trying to change / extend the Polysemy internals, I would suggest leaving it in to improve the developer experience and make it accessible to more people wanting to contribute.

src/Polysemy/Internal/Union.hs
A1kmm: Perhaps rInitial to distinguish it in compiler errors from any variable names automatically assigned by the compiler by appending a number?

src/Polysemy/Internal/Union.hs
A1kmm: Changed back to `m` and `n`.

src/Polysemy/Internal/Combinators.hs
A1kmm: I have fixed all instances of this to use the approach you suggested.

src/Polysemy/Internal/Union.hs
KingoftheHomeless: Good idea.

src/Polysemy/Internal/Union.hs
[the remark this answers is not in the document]
KingoftheHomeless: I've been hmm-ing and haw-ing about this for a while. I'm conflicted, but ultimately, I'm leaning towards "no". `-Details` clog up the machinery, and to gain access to existentially quantified type variables, there are easy targets both in `Weaving` and in `Union`: in `Weaving`, you can pattern match on the functorial state to gain access to the functor, and the effect to gain access to the existentially quantified row. On `Union`, you can gain access to the existentially quantified effect by pattern matching on `Weaving` or `ElemOf`. I agree, as it is now it poses a barrier to entry, but not, I think, a very big one (compared to the other stuff you need to learn for working with polysemy internals); and besides, I don't expect many newcomers to tinker with the internals.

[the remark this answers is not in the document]
A1kmm: I have added another commit to remove the `-Details` layer.

src/Polysemy/Internal/Union.hs
```diff
@@ -56,50 +57,48 @@ import Polysemy.Internal.CustomErrors
...
+      -- ^ The original effect GADT originally lifted via
+      -- 'Polysemy.Internal.send'.
+      , weaveState :: f ()
+      -- ^ A piece of state that other effects' interpreters have already
+      -- woven through this 'Weaving'. @f@ is a 'Functor', so you can always
+      -- 'fmap' into this thing.
+      , weaveDistrib :: forall x. f (Sem rBefore x) -> mAfter (f x)
+      -- ^ Distribute @f@ by transforming @Sem rBefore@ into @mAfter@. We have an
+      -- invariants on @rBefore@, which means in actuality this function looks like
+      -- @f ('Polysemy.Sem' rBefore x) -> mAfter (f x)@.
```
KingoftheHomeless, line 88: Remove:
```haskell
      --- We have an
      -- invariants on @rBefore@, which means in actuality this function looks like
      -- @f ('Polysemy.Sem' rBefore x) -> mAfter (f x)@.`
```
And replace it with:
```haskell
-- This is usually of the form @f ('Polysemy.Sem' (Some ': Effects ': r) x) -> Sem r (f x)@
```

src/Polysemy/Internal/Union.hs
```diff
@@ -56,50 +57,48 @@ import Polysemy.Internal.CustomErrors
...
+      , weaveInspect :: forall x. f x -> Maybe x
+      -- ^ A function for attempting to see inside an @f@. This is no
+      -- guarantees that such a thing will succeed (for example,
+      -- 'Polysemy.Error.Error' might have 'Polysemy.Error.throw'n.)
+      } -> Weaving e mAfter resultType
 
 instance Functor (Weaving e m) where
-  fmap f (Weaving e s d f' v) = Weaving e s d (f . f') v
+  fmap f (Weaving e s d f' v) =
+    Weaving e s d (f . f') v
```
KingoftheHomeless, line 101: Another linebreak I'm curious about.

src/Polysemy/Internal/Union.hs
```diff
@@ -111,22 +110,21 @@ weave
...
               (fmap f . getCompose)
               (v <=< v' . getCompose)
 {-# INLINE weave #-}
 
 
 hoist
-    :: ( Functor m
-       , Functor n
-       )
-    => (∀ x. m x -> n x)
```
KingoftheHomeless, line 126: 👍 Nice catch!

src/Polysemy/Internal/Union.hs
```diff
@@ -111,22 +110,21 @@ weave
     -> (∀ x. s x -> Maybe x)
```
KingoftheHomeless, line 110: While you're at it, can you remove the redundant `Functor m`  constraint here?

src/Polysemy/Internal/Union.hs
```diff
@@ -286,21 +284,22 @@ weaken (Union pr a) = Union (There pr) a
...
-inj :: forall e r m a. (Functor m , Member e r) => e m a -> Union r m a
+inj :: forall e r rInitial a. (Member e r) => e (Sem rInitial) a -> Union r (Sem rInitial) a
 inj e = injWeaving $
   Weaving e (Identity ())
-            (fmap Identity . runIdentity)
-            runIdentity
-            (Just . runIdentity)
+              (fmap Identity . runIdentity)
+              runIdentity
+              (Just . runIdentity)
```
KingoftheHomeless, line 292: I don't like these extra indents: the arguments no longer line up. That said, the original was wrong too. See my comment above for how arguments to a `Weaving` should look like

src/Polysemy/Output.hs
```diff
@@ -107,9 +107,10 @@ runLazyOutputMonoid
     => (o -> m)
     -> Sem (Output o ': r) a
     -> Sem r (m, a)
-runLazyOutputMonoid f = interpretViaLazyWriter $ \(Weaving e s _ ex _) ->
-  case e of
-    Output o -> ex s <$ Lazy.tell (f o)
+runLazyOutputMonoid f = interpretViaLazyWriter $
+  \(Weaving e s _ ex _) ->
+    case e of
+      Output o -> ex s <$ Lazy.tell (f o)
```
KingoftheHomeless, line 113: Not overly fond of this linebreak, but I can accept I have bad taste in this instance.

src/Polysemy/Tagged.hs
```diff
@@ -63,7 +63,8 @@ tagged
 tagged = hoistSem $ \u ->
   case decompCoerce u of
     Right (Weaving e s wv ex ins) ->
-      injWeaving $ Weaving (Tagged @k e) s (tagged @k . wv) ex ins
+      injWeaving $ Weaving
+        (Tagged @k e) s (tagged @k . wv) ex ins
```
KingoftheHomeless, line 67: Please reformat the arguments according to what I said above.

src/Polysemy/Tagged.hs
```diff
@@ -94,7 +95,8 @@ retag
     -> Sem r a
 retag = hoistSem $ \u -> case decomp u of
   Right (Weaving (Tagged e) s wv ex ins) ->
-    injWeaving $ Weaving (Tagged @k2 e) s (retag @_ @k2 . wv) ex ins
+    injWeaving $ Weaving
+      (Tagged @k2 e) s (retag @_ @k2 . wv) ex ins
```
KingoftheHomeless, line 99: Same here.

src/Polysemy/Writer.hs
```diff
@@ -100,7 +100,8 @@ runLazyWriter
      . Monoid o
     => Sem (Writer o ': r) a
     -> Sem r (o, a)
-runLazyWriter = interpretViaLazyWriter $ \(Weaving e s wv ex ins) ->
+runLazyWriter = interpretViaLazyWriter $
+  \(Weaving e s wv ex ins) ->
```
KingoftheHomeless, line 104: If you want to keep this linebreak, indent the case-of.

src/Polysemy/Bundle.hs
```diff
@@ -44,8 +44,8 @@ sendBundle
   -> Sem r a
 sendBundle = hoistSem $ \u -> case decomp u of
   Right (Weaving e s wv ex ins) ->
-    injWeaving $
-      Weaving (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
+    injWeaving $ Weaving
+      (Bundle (membership @e @r') e) s (sendBundle @e @r' . wv) ex ins
```
KingoftheHomeless, line 48: Don't like this linebreak. Arguments to a `Weaving` should all be in one line, or - if that runs too long - one of the following:
```haskell
Weaving e
        s
        wv
        ex
        ins
```

```haskell
Weaving
  e
  s
  wv
  ex
  ins
```

The original doesn't run past 80 chars, so I'd personally keep it on one line.

src/Polysemy/Internal/Union.hs
```diff
@@ -286,21 +284,22 @@ weaken (Union pr a) = Union (There pr) a
...
 ------------------------------------------------------------------------------
 -- | Lift an effect @e@ into a 'Union' capable of holding it,
 -- given an explicit proof that the effect exists in @r@
-injUsing :: forall e r m a. Functor m => ElemOf e r -> e m a -> Union r m a
-injUsing pr e = Union pr $
-  Weaving e (Identity ())
+injUsing :: forall e r rInitial a.
+  ElemOf e r -> e (Sem rInitial) a -> Union r (Sem rInitial) a
+injUsing pr e = Union pr $ Weaving
+            e (Identity ())
```
KingoftheHomeless, line 302: Please reformat the arguments according to above.

src/Polysemy/Internal/Combinators.hs
```diff
@@ -215,7 +218,8 @@ reinterpret2H f (Sem m) = Sem $ \k -> m $ \u ->
 reinterpret2
     :: forall e1 e2 e3 r a
      . FirstOrder e1 "reinterpret2"
-    => (∀ m x. e1 m x -> Sem (e2 ': e3 ': r) x)
+    => (∀ m x. e1 m x ->
+        Sem (e2 ': e3 ': r) x)
```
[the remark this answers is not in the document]
A1kmm: I've now exposed the more powerful version (`Sem rInitial`) on `reinterpret2` and `reinterpret3`.

src/Polysemy/Internal/Union.hs
```diff
@@ -56,50 +57,48 @@ import Polysemy.Internal.CustomErrors
...
-         -- ^ A function for attempting to see inside an @f@. This is no
-         -- guarantees that such a thing will succeed (for example,
-         -- 'Polysemy.Error.Error' might have 'Polysemy.Error.throw'n.)
-       }
-    -> Weaving e n b
+    :: forall f e rBefore a resultType mAfter. (Functor f)
+    => {
+      weaveEffect :: e (Sem rBefore) a
+      -- ^ The original effect GADT originally lifted via
+      -- 'Polysemy.Internal.send'.
```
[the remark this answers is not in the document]
A1kmm: Done.

src/Polysemy/Internal/Union.hs
A1kmm: Done

src/Polysemy/Bundle.hs
```diff
@@ -71,6 +71,7 @@ subsumeBundle
   -> Sem r a
 subsumeBundle = hoistSem $ \u -> hoist subsumeBundle $ case decomp u of
   Right (Weaving (Bundle pr e) s wv ex ins) ->
-    Union (subsumeMembership pr) (Weaving e s wv ex ins)
+    Union (subsumeMembership pr)
+              (Weaving e s wv ex ins)
```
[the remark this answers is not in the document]
A1kmm: Most of these are an artefact of how I introduced `WeavingDetails` and then removed it, without restoring the exact same formatting. I've gone through the PR and made sure that there are no remaining whitespace changes to those files (except for the one you identified as being wrong to start with) - which has reduced the number of files affected down to four.

PR #335, opened 2020-04-26
Add bracket_ to Polysemy.Resource

expipiplus1:
Minor version bump too

### DIFF
--- a/src/Polysemy/Resource.hs
+++ b/src/Polysemy/Resource.hs
@@ -6,6 +6,7 @@ module Polysemy.Resource
 
     -- * Actions
   , bracket
+  , bracket_
   , bracketOnError
   , finally
   , onException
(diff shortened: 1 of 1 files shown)


src/Polysemy/Resource.hs
```diff
@@ -48,6 +49,20 @@ data Resource m a where
 
 makeSem ''Resource
 
+------------------------------------------------------------------------------
+-- | A variant of 'bracket' where the return value from the first computation
+-- is not required.
+--
+-- cf. 'Control.Exception.bracket' and 'Control.Exception.bracket_'
+--
+-- @since 1.3.1.0
```
googleson78, line 58: I think we usually leave this as a `@since TODO` and not change the version `package.yaml`/`.cabal`, because there might not be a `1.3.1.0` release on hackage, making this haddock a bit confusing.
@KingoftheHomeless @TheMatten ?

KingoftheHomeless: I doubt there's gonna be a `1.3.1.0`, unless we discover something we desperately need to fix.
So yes, leave this as a TODO.

expipiplus1: No problem, fixed

googleson78:
Don't see a reason not to merge this - if normal bracket works, this should too.

Thanks and sorry for the stall!

expipiplus1:
np, thanks!

PR #337, opened 2020-04-28
Build with 8.10

### LINKED ISSUE #336: Build failure on GHC 8.10 (win64)
I'm following along some blog posts and saw that 8.10 is when we get cost-free polysemy abstractions, so I dutifully grabbed it, and cabal, and manually fixed up my path, got `polysemy` installed and imported into a `ghci` session, but when installing the plugin I got the following errors:

    > cabal install --lib polysemy-plugin

    Preprocessing library for polysemy-plugin-0.2.5.0..
    Building library for polysemy-plugin-0.2.5.0..
    [1 of 7] Compiling Paths_polysemy_plugin ( dist\build\autogen\Paths_polysemy_plugin.hs, dist\build\Paths_polysemy_plugin.o )
    [2 of 7] Compiling Polysemy.Plugin.Fundep.Stuff ( src\Polysemy\Plugin\Fundep\Stuff.hs, dist\build\Polysemy\Plugin\Fundep\Stuff.o )
    [3 of 7] Compiling Polysemy.Plugin.Fundep.Unification ( src\Polysemy\Plugin\Fundep\Unification.hs, dist\build\Polysemy\Plugin\Fundep\Unification.o )
    
    src\Polysemy\Plugin\Fundep\Unification.hs:133:23: error:
        Not in scope: type constructor or class `Ct'
        |
    133 |     -> [(Unification, Ct)]
        |                       ^^
    
    src\Polysemy\Plugin\Fundep\Unification.hs:134:25: error:
        Not in scope: type constructor or class `Ct'
        |
    134 |     -> ([Unification], [Ct])
        |                         ^^
    cabal: Failed to build polysemy-plugin-0.2.5.0. See the build log above for
    details.

_Edit: I tried checking this repo out @ 016c16fbb1b57a0d728e57e2cf8e36453e8edd8d thinking the version in Hackage might be different, but I got the same result._

googleson78:
Fixes #336 

* Ct (and friends) now comes from Constraints
* There was an ifdef'd part that uses a function "bool" which isn't imported when the ifdef triggers

### DIFF
--- a/polysemy-plugin/src/Polysemy/Plugin.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin.hs
@@ -64,6 +64,7 @@ module Polysemy.Plugin
 import Polysemy.Plugin.Fundep
 #if __GLASGOW_HASKELL__ >= 810
 import Polysemy.Plugin.Phases
+import Data.Bool (bool)
 #endif
 
 import GhcPlugins
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ cabal: "3.0"
 
 matrix:
   include:
+    - ghc: "8.10.1"
     - ghc: "8.8.1"
     - ghc: "8.6.5"
     - ghc: "8.4.4"
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -46,6 +46,7 @@ import           Polysemy.Plugin.Fundep.Utils
 import           TcEvidence
 import           TcPluginM (TcPluginM, tcPluginIO)
 import           TcRnTypes
+import           Constraint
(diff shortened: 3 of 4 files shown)


polysemy-plugin/src/Polysemy/Plugin.hs
```diff
@@ -64,6 +64,7 @@ module Polysemy.Plugin
 import Polysemy.Plugin.Fundep
 #if __GLASGOW_HASKELL__ >= 810
 import Polysemy.Plugin.Phases
+import Data.Bool (bool)
```
googleson78, line 67: @TheMatten could you take a look at this? I took a guess as to which `bool` is supposed to be used here, but I don't really know what the code does, and `blame` says you wrote the part with `bool`

TheMatten: Yeah, it should be that one :+1:

PR #345, opened 2020-06-16
implement some Control.Exception combinators

incertia:
implements the `bracket`, `bracket_`, `bracketOnError`, `finally`, and `onException` combinators from `Control.Exception`


incertia:
wait a second... this is taken care of by `Resource`...

incertia:
perhaps it might be worthwhile to add a `runResourceWithError :: Member (Error e) r => Sem (Resource ': r) a -> Sem r a` interpreter?

PR #348, opened 2020-06-18
Fix warnings from upgrading to ghc 8.8.3

googleson78:


src/Polysemy/Internal/Union.hs
```diff
@@ -265,14 +265,13 @@ decomp (Union p a) =
...
 extract :: Union '[e] m a -> Weaving e m a
 extract (Union Here a)   = a
-extract (Union (There g) _) = case g of {}
 {-# INLINE extract #-}
 
 
 ------------------------------------------------------------------------------
 -- | An empty union contains nothing, so this function is uncallable.
 absurdU :: Union '[] m a -> b
-absurdU (Union pr _) = case pr of {}
```
googleson78, line 275: Is making this less strict an issue?

googleson78:
'doh, I always forget about multiple ghc versions..

googleson78:
Not sure if I should add `#ifdef`s to get rid of the warnings..
Opinions on this? @KingoftheHomeless @TheMatten

src/Polysemy/Internal/Union.hs
TheMatten: Doesn't GHC show a warning about non-exhaustive match?

TheMatten:
What warnings do you get?

src/Polysemy/Internal/Union.hs
googleson78: It seems `ghc` can now tell that `ElemOf` (and by extension `Union`) cannot produce a value with an empty list as its index! Very nice to see this

googleson78:
Redundant/impossible match, + redundant import warnings, because more stuff is reexported in newer base (it seems)

src/Polysemy/Internal/Union.hs
TheMatten: Ah, I see - really nice! But I guess that it may be more convenient to just keep it as it was while we're supporting older versions, because it doesn't really change anything interesting about the behaviour.

googleson78: To be honest I would also prefer to have `-Werror` in CI to catch other mistakes, but then we would definitely require `ifdef`s here.

TheMatten:
We're already using `CPP`, so we may just as well just go for `#ifdef` - it will at least make code related to older versions visible/searchable, if we were to drop support for them at some point.

src/Polysemy/Internal/Union.hs
TheMatten: I mean, the older version doesn't emit any warnings on new GHC, does it?

TheMatten:
@googleson78 do we add that `CPP` or do we leave it as-is for now? I guess those warnings work as a sort of reminder that those matches can be dropped once we stop supporting older GHCs.

googleson78:
I `ifdef`'d them for the "easier to remove in the future" reason. Let's just be sure that CI passes

PR #359, opened 2020-07-10
Delete redundant imports

jeremyschlatter:


googleson78:
Redundant, but only on certain ghcs! I purposefully left them as redundant imports, because I'm not sure whether using CPP to remove redundant imports is worth it.

jeremyschlatter:
Oh, I didn't realize that! In that case I agree, keeping them in seems like the right call.

PR #360, opened 2020-07-10
Small code simplifications

jeremyschlatter:
These are all replacements of the form

    before: fmap f $ x
    after:  f <$> x

Minor change, but feels marginally easier to read to me.


TheMatten:
@KingoftheHomeless @googleson78 any objections? I think we could extend that to some other instances of saturated `fmap`.

googleson78:
I don't mind either

TheMatten:
@jeremyschlatter thanks!

PR #364, opened 2020-07-21
Update README

### LINKED ISSUE #358: Update readme with more accurate performance and comparison details
I use both fused effects and polysemy. I like both for different reasons. I recommend polysemy in many cases.

However, polysemy does not seem to be very accurate about its claims of performance in the readme. These claims seem to be the initial goals and beliefs about performance, that have not actually been realised? 

Both @lexi-lambda's [talk](https://www.youtube.com/watch?v=0jI-AlWEwYI) and @isovector [blog post](https://reasonablypolymorphic.com/blog/mea-culpa/) suggest that polysemy is in fact 'slow' in many? cases.

I think the fused effects readme is pretty reasonable here when comparing itself with polysemy.

> Like polysemy, fused-effects is a batteries-included effect system capable of scoped, reinterpretable algebraic effects.

> As of GHC 8.8, fused-effects outperforms polysemy, though new effects take more code to define in fused-effects than polysemy (though the Control.Carrier.Interpret module provides a low-friction API for rapid prototyping of new effects). Like freer-simple and unlike fused-effects, polysemy provides custom type errors if a given effect invocation is ambigous or invalid in the current context.

Specifically the following claims seem to be questionable.

- `polysemy is a library for writing high-power, low-boilerplate, zero-cost, domain specific languages` - is it really zero cost?
- `It's like mtl but composes better, requires less boilerplate, and avoids the O(n^2) instances problem.` - does it have the same performance as mtl then?
- `It's like freer-simple but more powerful and 35x faster.` - is this 35x accurate?
- `It's like fused-effects but with an order of magnitude less boilerplate.` - I think the `order of magnitude` bit is slightly overstated. Does that mean it takes 1/10th of the code in polysemy to work with effects? (You could possibly argue this as fused effects has some edge cases that require a lot of boilerplate. On another note I think fused effects is more complex to learn + use where polysemy is very intuitive). And again, does this mean polysemy is as fast as fused effects?
- `Effects are zero-cost, meaning that GHC1 can optimize away the entire abstraction at compile time. 1: Unfortunately this is not true in GHC 8.6.3, but will be true in GHC 8.10.1.` - Is it true the polysemy is zero cost in ghc 8.10.1 ?
- The `Stellar Engineering - Aligning the stars to optimize polysemy away - Several things need to be in place to fully realize our performance goals` seems to imply these steps make it zero cost, is that true?

I hope this is not seen as inflammatory! :sweat_smile: I just think that accuracy (if it is indeed inaccurate) here is better for everyone involved in polysemy, other effect libraries and the wider Haskell community.

TheMatten:
- fix #358
- apply some lint suggestions, try to fit 80 characters limit used in other files
- change formatting to make things slightly more readable
- fix links pointing to old URL and stuff

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Or by adding `-fplugin=Polysemy.Plugin` to your `package.yaml`/`.cabal` file `gh
   to start thinking about polysemy.
 - Sandy Maguire, the author, gave a talk on some of the
   [performance implementation](https://www.youtube.com/watch?v=-dHFOjcK6pA)
-- He have also written
+- He has also written
   [some](http://reasonablypolymorphic.com/blog/freer-higher-order-effects/)
   [blog posts](http://reasonablypolymorphic.com/blog/tactics/) on other
   implementation details.
@@ -162,12 +162,13 @@ program = catch @CustomException work \e -> writeTTY $ "Caught " ++ show e
       _             -> writeTTY input *> writeTTY "no exceptions"
 
 main :: IO (Either CustomException ())
-main = runFinal
-     . embedToFinal @IO
-     . resourceToIOFinal
-     . errorToIOFinal @CustomException
-     . teletypeToIO
-     $ program
+main
+  = runFinal
+  . embedToFinal @IO
+  . resourceToIOFinal
+  . errorToIOFinal @CustomException
+  . teletypeToIO
+  $ program
 ```
 
 Easy.
@@ -229,16 +230,17 @@ You're going to want to stick all of this into your `package.yaml` file.
 
 ## *What about performance?* ([TL;DR](#tldr))
 
-Previous versions of this `README` mentioned **library being** ***zero-cost***, as in
-having no visible effect on performance. While this was the original motivation
-and main factor in implementation of this library, it turned out that
+Previous versions of this `README` mentioned **the library being**
+***zero-cost***, as in having no visible effect on performance. While this was
+the original motivation and main factor in implementation of this library, it
+turned out that
 [**optimizations** we depend on](https://reasonablypolymorphic.com/blog/specialization/),
 while showing amazing results in small benchmarks, **don't work in
 [bigger, multi-module programs](https://github.com/ghc-proposals/ghc-proposals/pull/313#issuecomment-590143835)**,
 what greatly limits their usefulness.
 
 What's more interesting though is that
-this **isn't `polysemy`-specific** problem - basically **all popular effects
+this **isn't a `polysemy`-specific** problem - basically **all popular effects
 libraries** ended up being bitten by variation of this problem in one way or
 another, resulting in
 [visible drop in performance](https://github.com/lexi-lambda/ghc-proposals/blob/delimited-continuation-primops/proposals/0000-delimited-continuation-primops.md#putting-numbers-to-the-cost)
@@ -250,7 +252,7 @@ One factor may be that while GHC's optimizer is
 very, very good in general in optimizing all sorts of abstraction, it's
 relatively complex and hard to predict - authors of libraries may have not
 deemed location of code relevant, even though it had big effect at the end.
-The other is that maybe **it doesn't matter as much** as I we like to tell
+The other is that maybe **it doesn't matter as much** as we like to tell
 ourselves. Many of these effects
 libraries are used in production and they're doing just fine, because maximum
 performance usually matters in small, controlled areas of code, that often
--- a/README.md
+++ b/README.md
@@ -292,24 +292,6 @@ GHC with features that will allow for creation of effects implementation with
 stable and strong performance. It's what `polysemy` may choose at some point,
 but it will probably require few breaking changes.
 
-## Stellar Engineering - Aligning the stars to optimize `polysemy` ~~away~~ better
-
-Several things need to be in place to fully realize our performance goals:
-
-- GHC Version
-  - GHC 8.9+
-- Your code
-  - The module you want to be optimized needs to import `Polysemy.Internal`
-    somewhere in its dependency tree (sufficient to `import Polysemy`)
-- GHC Flags
-  - `-O` or `-O2`
-  - `-flate-specialise` (this should be automatically turned on by the plugin,
-    but it's worth mentioning)
-- Plugin
-  - `-fplugin=Polysemy.Plugin`
-- Additional concerns:
-  - additional core passes (turned on by the plugin)
-
 ## Acknowledgements, citations, and related work
 
 The following is a non-exhaustive list of people and works that have had a
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ the
 and nicely solves the
 [trouble with typed errors](https://www.parsonsmatt.org/2018/11/03/trouble_with_typed_errors.html).
 
-Concerned about type inference? `polysemy` comes with companion
+Concerned about type inference? `polysemy` comes with its companion
 [`polysemy-plugin`](https://github.com/isovector/polysemy/tree/master/polysemy-plugin),
 which helps it perform just as well as `mtl`'s! Add `polysemy-plugin` to your
 `package.yaml` or `.cabal` file's `dependencies` section to use. Then turn it on with a pragma in your source files:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,76 +19,72 @@
...
-- I've given a talk on some of the [performance implementation](https://www.youtube.com/watch?v=-dHFOjcK6pA)
-- I've also written [some](http://reasonablypolymorphic.com/blog/freer-higher-order-effects/) [blog posts](http://reasonablypolymorphic.com/blog/tactics/) on other implementation details.
-
+- Raghu Kaippully wrote a beginner friendly
+  [tutorial](https://haskell-explained.gitlab.io/blog/posts/2019/07/28/polysemy-is-cool-part-1/index.html).
+- Paweł Szulc gave a [great talk](https://youtu.be/idU7GdlfP9Q?t=1394) on how
+  to start thinking about polysemy.
+- Sandy Maguire, the author, gave a talk on some of the
+  [performance implementation](https://www.youtube.com/watch?v=-dHFOjcK6pA)
+- He have also written
```
googleson78, line 72: Typo: have -> has

README.md
```diff
@@ -101,15 +97,19 @@ data Teletype m a where
 makeSem ''Teletype
 
 teletypeToIO :: Member (Embed IO) r => Sem (Teletype ': r) a -> Sem r a
-teletypeToIO = interpret $ \case
+teletypeToIO = interpret \case
```
googleson78, line 100: This innocuous change drops support for ghc <8.6. I'm not sure if we care though?

googleson78: But if this is the only thing breaking e.g. ghc 8.4.3, I'm against it.

README.md
```diff
@@ -153,27 +152,26 @@ import Polysemy.Resource
...
 
 main :: IO (Either CustomException ())
-main =
-    runFinal
-  . embedToFinal @IO
-  . resourceToIOFinal
-  . errorToIOFinal @CustomException
-  . teletypeToIO
-  $ program
+main = runFinal
```
googleson78, line 165: I think this
```haskell
main
  = runFinal
  . embedToFinal @IO
  ...
```
would be more consistent with the other code above (`runTeletypePure`)

README.md
```diff
@@ -230,36 +227,105 @@ You're going to want to stick all of this into your `package.yaml` file.
     - TypeFamilies
 ```
 
-## Stellar Engineering - Aligning the stars to optimize `polysemy` away
+## *What about performance?* ([TL;DR](#tldr))
+
+Previous versions of this `README` mentioned **library being** ***zero-cost***, as in
```
googleson78, line 232: ... mentioned **the** library ...

README.md
```diff
@@ -230,36 +227,105 @@ You're going to want to stick all of this into your `package.yaml` file.
...
+having no visible effect on performance. While this was the original motivation
+and main factor in implementation of this library, it turned out that
+[**optimizations** we depend on](https://reasonablypolymorphic.com/blog/specialization/),
+while showing amazing results in small benchmarks, **don't work in
+[bigger, multi-module programs](https://github.com/ghc-proposals/ghc-proposals/pull/313#issuecomment-590143835)**,
+what greatly limits their usefulness.
+
+What's more interesting though is that
+this **isn't `polysemy`-specific** problem - basically **all popular effects
```
googleson78, line 241: this isn't **a** polysemy-specific problem

README.md
```diff
@@ -230,36 +227,105 @@ You're going to want to stick all of this into your `package.yaml` file.
...
+[visible drop in performance](https://github.com/lexi-lambda/ghc-proposals/blob/delimited-continuation-primops/proposals/0000-delimited-continuation-primops.md#putting-numbers-to-the-cost)
+compared to equivalent code without use of effect systems.
+
+*Why nobody noticed this for such a long time?*
+
+One factor may be that while GHC's optimizer is
+very, very good in general in optimizing all sorts of abstraction, it's
+relatively complex and hard to predict - authors of libraries may have not
+deemed location of code relevant, even though it had big effect at the end.
+The other is that maybe **it doesn't matter as much** as I we like to tell
```
googleson78, line 253: I think you forgot to delete the `I` here?

README.md
TheMatten: Funny thing is, `BlockArguments` were already in `LANGUAGE` pragma on top of the snippet. And I'm not sure if anyone starts new project with versions of GHC older than 8.6 today.

README.md
```diff
@@ -230,36 +227,105 @@ You're going to want to stick all of this into your `package.yaml` file.
...
+
+### TL;DR
+
+Basically all current effects libraries (including `polysemy` and
+even `mtl`) got performance wrong - **but**, there's ongoing work on extending
+GHC with features that will allow for creation of effects implementation with
+stable and strong performance. It's what `polysemy` may choose at some point,
+but it will probably require few breaking changes.
+
+## Stellar Engineering - Aligning the stars to optimize `polysemy` ~~away~~ better
```
googleson78, line 293: Perhaps we could mention `-fexponse-all-unfoldings` as a last resort, which afaiu is one of the "barriers" required for getting zero-cost effects?

TheMatten: If I remember right, we were talking about dropping `loopbreaker` (which should be needed to make this work) because it's not going to be useful in general anyway and it's an ugly hack - I actually wonder how much of those flags are still going to make any difference in that case. And even if we kept supporting it, I'm worried that if we tell people to enable `-fexpose-all-unfoldings` for small projects, they will do so just to later find out that their project grew too big to compile with it.

README.md
```diff
@@ -19,76 +19,72 @@
...
-which should perform just as well as `mtl`'s! Add `polysemy-plugin` to your package.yaml
-or .cabal file's dependencies section to use. Then turn it on with a pragma in your source-files:
+hacks~~ band-aids like
+[classy lenses](http://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-TH.html#v:makeClassy),
+the
+[`ReaderT` pattern](https://www.fpcomplete.com/blog/2017/06/readert-design-pattern)
+and nicely solves the
+[trouble with typed errors](https://www.parsonsmatt.org/2018/11/03/trouble_with_typed_errors.html).
+
+Concerned about type inference? `polysemy` comes with companion
```
isovector, line 45:
```suggestion
Concerned about type inference? `polysemy` comes with its companion
```

README.md
isovector: lgtm since it's used in teletype pure anyway

README.md
```diff
@@ -19,76 +19,72 @@
...
+[Necessary Language Extensions](https://github.com/polysemy-research/polysemy#necessary-language-extensions)
 before trying these yourself!
 
 Teletype effect:
 
 ```haskell
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE LambdaCase, BlockArguments #-}
-{-# LANGUAGE GADTs, FlexibleContexts, TypeOperators, DataKinds, PolyKinds #-}
+{-# LANGUAGE TemplateHaskell, LambdaCase, BlockArguments, GADTs
```
isovector, line 86: this example isn't very helpful, really

README.md
```diff
@@ -230,36 +228,106 @@ You're going to want to stick all of this into your `package.yaml` file.
...
+what greatly limits their usefulness.
+
+What's more interesting though is that
+this **isn't a `polysemy`-specific** problem - basically **all popular effects
+libraries** ended up being bitten by variation of this problem in one way or
+another, resulting in
+[visible drop in performance](https://github.com/lexi-lambda/ghc-proposals/blob/delimited-continuation-primops/proposals/0000-delimited-continuation-primops.md#putting-numbers-to-the-cost)
+compared to equivalent code without use of effect systems.
+
+*Why nobody noticed this for such a long time?*
```
isovector, line 249:
```suggestion
*Why did nobody notice this?*
```

README.md
isovector: let's just remove this entire section

README.md
TheMatten: Hmm, maybe it's good to have "dumb" example first, not doing much more than interpreting simple effect, as a sort of showcase of `mkSem` and `interpret`.

README.md
TheMatten: Okay - maybe we can start another issue for creating new set of flags that is known to somewhat help in absence of `loopbreaker`.

README.md
TheMatten: Should I keep it as it is?

PR #369, opened 2020-07-28
raise2Under and raise3Under

tek:
These combinators are like raiseUnder, but place the new effect at the
second and third position in the list.

This is for interpreters that consume two effects, delegating to two other interpreters, who share a dependency which is then interpreted by the outer interpreter.

Please advise on what else to do, naming, general sense and so on!


TheMatten:
[relevant discussion](https://funprog.srid.ca/polysemy/raiseunder-and-multiple-effect-interpretation.html)

PR #370, opened 2020-07-28
Add 'Raise' and 'Subsume'

TheMatten:
See https://funprog.srid.ca/polysemy/raiseunder-and-multiple-effect-interpretation.html and #369.

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -26,7 +26,6 @@ module Polysemy
   , embedFinal
 
     -- * Lifting
-
   , raise
   , raiseUnder
   , raiseUnder2
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -32,6 +32,8 @@ module Polysemy
   , raiseUnder3
   , raise2Under
   , raise3Under
+  , raise'
+  , subsume'
 
     -- * Trivial Interpretation
   , subsume
@@ -148,4 +150,3 @@ import Polysemy.Internal.Forklift
 import Polysemy.Internal.Kind
 import Polysemy.Internal.Tactics
 import Polysemy.Internal.TH.Effect
-
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -5,35 +5,36 @@ module Polysemy
   , MemberWithError
   , Members
 
-  -- * Running Sem
+    -- * Running Sem
   , run
   , runM
   , runFinal
 
-  -- * Type synonyms for user convenience
+    -- * Type synonyms for user convenience
   , InterpreterFor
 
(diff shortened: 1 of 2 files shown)


tek:
:clap:

TheMatten:
`raiseNUnder` have to stay as they are, because `Find` can't believe us about effects above being different from each other.

tek:
I wonder whether a better name `raiseNUnder` would be `insertAtN`…it's not a very intuitive name

TheMatten:
I like @googleson78's realization about what `raiseNUnder` and `raiseUnderN` should mean - we could pursue that in v2 in keep it for now as-is in v1. Technically it generalizes too - `raiseNUnderM` :stuck_out_tongue_winking_eye:.

tek:
guess I'm unclear about why it's called `raise` in the first place

TheMatten:
I guess [this](https://hackage.haskell.org/package/freer-simple-1.2.1.1/docs/Control-Monad-Freer.html#v:raise) is the reason :smile:

tek:
I see :upside_down_face: I would argue that the name `raiseUnder` suggests, purely rhetorically, that the `e` is raised, while it apparently is the `r`. Not saying that my intuition is a reference here or that I have a better solution, but the `UnderN` `NUnder` confusion suggests that there might be room for improvement :slightly_smiling_face:

KingoftheHomeless:
Actually, wait. We don't need this, do we? The `raiseUnderN` class of functions can be achieved through repeated application of `raiseUnder`. I don't see much point in this if it doesn't also solve the `raiseNUnder` functions.

TheMatten:
We can't change `raiseNUnder` because of how `Find` works, but we can actually use `subsume'` in same way at call sites, where top of the stack has concrete effects. So `subsume'` is still works as `raiseNUnder` generalization.

When it comes to names, I would prefer something short - that `'` can be seen as placeholder for number/postfix (possibly could be replaced with `_`).

KingoftheHomeless:
Oh, well that's fine then.

I'd prefer `_` instead of `'`.

TheMatten:
Even in case with polymorphic "tail", that typeclass duo should resolve into concrete instance - and GHC is really good at inlining instances (see Generics).

tek:
> Oh, well that's fine then.
> 
> I'd prefer `_` instead of `'`.

why? isn't `_` an indicator of `void`?

TheMatten:
@tek I would say that depends a lot on context - and in this context connection to `void` doesn't make any sense. Instead, I would say `_` looks a lot like a "placeholder" for some postfix.

TheMatten:
Can be considered ready?

Issue #368, opened 2020-07-27
Memory leak with Applicative (*>)


goertzenator:
I have a program that uses Polysemy and Streaming, and we've observed a memory leak when using the Streaming function `mapM_`.  When this function is pulled into my code, I can fix the leak by changing `*>` to `>>`.  Shouldn't these two operators do the same thing for `Sem`?

``` Haskell
myMapM_ :: Monad m => (a -> m b) -> Stream (Of a) m r -> m r
myMapM_ f = loop where
  loop str = case str of
    Return r       -> return r
    Effect m       -> m >>= loop
    Step (a S.:> as) -> f a >> loop as -- older impl, doesn't leak
 -- Step (a S.:> as) -> f a *> loop as -- newer impl, leaks!
```

The heap plot using `>>` looks like...
![image](https://user-images.githubusercontent.com/605072/88592460-ab06be00-d023-11ea-8cfe-fbaa34327460.png)

... and the heap plot using `*>` looks like...

![image](https://user-images.githubusercontent.com/605072/88592935-6596c080-d024-11ea-9dfc-2c9be6422dae.png)

KingoftheHomeless:
This is probably due to `Sem`'s `Applicative` instance relying on the default implementations of `*>` and `<*`.

I'd welcome a small pull request that adds the following under `instance Applicative (Sem r)` in `Polysemy.Internal`:

```haskell
liftA2 f ma mb = Sem $ \k -> liftA2 (runSem ma k) (runSem mb k)

ma <* mb = Sem $ \k -> runSem ma k <* runSem mb k

ma *> mb = Sem $ \k -> runSem ma k *> runSem mb k
```
Alternatively, many monads are bad at optimizing `*>`, (the default implementation is _really_ slow if `(<*>)` is defined as `ap` for the underlying monad), so the above definition may not be safe enough. An alternative is using:

```haskell
ma *> mb = Sem $ \k -> runSem ma k >>= \_ -> runSem mb k
```
Which is guaranteed to perform as good as `>>=`.

If you have the patience to fork `polysemy`, test out the changes and see if it makes a difference. Otherwise, a pull request with the former definitions are probably good enough.

goertzenator:
Thanks, these additions seem to work and I'll do a PR.  Any objection to just defining `*>` as...

``` haskell
(*>) = (>>)
```

KingoftheHomeless:
[...]

PR #372, opened 2020-07-29
add Applicative methods

goertzenator:
Add Applicative `liftA2`, `(<*)`, and `(*>)` as directed in issue https://github.com/polysemy-research/polysemy/issues/368 .

### DIFF
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -274,9 +274,6 @@ instance Applicative (Sem f) where
   {-# INLINE (*>) #-}
 
 instance Monad (Sem f) where
-  return = pure
-  {-# INLINE return #-}
-
   Sem ma >>= f = Sem $ \k -> do
     z <- ma k
     runSem (f z) k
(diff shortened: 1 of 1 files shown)


src/Polysemy/Internal.hs
```diff
@@ -262,6 +262,14 @@ instance Applicative (Sem f) where
   Sem f <*> Sem a = Sem $ \k -> f k <*> a k
   {-# INLINE (<*>) #-}
 
+  liftA2 f ma mb = Sem $ \k -> liftA2 f (runSem ma k) (runSem mb k)
+  {-# INLINE liftA2 #-}
+
+  ma <* mb = Sem $ \k -> runSem ma k <* runSem mb k
+  {-# INLINE (<*) #-}
+
+  ma *> mb = Sem $ \k -> runSem ma k >>= \_ -> runSem mb k
```
KingoftheHomeless, line 271: Leave a comment as to why we do this. You can reference #368.

src/Polysemy/Internal.hs
```diff
@@ -262,6 +262,14 @@ instance Applicative (Sem f) where
...
+  {-# INLINE (*>) #-}
 
 instance Monad (Sem f) where
   return = pure
```
KingoftheHomeless, line 275: Could you also remove this definition?

KingoftheHomeless: (That is, `return = pure`)

PR #375, opened 2020-09-08
correct two switched INLINE pragmas

tek:
I assume this isn't some magic trick.


googleson78:
These should actually work regardless of where they are placed, right?

tek:
yes, it is purely cosmetic

TheMatten (approved):
Should be cosmetic, but I guess it makes a lot of sense from a clarity standpoint :D

tek:
my thoughts exactly :smiley_cat:

Issue #378, opened 2020-10-07
Async effect leaks implementation detail (and therefore prevents pure interpretation)


alaendle:
This is more something that could be discussed - I really wonder why the Async effect presents `Control.Concurrent.Async.Async` in it's API. This offers potential misuse by e.g. directly awaiting the result; so from my perspective I would favour the encapsulation into a type that just represents a simple handle and couldn't be used by the client otherwise.

The other improvement this might offer is that it might be possible to interpret the Async effect in a pure way. My guess is that nothing contradicts that everything inside async is interpreted immediately as async gets interpreted, and wait just returns this result.

So please don't get me wrong, I just stumbled across this question and was wondering why things are as they are. In no way did I want to criticize. Also I should mention that I'm relatively new to FP and my point of view corresponds to a simple polysemy user with an strong OOP background; so maybe my assumptions are just wrong. I'm really looking forward to learn by hearing other peoples opinion.

tek:
absolutely reasonable! in my opinion it should be `Async handle :: Effect`. This way you can still depend on the concrete implementation.

tek:
mh, probably `(handle :: * -> *)`

tek:
created a pull request to showcase how it would look.

adamConnerSax:
Question, though I'm not sure it's yet coherent:
I've got a library that uses polysemy to provide a set of useful effects (Pandoc doc writing, caching, unique-id generation, logging) for building data-driven html.  I also expose/wrap async so I can do concurrent computations when building analysis-heavy documents.  Are you suggesting that I should rather create an effect to which I can submit computations-- which need to all be complete before continuing--and then choose in the interpretation stage if that effect uses async or not?  
I think that makes sense, I just want to make sure I understand.

tek:
I would assume that `sequenceConcurrently` is the main reason for this issue, so abstracting that would probably be the most sensible solution

tek:
parallelism and asynchronicity are quite different from the modeling perspective, but in Haskell I guess they are notoriously underspecified

adamConnerSax:
Yes, ```sequenceConcurrently```!
And I do vaguely understand that parallelism and asynchronicity are different. 
[...]

PR #379, opened 2020-10-07
don't expose the implementation on Async

tek:
see #378. adds a parameter to Async that replaces A.Async.

Problem, of course, is that `async` now needs the plugin for inference to succeed.

### DIFF
--- a/src/Polysemy/Async.hs
+++ b/src/Polysemy/Async.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
 
 module Polysemy.Async
   ( -- * Effect
@@ -31,10 +32,10 @@ import           Polysemy.Final
 -- 'Polysemy.Error.Error' effect didn't fail locally.
 --
 -- @since 0.5.0.0
-data Async m a where
-  Async :: m a -> Async m (A.Async (Maybe a))
-  Await :: A.Async a -> Async m a
-  Cancel :: A.Async a -> Async m ()
+data Async (h :: * -> *) m a where
+  Async :: m a -> Async h m (h (Maybe a))
+  Await :: h a -> Async h m a
+  Cancel :: h a -> Async h m ()
 
 makeSem ''Async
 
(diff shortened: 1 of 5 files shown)


googleson78:
Can you implement the "additional improvement" #378 had in mind?
```
The other improvement this might offer is that it might be possible to interpret the Async effect in a pure way. My guess is that nothing contradicts that everything inside async is interpreted immediately as async gets interpreted, and wait just returns this result.
```

Also, for interests sake - have you thought of any other `h`s you might use instead of `Async` here?

tek:
right! that would be prudent to provide :sweat_smile: 

The first thing that comes to my mind for the other `h`s would be a bare `ThreadId`, however sensible that might be.
Searching on Hackage suggests for example https://hackage.haskell.org/package/async-pool-0.9.1/docs/Control-Concurrent-Async-Pool.html, which appears to use a custom `Async` containing additional data, while still using `async` as backend.
I would assume that there might be more of those. No idea how practical it might be to abstract over those, but it does give the OP a point.

tek:
alright, I tried to approximate the semantics of the native interpreter, in that the thunk is only evaluated in the `Await` interpretation.

The easiest way to achieve that seemed to be to use `Sem r` as handle :slightly_smiling_face: haven't tried it out yet though.

tek:
mh, just realized that OP said to evaluate in `Async.Async`. what do you think?

alaendle:
> Problem, of course, is that `async` now needs the plugin for inference to succeed.

Maybe my thoughts are too briefly, but what about a simple solution like:
```haskell
newtype AsyncVal a = AsyncVal (Either (A.Async a) a)

data Async m a where
  Async :: m a -> Async m (AsyncVal a)
  Await :: AsyncVal a -> Async m a

makeSem ''Async
```
Here the AsyncVal constructor just wouldn't be exported and the pure interpreter uses `Right` values, while the concurrent interpreter uses `Left`values. For sure this has the disadvantage that no other meaningful interpreters could be implemented by clients. Main difference compared to the current draft would be that  the users of `async` wouldn't need to decide which (pure/impure) interpreter they need - but I'm unsure if this is a good or a bad thing. I tend to like it because other effects also don't (need to) know how the got interpreted - and I'm not sure if `Async` is special enough to justify a exceptional case.

googleson78:
The disadvantage with the `Either` approach is that now you need to handle "invalid" cases internally (e.g. you get `A.Async`s in your pure interpreter that you need to do something to), because your interpreter can't know that it's always going to be one of the two for the entire body. With the `h a` approach you can specialise `h ~ Identity` or `h ~ Async` in your interpreters.

An interesting question that @tek is referring to is also this: in the hypothetical pure interpreter, should the inner actions get run when you call `async`, or when you `await` the result of an `async`?

alaendle:
Just fyi (@tek), I played around with your implementation and merged the latest bits from main (https://github.com/alaendle/polysemy/tree/generic-async) - and it works nicely in conjunction with https://github.com/input-output-hk/io-sim - so again thanks for your effort 😃 However you are right - without the plugin I really needed a lot of type annotations.

tek:
cool :slightly_smiling_face:

PR #382, opened 2020-10-23
Remove optimization phases, update `package.yaml`-s

TheMatten:
Fixes polysemy-research/polysemy-zoo#65 by removing optimization phases, which weren't really used up until now and don't solve the original problem anyway. Additionally I've fixed warnings in `package.yaml` files and fixed bound on `th-abstraction` being modified only in `polysemy.cabal`.


googleson78:
travis not passing, but of course it doesn't show up in the PR for some reason :)
https://travis-ci.org/github/polysemy-research/polysemy/builds/738241715

package.yaml
```diff
@@ -119,6 +119,8 @@ tests:
     - inspection-testing >= 0.4.2 && < 0.5
     - hspec >= 2.6.0 && < 3
     - doctest >= 0.16.0.1 && < 0.17
+    generated-other-modules:
+    - Build_doctests
```
googleson78, line 123: What does adding this here do?

TheMatten: I was getting this warning for both packages:
```
Warning: The following modules should be added to exposed-modules or other-modules in /home/matej/Haskell/polysemy/polysemy.cabal:
             - In test:polysemy-test:
                 Build_doctests
             - In test:polysemy-test:
                 Build_doctests
         
         Missing modules in the cabal file are likely to cause undefined reference errors from the linker, along with other problems.
```
Should I split it? I find it to be a pretty small change.

TheMatten:
> travis not passing, but of course it doesn't show up in the PR for some reason :)
> https://travis-ci.org/github/polysemy-research/polysemy/builds/738241715

```
Warning: Packages using 'cabal-version: 2.0' and the autogenerated module
Paths_* must include it also on the 'autogen-modules' field besides
'exposed-modules' and 'other-modules'. This specifies that the module does not
come with the package and is generated on setup. Modules built with a custom
Setup.hs script also go here to ensure that commands like sdist don't fail.
```

So I guess I should just add them to `generated-other-modules` too? :smile:

TheMatten:
CI is green now :tada:

package.yaml
googleson78: I don't mind, I was just wondering

PR #384, opened 2020-10-29
simpler version of interpretH

tek:
a small suggestion for a version of `interpretH` that provides functions that take care of the ceremony for running the interpreter on the higher order thunk.
prompted by people displaying difficulties with dealing with that.

### DIFF
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -14,6 +14,7 @@ module Polysemy.Internal.Combinators
 
     -- * Higher order
   , interpretH
+  , interpretHSimple
   , interceptH
   , reinterpretH
   , reinterpret2H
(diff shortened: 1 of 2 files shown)


googleson78:
I'm in favour of this, I think a lot of the use cases are like this.

KingoftheHomeless:
I'd like to hold off for a bit. Undoubtedly, _something_ needs to be done about `interpretH`; we need to find a cleaner interface, implement that as the new standard, and be _certain_ that we'll be satisfied with it. `interpretHSimple` is better, but is still mired in the need to thread state and such, which is another big thing people get stuck on.

I'm considering something I mentioned in Zulip once:
```haskell
interpretHGood :: (  forall z q x
                   . (forall x. z y -> Sem (Opaque q ': r) y)
                  -> e z x
                  -> Sem (Opaque q ': r) x
                  )
               -> Sem (e ': r) a
               -> Sem r a
```
This is _almost_ possible with our current architecture, but `weave` doesn't quite allow it.  Implementing this requires a rather massive overhaul over the library's internals, involving switching from the `weave` abstraction to something similar, but not quite. Remarkably, I think this is possible without introducing any breaking changes to the library (outside of internal modules).

I'm being vague about this because although I know it's possible, I haven't settled for what the best abstraction for the "something similar, but not quite" would be. The obvious option would be something similar to `MonadTransControl` (which the `weave` abstraction is essentially equivalent to). [I've sketched up a proof-of-concept based on a gist, but it's rather complicated.](https://gist.github.com/KingoftheHomeless/a79046a3e77d0660997423f21624dd6c)

PR #385, opened 2020-10-30
Move KVStore from polysemy-zoo to polysemy

locallycompact:
Hi. I'd like to move KVStore into the main repository. I find this very useful and use it quite heavily.


tek:
notably, it has no dependencies

googleson78:
Any opposition to this? @TheMatten @KingoftheHomeless 

It seems a bit too specific/non-traditional when compared with the other effects, but if people are often using it, I'm fine with including it.

Another idea is having a `polysemy-extra`, that has stuff such as this, but without the huge amount of dependencies that the zoo has (or perhaps just splitting the zoo so that it doesn't depend on `random`, `streaming`, etc)

thma:
I always write a KVStore from scratch to avoid the dependencies of polysemy-zoo.

So I think it's a useful addition to polysemy.

tek:
> Another idea is having a `polysemy-extra`, that has stuff such as this, but without the huge amount of dependencies that the zoo has (or perhaps just splitting the zoo so that it doesn't depend on `random`, `streaming`, etc)

is good idea

KingoftheHomeless:
`polysemy-extra` is something we've though about before, mainly for moving stuff from `polysemy-zoo`.
I agree; there's sufficient reason to add it. It doesn't have to start big; we can begin by porting the obvious useful effects like `KVStore` and `Fresh`, and go from there.

locallycompact:
I actually snagged that package name last week for my own nefarious purposes but I'm happy to migrate it here and share the love. http://hackage.haskell.org/package/polysemy-extra

tek:
hackage squatter!!

locallycompact:
I published kvstore under its own package here and updated polysemy-extra to depend on it and not the zoo.

https://hackage.haskell.org/package/polysemy-kvstore-0.1.2.0

isovector:
I'm quite open to this change! But this PR is empty and old, so I'm going to close it. Feel free to reopen!

Issue #386, opened 2020-10-30
Revamp `interpretH`


KingoftheHomeless:
`interpretH` is:
* __Very__ confusing to newcomers and experienced users alike
* Clunky to use even if you know what you're doing.
* Not quite powerful enough for some purposes

We need to fix it. The most conservative approach would be to implement an alternative `interpretH` -- something like #384 -- that addresses these issues; a more radical approach would be to replace `interpretH` entirely.

Either way, this will be part of the ever-expanding issues that v2 should cover. We need to do some work on this to figure out what the best alternate interface would be. [My proposal is the following:](https://github.com/polysemy-research/polysemy/pull/384#issuecomment-719587819) 
```haskell
interpretHGood :: (  forall z q x
                   . (forall x. z y -> Sem (Opaque q ': r) y)
                  -> e z x
                  -> Sem (Opaque q ': r) x
                  )
               -> Sem (e ': r) a
               -> Sem r a
```
but that implementation requires an extensive rework of the library's internals. This also shouldn't replace `interpretH` entirely, because there are some power `interpretH` gains from access to the functorial state that this doesn't have.

For v2, we should _at the very least_ implement a simple stop-gap like #384. If we're able to come up with a satisfying replacement for `interpretH` in time for v2, we'll include it; otherwise, we'll have it as a long-term goal.

tek:
another approach @googleson78 and I talked about was to stick the machinery from #384 into `Tactics`, like the inspector is. do you see a problem with that, or would you consider it a worse solution?

KingoftheHomeless:
I'm in favour of it. #384 relies on running the interpreter recursively automatically, which `Tactics` doesn't currently support. We should be able to extend it to be able to, though.
What should the naming schema be? `runTSimple` and `bindTSimple`?

KingoftheHomeless:
There's also the question about how such an extension would interact with #239, but my opinion is that we cross that bridge when we come to it. At the very worst, we have to give up unification of `Strategy` and `Tactics`, which isn't ideal, but isn't that bad either.

tek:
I think the names should be something like `interpret(Nested|HigherOrder|Embedded)`, reflecting that the current interpreter is run as well as that the thunk's Sem has a different `r` than the result type.
Of course those names are a mouthful and I'm not sure what terms are really intuitive…

PR #393, opened 2020-11-18
Simple variants of runT and bindT

KingoftheHomeless:
A modified variant of @tek's [simple `runT` and `bindT` implementation](https://github.com/tek/polysemy/commit/36325a2027c66de5eec8b1b9548eb3bd992f8645). In comparison, this version:

- Doesn't require any extra type variables
- Allows the new combinators to also be used in `reinterpretHn`, and not just in `interpretH`
- Calls the combinators `runTH` and `bindTH` instead of `interpretHO` and `interpretBindHO`.

This partially adresses #386.

These names of `runTH` and `bindTH` still aren't final unless we want them to be.

### DIFF
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -172,6 +172,7 @@ reinterpretH f sem = Sem $ \k -> runSem sem $ \u ->
       fmap y $ usingSem k
              $ runTactics s (raiseUnder . d) v (reinterpretH f . d)
              $ f e
+{-# INLINE[3] reinterpretH #-}
 -- TODO(sandy): Make this fuse in with 'stateful' directly.
 
 
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -80,12 +80,11 @@ interpretH
        -- already in 'Sem'.
     -> Sem (e ': r) a
     -> Sem r a
-interpretH f (Sem m) = m $ \u ->
+interpretH f (Sem m) = Sem $ \k -> m $ \u ->
   case decomp u of
-    Left  x -> liftSem $ hoist (interpretH f) x
+    Left  x -> k $ hoist (interpretH f) x
     Right (Weaving e s d y v) -> do
-      a <- runTactics s d v $ f e
-      pure $ y a
+      fmap y $ usingSem k $ runTactics s d v (interpretH f . d) $ f e
 {-# INLINE interpretH #-}
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


isovector:
`TH` sounds like template haskell to me. What is the `H` supposed to suggest?

src/Polysemy/Internal/Combinators.hs
```diff
@@ -166,13 +165,13 @@ reinterpretH
...
+reinterpretH f sem = Sem $ \k -> runSem sem $ \u ->
   case decompCoerce u of
     Left x  -> k $ hoist (reinterpretH f) $ x
     Right (Weaving e s d y v) -> do
-      a <- usingSem k $ runTactics s (raiseUnder . d) v $ f e
-      pure $ y a
-{-# INLINE[3] reinterpretH #-}
+      fmap y $ usingSem k
+             $ runTactics s (raiseUnder . d) v (reinterpretH f . d)
+             $ f e
```
tek, line 174: did you purposefully remove that inline pragma?

src/Polysemy/Internal/Combinators.hs
```diff
@@ -80,12 +80,11 @@ interpretH
        -- already in 'Sem'.
     -> Sem (e ': r) a
     -> Sem r a
-interpretH f (Sem m) = m $ \u ->
+interpretH f (Sem m) = Sem $ \k -> m $ \u ->
```
tek, line 83: just out of curiosity, why did you change that?

KingoftheHomeless:
"Higher". It's a kind of reference that the corresponding action that would be returned by `runT` is interpreted recursively and run immediately.
... Yeah, the names aren't stellar. As @tek 's pointed out, we'd ideally express it as `monadicallySequenceEmbeddedHigherOrderThunkAndRunCurrentInterpreterOnIt`, but that's just a tad too verbose.

src/Polysemy/Internal/Combinators.hs
KingoftheHomeless: No, I've since readded that back in.

src/Polysemy/Internal/Combinators.hs
KingoftheHomeless: Running a `Sem` as the universally quantified monad `m` inside of another `Sem` has always been more efficient than running a `Sem` as a `Sem` directly, so I figured I'd do that optimization while I was at it.

tek:
> ... Yeah, the names aren't stellar. As @tek 's pointed out, we'd ideally express it as `monadicallySequenceEmbeddedHigherOrderThunkAndRunCurrentInterpreterOnIt`, but that's just a tad too verbose.

looking at the doc, we can even replace `Thunk` with `Continuation` to get full Bingo

src/Polysemy/Internal/Combinators.hs
googleson78: so adding more stuff makes it go faster? :grin:

tek (approved):

KingoftheHomeless:
We decided to come full circle and name them `runTSimple` and `bindTSimple` as per the original suggestions in #386.

PR #394, opened 2020-11-18
Update docs for runTH and bindTH

googleson78:


src/Polysemy/Internal/Tactics.hs
```diff
@@ -38,7 +38,27 @@ import Polysemy.Internal.Union
...
+-- 'Polysemy.Resource.Bracket' alloc dealloc use -> do
+--   a <- 'runTH' alloc
+--   res <- 'bindTH' use a
+--   'bindTH' dealloc a
+--   pure res
+-- @
+-- To run the embedded higher-order actions we use 'runTH' for actions
+-- (e.g. @m a@) and 'bindTH' for kleisli actions (e.g. @a -> m b@).
+-- These functions will run the higher-order actions with
+-- \"the current interpreters\" (TODO good explanation on what this means???)
```
googleson78, line 52: ?

src/Polysemy/Internal/Tactics.hs
```diff
@@ -38,7 +38,27 @@ import Polysemy.Internal.Union
...
+--
+-- If you want more control over how other potentially embedded
+-- 'Polysemy.Resource.Resource's are run inside your interpeter,
+-- you might want to take a look at 'pureT', 'runT' and 'bindT'.
+-- An example of when you might need them, is when you are interpreting
+-- 'Polysemy.Reader.Local', in which you want to run inner actions
+-- with a modified environment.
 --
+-- TODO modify the section below here?? maybe add the Reader example?
```
googleson78, line 61: ?

src/Polysemy/Internal/Tactics.hs
googleson78: Just remove this, it's probably what users will naturally assume

src/Polysemy/Internal/Tactics.hs
googleson78: 0. add the reader example to show off when `runT` is useful
1. drop the currently existing section with `Resource`, as it is "subsumed" by the `runTH` and `bindTH` variants, and it doesn't show off the added power of `runT` and `bindT`

src/Polysemy/Internal/Tactics.hs
```diff
@@ -38,41 +40,36 @@ import Polysemy.Internal.Union
...
--- dealloc' :: f a1 -> 'Polysemy.Sem' ('Polysemy.Resource.Resource' ': r) (f ())
--- use'     :: f a1 -> 'Polysemy.Sem' ('Polysemy.Resource.Resource' ': r) (f x)
+-- 'Polysemy.Reader.runReader' i = 'Polysemy.interpretH' $ \case
+--   'Polysemy.Reader.Ask' -> pureT i
+--   'Polysemy.Reader.Local' f m -> do
+--     mm <- 'runT' m
+--     'Polysemy.raise' $ 'Polysemy.Reader.runReader' (f i) mm
 -- @
---
--- The @f@ type here is existential and corresponds to "whatever
```
googleson78, line 61: Unfortunately there is no long an example to attach this talking point to, and it seems rather important, as it explains the that a `forall f. Functor f => f ...` is being used as "state".

Is there some other example that can be given here to demonstrate this? @KingoftheHomeless @tek

googleson78:
Here's a screen of the rendered thing
2020-11-19-003102_1459x888_scrot

src/Polysemy/Internal/Tactics.hs
```diff
@@ -38,41 +41,48 @@ import Polysemy.Internal.Union
...
+-- Instead, we need to use 'runT' and 'bindT', which give more control over how inner actions are run.
+--
+-- As an example, let's look at the implementation for running a 'Polysemy.Reader.Reader'
+-- with a constant value:
+--
+-- @
+-- 'Polysemy.Reader.runReader' i = 'Polysemy.interpretH' $ \\case
+--   'Polysemy.Reader.Ask' -> pureT i
+--   'Polysemy.Reader.Local' f m -> do
+--     mm <- 'runT' m
```
tek, line 77: This `<` needs to be escaped

src/Polysemy/Internal/Tactics.hs
tek: can't we just put it at the end of the doc comment, so that the association with the type decl is apparent?

src/Polysemy/Internal/Tactics.hs
googleson78: Why? Seems to render find, or am I missing something? I'm not too good at haddock!

tek: oh yeah. weird, I remember that being a problem

src/Polysemy/Internal/Tactics.hs
tek: ok obviously that should be "top of the comment" :smile: 
in any case, the first sentence mentions "threading the state", so it might be sensible to just expand on that there

src/Polysemy/Internal/Tactics.hs
```diff
@@ -28,8 +28,15 @@ import Polysemy.Internal.Union
 -- effect @Tactics@, which is capable of rewriting monadic actions so they run
 -- in the correct stateful environment.
 --
--- Inside a 'Tactical', you're capable of running 'pureT', 'runT' and 'bindT'
--- which are the main tools for rewriting monadic stateful environments.
+-- The @f@ type here is existential and corresponds to "whatever
```
googleson78, line 31: I added the "existential" comment back here.

src/Polysemy/Internal/Tactics.hs
```diff
@@ -38,41 +45,53 @@ import Polysemy.Internal.Union
...
--- alloc'   ::         'Polysemy.Sem' ('Polysemy.Resource.Resource' ': r) (f a1)
--- dealloc' :: f a1 -> 'Polysemy.Sem' ('Polysemy.Resource.Resource' ': r) (f ())
--- use'     :: f a1 -> 'Polysemy.Sem' ('Polysemy.Resource.Resource' ': r) (f x)
--- @
+-- If we have a kleisli action (e.g. @use :: a -> m b@) we can instead use 'bindTSimple'.
 --
--- The @f@ type here is existential and corresponds to "whatever
--- state the other effects want to keep track of." @f@ is always
--- a 'Functor'.
+-- Note that because of the types of @'bindTSimple' use resource@ and @'bindTSimple' dealloc resource@
```
googleson78, line 63: I also add this "same environment" comment back here.

src/Polysemy/Internal/Tactics.hs
googleson78: Added it back near the top + another piece of documentation that I previously deleted

src/Polysemy/Internal/Tactics.hs
googleson78: Haddock rendering doesn't do a great job of distinguishing `@@` blocks from normal text..

tek (approved):

src/Polysemy/Internal/Tactics.hs
```diff
@@ -38,41 +45,53 @@ import Polysemy.Internal.Union
...
 -- @
--- 'Polysemy.Resource.Bracket' alloc dealloc use -> do
---   alloc'   <- 'runT'  alloc
---   dealloc' <- 'bindT' dealloc
---   use'     <- 'bindT' use
+-- runResource = Polysemy.interpretH $ \\case
+--  'Polysemy.Resource.Bracket' alloc dealloc use -> do
+--    resource <- 'runTSimple' alloc
+--    result <- 'bindTSimple' use resource
+--    'bindTSimple' dealloc resource
```
KingoftheHomeless, line 55: This is wrong, since this will have `dealloc` inherit the state from `alloc` and not `result`.
The "proper" would be:
```haskell
  fresource <- runTSimple alloc
  fresourceresult <- bindTSimple (\a -> fmap ((,) a) <$> use a) fresource
  bindTSimple (\(resource, result) ->  result <$ dealloc resource) fresourceresult
```
This is very complicated. Also, this documentation is misleading since this isn't actually how `runResource` is implemented; it does some extra stuff like using the inspector to check if `use` failed.

All in all, we should pick a different example than `Resource`.

KingoftheHomeless (changes requested):
Looks pretty good, but we need something better than the `runResource` example. Not sure what we could pick.

src/Polysemy/Internal/Tactics.hs
googleson78: Hm, does the current example do the same thing regarding the "state inheritance"? This is the impression I was left with (and it also seems to be the point of the example), from the current documentation.

I guess it would be misleading if the actual `runResource` runs in a different way.

isovector:
Subsumed by #422 (since all of these functions are going bye-bye.) Closing, but feel free to open if you feel like it!

PR #396, opened 2020-12-17
`seqEitherT` and `seqMaybeT`

tek:
Convenience combinators for the case where a higher-order interpreter
has to deal with `Either e (f a)`.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -39,6 +39,7 @@ custom-setup:
     - cabal-doctest >=1.0.6 && <1.1
 
 default-extensions:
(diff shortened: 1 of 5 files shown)


KingoftheHomeless:
I'm not sure. Combinators like these will prove to become redundant if we decide to pull the trigger and do the major rework described in https://github.com/polysemy-research/polysemy/pull/384#issuecomment-719587819, and I'm very tempted to do so.

... In fact, it's about time for me to start working on that. You can probably expect some draft in time for christmas.

PR #397, opened 2020-12-18
Core rewrite for new family of easy higher-order interpreters

### LINKED ISSUE #367: Can the Tactical Functor be Traversable?
I've been playing around with Polysemy recently, and have had some reasonable success with first order interpreters--they really deliver on "write the code the way you wish you could."

I recently tried creating some higher-order interpreters and found it (maybe unsurprisingly) much more challenging.  Should these even be really expected within application code, or are these really reserved for things like Resource and Error?

I've started to get my head around the types, but I seems like certain cases with additional actions from constraints are especially challenging.  `runT` and `bindT` work together to allow chaining the functor through their computations, but if you try to use an effect _not_ in a parameter, the arbitrary Functor requirement gets a bit awkward.  In certain cases it seems like you _must_ inspect the state or you end up stuck.

My thought was, that if the Functor was instead Traversable, you could use that to "upgrade" the action in the same way that you use `bindT` to upgrade the action in the parameter.  As an example:

```
higher = (Member Other r) => Sem (Higher ': r) a -> Sem r a
higher = interpretH \case
  Higher paramA paramB -> do
    paramA' <- runT paramA
    a <- paramA'
    paramB' <- bindT paramB'
    b <- paramB'
    traverse other b
```

This essentially lets us perform the `other` action inside the (now Traversable) Functor, but then flip the contexts to end up with the correct type.  Ultimately, we're saved from otherwise having to inspect the state.

Is Traversable a reasonable constraint to get the ergonomics here?  What actual Functors are used under the hood?  I was assuming that while Maybe is used for inspecting, it's not actually ever used as `f`.

Also, if I'm simply thinking about this problem incorrectly, I'd much appreciate a push in the correct direction.  Thanks!

### LINKED ISSUE #386: Revamp `interpretH`

KingoftheHomeless:
This is the massive rework mentioned in #386 and https://github.com/polysemy-research/polysemy/pull/384#issuecomment-719587819.

This essentially switches `polysemy` from a `weave`-based to a `MonadTransControl`-based effect system, and miraculously does so without _any_ breakage to any existing interface (outside of internal modules). This rewrite is done in order to support a new family of higher-order interpreter combinators -- currently called `interpretNew` -- which let you run higher-order thunks through a simple embedding action `runH` -- no state threading required!

Note, though, that `MonadTransControl` is more or less equivalent to `weave`, and this system still inherits all of `weave`'s problems (_except_ the difficulty of writing higher-order interpreters). So although `polysemy` wouldn't _technically_ be `weave`-based any longer, I'd still be calling it `weave`-based since the difference is so small to the point of being irrelevant.

This fixes #386.

Part of the rewrite is strengthening the `Tactical` environment such that the effectful state is represented by a `Traversable` instead of a `Functor`. This is because you can expect the state of any well-behaved `MonadTransControl` to be `Traversable` (at the very least, every real-world example I've encountered is.) So this also fixes #367.

### DIFF
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -107,6 +107,13 @@ module Polysemy
   , transform
 
     -- * Combinators for Interpreting Higher-Order Effects
+  , interpretNew
+  , interceptNew
+  , reinterpretNew
+  , reinterpret2New
+  , reinterpret3New
+
+    -- * Combinators for Interpreting Higher-Order Effects using the 'Tactical' enviroment
   , interpretH
   , interceptH
   , reinterpretH
@@ -123,6 +130,14 @@ module Polysemy
     -- * Composing IO-based Interpreters
   , (.@)
   , (.@@)
+
+    -- * 'RunH'
+    -- | When interpreting higher-order effects using 'interpretNew'
+    -- and friends, you can't execute higher-order "thunks" given my
+    -- the interpreted effect directly. Instead, these must be executed
+    -- using 'runH'.
+  , RunH
+  , runH
 
     -- * Tactics
     -- | Higher-order effects need to explicitly thread /other effects'/ state
@@ -143,6 +158,7 @@ module Polysemy
   , bindT
   , getInspectorT
   , Inspector (..)
+
   ) where
 
 import Polysemy.Final
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -133,7 +133,7 @@ module Polysemy
 
     -- * 'RunH'
     -- | When interpreting higher-order effects using 'interpretNew'
-    -- and friends, you can't execute higher-order "thunks" given my
+    -- and friends, you can't execute higher-order "thunks" given by
     -- the interpreted effect directly. Instead, these must be executed
     -- using 'runH'.
   , RunH
--- a/src/Polysemy.hs
+++ b/src/Polysemy.hs
@@ -133,7 +133,7 @@ module Polysemy
 
     -- * 'RunH'
     -- | When interpreting higher-order effects using 'interpretNew'
-    -- and friends, you can't execute higher-order "thunks" given by
+    -- and friends, you can't execute higher-order \"thunks\" given by
     -- the interpreted effect directly. Instead, these must be executed
     -- using 'runH'.
   , RunH
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -498,7 +498,7 @@ interpretNew h (Sem sem) = Sem $ \(k :: forall x. Union r (Sem r) x -> m x) ->
         in
           fmap ex $ lwr $ go1 (h e)
 
--- TODO (KingoftheHomeless): If performance matter, optimize the definitions
+-- TODO (KingoftheHomeless): If it matters, optimize the definitions
 -- below
 
 ------------------------------------------------------------------------------
--- a/src/Polysemy/Internal/Combinators.hs
+++ b/src/Polysemy/Internal/Combinators.hs
@@ -13,6 +13,16 @@ module Polysemy.Internal.Combinators
   , transform
 
     -- * Higher order
+  , RunH(..)
+  , runH
+
+  , interpretNew
+  , interceptNew
+  , reinterpretNew
+  , reinterpret2New
+  , reinterpret3New
+
+    -- * Higher order with 'Tactical'
   , interpretH
   , interceptH
   , reinterpretH
@@ -22,6 +32,7 @@ module Polysemy.Internal.Combinators
   -- * Conditional
   , interceptUsing
   , interceptUsingH
+  , interceptUsingNew
 
     -- * Statefulness
   , stateful
@@ -73,6 +84,10 @@ interpret = firstOrder interpretH
 -- | Like 'interpret', but for higher-order effects (ie. those which make use of
 -- the @m@ parameter.)
 --
+-- 'interpretNew' is /heavily recommended/ over this. Only use 'interpretH'
+-- if you need the additional power of the 'Tactical' environment -- that is,
+-- the ability to inspect and manipulate the underlying effectful state.
+--
 -- See the notes on 'Tactical' for how to use this function.
 interpretH
     :: (∀ x rInitial . e (Sem rInitial) x -> Tactical e (Sem rInitial) r x)
@@ -83,8 +98,10 @@ interpretH
 interpretH f (Sem m) = Sem $ \k -> m $ \u ->
   case decomp u of
     Left  x -> k $ hoist (interpretH f) x
-    Right (Weaving e s d y v) -> do
-      fmap y $ usingSem k $ runTactics s d v (interpretH f . d) $ f e
+    Right (Weaving e mkT lwr ex) -> do
+      let s = mkInitState lwr
+          Distrib d = mkDistrib mkT lwr
+      fmap ex $ usingSem k $ runTactics s d (interpretH f . d) $ f e
 {-# INLINE interpretH #-}
 
 ------------------------------------------------------------------------------
@@ -95,18 +112,16 @@ interpretInStateT
     -> s
     -> Sem (e ': r) a
     -> Sem r (s, a)
-interpretInStateT f s (Sem m) = Sem $ \k ->
-  (S.swap <$!>) $ flip S.runStateT s $ m $ \u ->
+interpretInStateT f s (Sem sem) = Sem $ \k ->
+  (S.swap <$!>) $ flip S.runStateT s $ sem $ \u ->
     case decomp u of
-        Left x -> S.StateT $ \s' ->
-              (S.swap <$!>)
-            . k
-            . weave (s', ())
-                    (uncurry $ interpretInStateT f)
-                    (Just . snd)
-            $ x
-        Right (Weaving e z _ y _) ->
-          y . (<$ z) <$> S.mapStateT (usingSem k) (f e)
+        Left x ->
+          liftHandlerWithNat
+            (\m -> S.StateT $ \s' -> swap <$!> interpretInStateT f s' m)
+            k x
+        Right (Weaving e _ lwr ex) -> do
+          let z = mkInitState lwr
+          ex . (<$ z) <$> S.mapStateT (usingSem k) (f e)
 {-# INLINE interpretInStateT #-}
 
 
@@ -118,17 +133,16 @@ interpretInLazyStateT
     -> s
     -> Sem (e ': r) a
     -> Sem r (s, a)
-interpretInLazyStateT f s (Sem m) = Sem $ \k ->
-  fmap swap $ flip LS.runStateT s $ m $ \u ->
+interpretInLazyStateT f s (Sem sem) = Sem $ \k ->
+  fmap swap $ flip LS.runStateT s $ sem $ \u ->
     case decomp u of
-        Left x -> LS.StateT $ \s' ->
-          k . fmap swap
-            . weave (s', ())
-                    (uncurry $ interpretInLazyStateT f)
-                    (Just . snd)
-            $ x
-        Right (Weaving e z _ y _) ->
-          y . (<$ z) <$> LS.mapStateT (usingSem k) (f e)
+        Left x ->
+          liftHandlerWithNat
+            (\m -> LS.StateT $ \s' -> swap <$> interpretInLazyStateT f s' m)
+            k x
+        Right (Weaving e _ lwr ex) -> do
+          let z = mkInitState lwr
+          ex . (<$ z) <$> LS.mapStateT (usingSem k) (f e)
 {-# INLINE interpretInLazyStateT #-}
 
 
@@ -157,6 +171,10 @@ lazilyStateful f = interpretInLazyStateT $ \e -> LS.StateT $ fmap swap . f e
 ------------------------------------------------------------------------------
 -- | Like 'reinterpret', but for higher-order effects.
 --
+-- 'reinterpretNew' is /heavily recommended/ over this. Only use 'reinterpretH'
+-- if you need the additional power of the 'Tactical' environment -- that is,
+-- the ability to inspect and manipulate the underlying effectful state.
+--
 -- See the notes on 'Tactical' for how to use this function.
 reinterpretH
     :: forall e1 e2 r a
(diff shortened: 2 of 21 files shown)


src/Polysemy.hs
```diff
@@ -124,6 +131,14 @@ module Polysemy
   , (.@)
   , (.@@)
 
+    -- * 'RunH'
+    -- | When interpreting higher-order effects using 'interpretNew'
+    -- and friends, you can't execute higher-order "thunks" given my
```
tek, line 136: typo: "my"

src/Polysemy/Internal/Combinators.hs
```diff
@@ -381,5 +424,151 @@ transform
...
+
+          go2 :: forall x. Sem (RunH z ': r) x -> t (Sem r) x
+          go2 = usingSem $ \u' -> case decomp (hoist go2 u') of
+            Right (Weaving (RunH z) _ lwr' ex') ->
+              (ex' . (<$ mkInitState lwr')) <$> mkT id z
+            Left g -> liftHandler liftSem g
+        in
+          fmap ex $ lwr $ go1 (h e)
+
+-- TODO (KingoftheHomeless): If performance matter, optimize the definitions
```
tek, line 501: typo: "matter"

src/Polysemy/Internal/Union.hs
```diff
@@ -73,59 +82,55 @@ instance Functor (Union r mWoven) where
...
-      --   Sem r (f x)@
-      , weaveResult :: f a -> resultType
-      -- ^ Even though @f a@ is the moral resulting type of 'Weaving', we
-      -- can't expose that fact; such a thing would prevent 'Polysemy.Sem'
-      -- from being a 'Monad'.
-      , weaveInspect :: forall x. f x -> Maybe x
-      -- ^ A function for attempting to see inside an @f@. This is no
-      -- guarantees that such a thing will succeed (for example,
-      -- 'Polysemy.Error.Error' might have 'Polysemy.Error.throw'n.)
+      , weaveTrans :: forall n x. Monad n => (forall y. mAfter y -> n y) -> Sem rInitial x -> t n x
```
KingoftheHomeless, line 92: This is `forall x. Sem rInitial x -> t m x`, but with `t m` Yoneda-encoded. This is to improve the efficiency of `hoist`; if you represent `t m` directly, then you would need to use `hoistT` for every use of `hoist`/`weave`, which is ineffecient for some `t`:s, like `NonDetC`.

src/Polysemy/Internal/Union.hs
```diff
@@ -73,59 +82,55 @@ instance Functor (Union r mWoven) where
...
+      , weaveLowering :: forall z x. Monad z => t z x -> z (StT t x)
```
KingoftheHomeless, line 93: This is the lowering function of a `liftWith`, and corresponds to running a `t z` under the initial state of the computation. So this corresponds to the initial state `f ()` under `weave`.

src/Polysemy/Internal/Union.hs
```diff
@@ -73,59 +82,55 @@ instance Functor (Union r mWoven) where
...
-      -- 'fmap' into this thing.
-      , weaveDistrib :: forall x. f (Sem rInitial x) -> mAfter (f x)
-      -- ^ Distribute @f@ by transforming @Sem rInitial@ into @mAfter@. This is
-      -- usually of the form @f ('Polysemy.Sem' (Some ': Effects ': r) x) ->
-      --   Sem r (f x)@
-      , weaveResult :: f a -> resultType
-      -- ^ Even though @f a@ is the moral resulting type of 'Weaving', we
-      -- can't expose that fact; such a thing would prevent 'Polysemy.Sem'
-      -- from being a 'Monad'.
-      , weaveInspect :: forall x. f x -> Maybe x
```
KingoftheHomeless, line 95: N.B. the inspector now implicitly exists by the fact that `StT t` is `Traversable`, so you can define:
```haskell
inspect :: MonadTransControl t => StT t a -> Maybe a
inspect = foldr (const . Just) Nothing
```

src/Polysemy/NonDet.hs
```diff
@@ -106,18 +103,24 @@ instance Monad (NonDetC m) where
...
   {-# INLINE (>>=) #-}
 
+instance MonadTrans NonDetC where
+  lift m = NonDetC $ \c b -> m >>= (`c` b)
+
+instance MonadTransControl NonDetC where
+  type StT NonDetC = []
+
+  hoistT n nd = NonDetC $ \c b ->
+    join $ n $ unNonDetC nd (\a r -> return $ c a (join (n r))) (return b)
```
KingoftheHomeless, line 113: Note: `NonDetC` is not a well-behaved `MonadTransControl` (due to #246), but this definition of `hoistT` _is_ well-behaved (morally, `hoistT id == id`). This gives rise to an inconsistency, where:
```haskell
controlT (\lower -> n (lower tm)) /= hoistT n tm
```

src/Polysemy/Internal/Combinators.hs
```diff
@@ -381,5 +424,151 @@ transform
...
-    Right (Weaving e s wv ex ins) ->
-      injWeaving (Weaving (f e) s wv ex ins)
+    Right (Weaving e mkT lwr ex) ->
+      injWeaving (Weaving (f e) mkT lwr ex)
+
+
+-- | An effect for running monadic actions within a higher-order effect
+-- currently being interpreted.
+newtype RunH z (m :: * -> *) a where
+  RunH :: z a -> RunH z m a
```
KingoftheHomeless, line 434: Why not just reuse `Embed`/`embed`? Since `z` is universally quantified in the handler passed to `interpretNew`, if the handler tries to use, say, `Embed IO`, then those uses will actually get stuck because `GHC` can't prove that `z` isn't `IO`. This can be fixed by having embedding of higher-order thunks be a completely separate effect.

Besides, I don't like overloading `Embed` that much. Having a separate effect for such a specific use-case as embedding higher-order thunks makes it easier for users to grok.

src/Polysemy/Internal/WeaveClass.hs
```diff
@@ -0,0 +1,159 @@
...
+  hoistT n (ComposeT m) = ComposeT (hoistT (hoistT n) m)
+
+  liftWith main = ComposeT $
+    liftWith $ \lowerT ->
+    liftWith $ \lowerU ->
+    main (\(ComposeT m) -> Compose <$> lowerU (lowerT m))
+
+  restoreT m = ComposeT (restoreT (restoreT (fmap getCompose m)))
+
+newtype Distrib f q m = Distrib (forall x. f (q x) -> m (f x))
```
KingoftheHomeless, line 84: Hack: lets you bind the distribution function in a `let` without the monomorphism restriction kicking in.

KingoftheHomeless:
Although this doesn't introduce any breaking changes, we may want to do so anyway. For example, `Final` can be revamped to have the following be the new `withWeavingToFinal`:
```haskell
withWeavingToFinal
  :: Member (Final m) r
  => (forall t. MonadTransControl t => (forall x. Sem r x -> t m x) -> t m a)
  -> Sem r a
```
However, I think such breaking changes should be done in a separate PR, since that's a much more difficult topic.

src/Polysemy.hs
```diff
@@ -124,6 +131,14 @@ module Polysemy
   , (.@)
   , (.@@)
 
+    -- * 'RunH'
+    -- | When interpreting higher-order effects using 'interpretNew'
+    -- and friends, you can't execute higher-order "thunks" given by
```
googleson78, line 136: Not sure if this `"asdf"` isn't supposed to refer to a module `asdf`, according to haddock markup?

KingoftheHomeless: Actually, from what I can tell haddock can tell `thunks` isn't a valid identifier for a module and won't try to convert it to a link.

... That said, I should escape it anyway.

src/Polysemy/Internal/WeaveClass.hs
```diff
@@ -0,0 +1,159 @@
...
+  liftWith :: Monad m
+           => ((forall z x. Monad z => t z x -> z (StT t x)) -> m a)
+           -> t m a
+
+  restoreT :: Monad m => m (StT t a) -> t m a
+
+controlT :: (MonadTransControl t, Monad m)
+         => ((forall z x. Monad z => t z x -> z (StT t x)) -> m (StT t a))
+         -> t m a
+controlT main = liftWith main >>= restoreT . pure
```
googleson78, line 54: `liftWith main <&> restoreT`?

src/Polysemy/Internal/Writer.hs
```diff
@@ -205,17 +205,11 @@ interpretViaLazyWriter f sem = Sem $ \(k :: forall x. Union r (Sem r) x -> m x)
...
-            g
-        return (a, o)
+      Right (Weaving e mkT lwr ex) -> f $ Weaving e (\n -> mkT (n . go)) lwr ex
+      Left g ->
+        liftHandlerWithNat
+          (Lazy.WriterT . fmap swap . interpretViaLazyWriter f)
+          k g
     {-# INLINE go #-}
-  in do
-    ~(a,s) <- Lazy.runWriterT (go sem)
```
googleson78, line 219: I think this was purposefully lazier than `swap` initially, for some performance reasons. I don't know if they are relevant (I think Sandy even had a specialised `swap` for this purpose at some point?)

src/Polysemy/Internal/Combinators.hs
```diff
@@ -381,5 +424,151 @@ transform
...
+
+-- | Run a monadic action given by a higher-order effect that is currently
+-- being interpreted.
+--
+-- @since TODO
+runH :: Member (RunH z) r => z a -> Sem r a
```
googleson78, line 440: Does this magically ensure that each following HO action you run "inherits" the state of the previous one, as opposed to the `*Simple` variants, where you're still threading it manually?

KingoftheHomeless: Yes.

src/Polysemy/Internal/Writer.hs
KingoftheHomeless: You're right on these being purposefully lazier swaps, but not as to why. [Sandy wrote these lazy swaps to try and fix `runFixpoint` diverging](https://github.com/polysemy-research/polysemy/pull/194#discussion_r308022711) before I wrote #187. The specialized swaps have since been removed, but this usage remained, because I was unsure if it mattered. I now know it doesn't: `Lazy.WriterT` `Monad` instance always constructs the resulting tuple lazily, so there's no scenario where `swap <$> Lazy.runWriterT (go sem)` would lead to divergence but `(\ ~(a,s) -> (s, a))<$> Lazy.runWriterT (go sem)` wouldn't (the latter being equivalent to the old definition).

src/Polysemy/Internal/Combinators.hs
```diff
@@ -381,5 +424,151 @@ transform
...
+-- runBind = 'interpretNew' \\case
+--   Bind ma f -> do
+--     a <- 'runH' ma
+--     b <- 'runH' (f a)
+--     return b
+-- @
+--
+-- @since TODO
+interpretNew :: forall e r a
+              . (forall z x. e z x -> Sem (RunH z ': r) x)
```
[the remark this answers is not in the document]
tek: would it be possible at all to switch the interpreter for higher order actions without `Tactics`?

src/Polysemy/Internal/Combinators.hs
tek: does this mean that where you would use `liftT` you now have to `raise` or do you expect that to be mostly inferred?

src/Polysemy/Internal/WeaveClass.hs
tek: filthy little monolocalbindses!

src/Polysemy/Internal/Combinators.hs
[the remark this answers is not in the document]
KingoftheHomeless: It is, but it'd require exposing the fact that there's a `MonadTransControl` involved, complicating the interface. It'd look something like this:
```haskell
data RunH z t e r m a where
  RunH :: z a -> RunH z t m a
  ProcessH :: z a -> RunH z t m (t (Sem (e ': r)) a)
  EmbedH :: t (Sem r) a -> RunH z t r m a 

interpretNew
  :: forall e r a
  => (forall z t x. MonadTransControl t => e z x -> Sem (RunH z t e r ': r) x)
  -> Sem r a -> Sem r a
```
then `runReader` could be implemented as follows:
```haskell
runReader i = interpretNew \case
  Ask -> return i
  Local f m -> do
    m' <- processH m
    embedH $ controlT $ \lower -> runReader (f i) (lower m')
```
Which is so complicated I'd prefer to save it for a rework of `Tactics`.

src/Polysemy/Internal/Combinators.hs
KingoftheHomeless: You'll need to use `raise` when working with `interpretNew` in two scenarios:

* You want to `send` a _completely polymorphic_ effect `e`, since then GHC can't tell it's not `RunH z`.
* You want to bring actions that are arguments to the interpreter into the environment.

So for example, if you want to implement `runOutputSem` using `interpretNew`:
```haskell
runOutputSem act = interpretNew \case
  Output o -> raise (act o)
```

Otherwise, you shouldn't ever need to `raise`. Indeed, most uses of `interpret` can be replaced with `interpretNew` and otherwise have the exact same definition and still work.

tek: awesome!

isovector:
Subsumed by #422. Closing, but feel free to open if you feel like it!

PR #400, opened 2021-01-24
add basic nix build

tek:
not quite sure what I should include in this – most of my stuff in other projects is specialized for my personal workflow.
I'll think about it some more; until then you can look and give me some inspiration if you have anything


nix/build.nix
```diff
@@ -0,0 +1,18 @@
...
+  nixpkgsSrc = niv.nixpkgs;
+
+  nixpkgs = import niv.nixpkgs {};
+  overrides = self: super: {
+    polysemy = self.callCabal2nix "polysemy" base {};
+    polysemy-plugin = self.callCabal2nix "polysemy-plugin" "${base}/polysemy-plugin" {};
+  };
+  ghc = nixpkgs.haskell.packages.ghc8102.override { inherit overrides; };
+in {
+  inherit ghc;
```
evanrelf, line 15: I would call this `haskellPackages` instead of `ghc`, because that's the set of Haskell packages for GHC 8.10.2, not the GHC package itself.

And with that in mind, I'm not sure what the value of `inherit ghc;` on line 15 is?

nix/build.nix
```diff
@@ -0,0 +1,18 @@
+{
+  base,
+}:
+let
+  niv = import ./sources.nix;
+  nixpkgsSrc = niv.nixpkgs;
```
evanrelf, line 6: `nixpkgsSrc` isn't used, so you could remove it

evanrelf (commented):
I'd also recommend adding a `shell.nix` file to make contributing easier. 

Something like this could be a good start:

```nix
# shell.nix

let
  sources = import ./sources.nix;

  pkgs = import sources.nixpkgs {};

  default = import ./default.nix;

in
  pkgs.mkShell {
    buildInputs = [
      pkgs.cabal-install
      default.polysemy
      default.polysemy-plugin
    ];
  }
```

nix/build.nix
tek: I named it `ghc` because I'm used to having both that and `ghcjs` from obelisk projects; in this case it doesn't make much sense, I agree.

The purpose of `inherit ghc` is to expose the overrides, for example for use in `shell.nix`. So far there's nothing in there, so it doesn't provide much utility, but I'd say it's still convenient for manual use.

nix/build.nix
tek: :+1:

nix/build.nix
evanrelf: Ah I see

README.md
```diff
@@ -228,6 +228,15 @@ You're going to want to stick all of this into your `package.yaml` file.
...
 ```
 
+## Building with Nix
+
+```bash
+nix-build -A polysemy
+nix-shell --pure --run cabal v2-haddock polysemy
+```
+
+TODO
```
googleson78, line 238: TODO

nix/sources.json
```diff
@@ -0,0 +1,26 @@
+{
```
googleson78, line 1: This file looks pretty copy-pasted, can't nix fetch it from somewhere itself?

nix/sources.nix
```diff
@@ -0,0 +1,138 @@
+# This file has been generated by Niv.
```
googleson78, line 1: Is it standard practice to commit generated files? Can't `Niv` regenerate this on the fly, or are there some custom things that it cannot infer alone?

tek: basically the `niv` workflow is orthogonal to the build. you need niv only to init the adapter file here and to update and add dependencies in `sources.json`. you don't need it for the build, and it would be problematic if you did, since it would need to be installed globally, making it impossible to build without extra instructions.

nix/sources.json
tek: this pins the dependencies, they are entirely specific to this project.

README.md
tek: DO!

evanrelf (approved):

README.md
```diff
@@ -228,6 +228,33 @@ You're going to want to stick all of this into your `package.yaml` file.
     - TypeFamilies
 ```
 
+## Building with Nix
+
+The project provides basic nix config for building in development.
```
googleson78, line 233:
```suggestion
The project provides a basic nix config for building in development.
```

README.md
```diff
@@ -228,6 +228,33 @@ You're going to want to stick all of this into your `package.yaml` file.
...
+nix-build -A polysemy-plugin
+```
+
+To inspect a dependency:
+
+```bash
+nix repl
+
+> p = import ./.
+> p.unagi-chan
```
googleson78, line 248: This is relevant to the project, because we have `unagi-chan` as a dependency - is this correct?

tek: yeah I just chose one at random to showcase that the nix derivation contains the whole build

evanrelf:
FYI `cachix` lets you configure how it adds the substituter:

```
$ cachix use --help
...
  -m,--mode ARG            Mode in which to configure binary caches for Nix.
                           Supported values: nixos, root-nixconf, user-nixconf
...
```

tek:
oh! thanks!
I guess `user-nixconf` should work best here, huh

evanrelf:
It may be easier to build for Nix with Cachix on GitHub Actions, if you're not too attached to Travis CI. Here's an example from one of my projects:

https://github.com/evanrelf/drama/blob/13504719fe617b272f9420f5923cee25311eb850/.github/workflows/ci.yml

22 lines to install Nix, set up Cachix, and run `nix-build`. Things are automatically pushed to Cachix.

tek:
I'll check it out!

tek:
finally got around to crafting those flakes!

tek:
alright, I guess that's about it. please review again folks! @evanrelf @googleson78 @TheMatten

flake.nix
```diff
@@ -0,0 +1,65 @@
...
+  outputs = { nixpkgs, flake-utils, ... }:
+  flake-utils.lib.eachSystem ["x86_64-linux"] (system:
+  let
+    overrides = self: super:
+    let
+      c2n = name: src: self.callCabal2nixWithOptions name src "--no-hpack" {};
+      hack = pkg: ver: sha256: self.callHackageDirect { inherit pkg ver sha256; } {};
+    in {
+      polysemy = c2n "polysemy" ./.;
+      polysemy-plugin = c2n "polysemy-plugin" ./polysemy-plugin;
```
evanrelf, line 18: I recommend replacing the raw paths you're using as source arguments (`./.` and `./polysemy-plugin`) with source that has had `.gitignore` taken into account. This will avoid unnecessary rebuilding.

e.g. replace `./.` with `pkgs.nix-gitignore.gitignoreSource [] ./.`.

You'll need to have Nixpkgs in scope (the `self` on line 25, not the `haskellPackages` `self` on line 12).

You can also add additional ignore patterns or ignore files to that list. In my projects, I like to have a `.nixignore` file used as a _whitelist_ instead of a blacklist. That way stuff like `README.md` or whatever doesn't change the hash for the Nix build.

evanrelf (approved):
Thanks for doing this! I didn't have time to review this _super_ thoroughly, but hopefully my comments help.

googleson78:
thanks a lot @evanrelf !

flake.nix
tek: that's a pretty neat feature, I'll surely use that for my other projects!

I'm not sure what exactly needs to be in the build result – if you're only running a dev build, it's clear, but since you can export overlays with flake, for example, I'd assume that you'd want all relevant files to be in the package.

Also, wouldn't it be simpler to directly use `filterSource` if you're going to use an include list?

README.md
```diff
@@ -220,6 +220,50 @@ You're going to want to stick all of this into your `package.yaml` file.
...
+
+The project provides a basic nix config for building in development.
+It is defined as a [flake] with backwards compatibility stubs in `default.nix` and `shell.nix`.
+
+To build the main library or plugin:
+
+```bash
+nix build
+nix build '.#polysemy-plugin'
+```
```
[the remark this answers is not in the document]
tek: :heavy_check_mark:

flake.nix
```diff
@@ -0,0 +1,65 @@
...
+
+    devShell = pkgs.stdenv.mkDerivation {
+      name = "polysemy-dev-shell";
+      buildInputs = with ghc; [
+        cabal-install
+        haskell-language-server
+        ghcid
+        (ghcWithPackages (g: g.polysemy.buildInputs))
+      ];
+    };
```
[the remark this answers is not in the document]
tek: :heavy_check_mark:

flake.nix
```diff
@@ -0,0 +1,65 @@
...
+        };
+      });
+    };
+
+    pkgs = import nixpkgs {
+      inherit system;
+      overlays = [overlay];
+    };
+
+    ghc = pkgs.haskell.packages.ghc884;
```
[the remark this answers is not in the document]
tek: :heavy_check_mark:

flake.nix
```diff
@@ -0,0 +1,65 @@
...
+    nixpkgs.url = github:NixOs/nixpkgs/nixos-20.09;
+    flake-utils.url = github:numtide/flake-utils;
+  };
+
+  outputs = { nixpkgs, flake-utils, ... }:
+  flake-utils.lib.eachSystem ["x86_64-linux"] (system:
+  let
+    overrides = self: super:
+    let
+      c2n = name: src: self.callCabal2nixWithOptions name src "--no-hpack" {};
```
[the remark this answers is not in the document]
tek: because we have `.cabal` checked in and it will be generated from different environments due to still using stack.
but now that I've added the source filter, we don't need it anymore!

default.nix
```diff
@@ -0,0 +1,7 @@
+let
+  compat = fetchTarball {
+    url = "https://github.com/edolstra/flake-compat/archive/99f1c2157fba4bfe6211a321fd0ee43199025dbf.tar.gz";
+    sha256 = "0x2jn3vrawwv9xp15674wjz9pixwjyj3j771izayl962zziivbx2";
+  };
+in
```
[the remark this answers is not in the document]
tek: :heavy_check_mark:

flake.nix
```diff
@@ -0,0 +1,65 @@
...
+      hack = pkg: ver: sha256: self.callHackageDirect { inherit pkg ver sha256; } {};
+    in {
+      polysemy = c2n "polysemy" ./.;
+      polysemy-plugin = c2n "polysemy-plugin" ./polysemy-plugin;
+      first-class-families = hack "first-class-families" "0.8.0.1" "0h1rxbc7zsxrlhx5xcl58wjx3qi2wny8wb3sk7c1qnydf4ckcckz";
+      ghc-lib = hack "ghc-lib" "0.20190204" "1i2zjlr30bxpabsm8f995x455v847gbpzczm2w4089fqmsnzpzfd";
+      dump-core = hack "dump-core" "0.1.3.2" "1mi8p736yn00z549pwnjv4ydwbs8mwg6dla3ly447c027nq8py6g";
+      monadLib = hack "monadLib" "3.10" "1v4ynjcb963s3lfw3v71qdzvld1mmz1faf8swhvicma5jbvwchy2";
+    };
```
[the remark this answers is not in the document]
tek: :heavy_check_mark:

.github/workflows/nix-build.yml
```diff
@@ -0,0 +1,18 @@
+name: Build Polysemy with Nix
+on:
+  push:
+    branches:
+      - nix-build
```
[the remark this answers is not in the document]
tek: after merge!

tek:
@evanrelf thanks! updated accordingly.

flake.nix
evanrelf: Even if you're uncertain what is safe to exclude from the source, it would still be useful to clean out `.gitignore`-d files. But it's not a correctness issue omitting it, so it's up to you.

evanrelf:
@googleson78 @tek Happy to help where I can 🙂

TheMatten (approved):
LGTM, thanks for flakes support! :slightly_smiling_face:

PR #406, opened 2021-03-30
add runAtomicStateLocal

tek:
While `AtomicState` and `State` can't be merged, at least it should be made more accessible and clear that it is a superstructure.

Naming?


isovector (approved):
LGTM

PR #409, opened 2021-04-09
Add traceToStdout and traceToStderr, deprecating traceToIO

### LINKED ISSUE #408: `traceToIO` with `stderr`?
Intended use of `Trace` is for logging but it's main interpreter `traceToIO` is hardcoded to output to `stdout` while more common place to put error/diagnostics/logs for unix programs is `stderr`. 

Not sure what's the right action here but, imo, in ideal world `traceToIO` should output to `stderr` by default.

tek:
fixes #408

### DIFF
--- a/src/Polysemy/Trace.hs
+++ b/src/Polysemy/Trace.hs
@@ -8,6 +8,9 @@ module Polysemy.Trace
   , trace
 
     -- * Interpretations
+  , traceToHandle
+  , traceToStdout
+  , traceToStderr
   , traceToIO
   , runTraceList
   , ignoreTrace
@@ -19,6 +22,7 @@ module Polysemy.Trace
 
 import Polysemy
 import Polysemy.Output
+import System.IO (stdout, stderr, hPutStrLn, Handle)
 
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/Polysemy/Trace.hs
```diff
@@ -29,14 +33,45 @@ data Trace m a where
...
+{-# INLINE traceToHandle #-}
+
+
+------------------------------------------------------------------------------
+-- | Run a 'Trace' effect by printing the messages to stdout.
+--
+-- @since TODO
+traceToStdout :: Member (Embed IO) r => Sem (Trace ': r) a -> Sem r a
+traceToStdout =
+  traceToHandle stdout
```
KingoftheHomeless, line 52: Any reason these aren't one-liners?

tek: you mean as in unfolding `traceToHandle`?

KingoftheHomeless: No, I mean
```
traceToStdout =traceToHandle stdout
```

tek: oh! no, I just always put a newline after the `=`. gonna fix

KingoftheHomeless (approved):

PR #410, opened 2021-04-16
start migration to ghc9

tek:
The lib only has two test failures regarding error messages, but in the plugin everything's broken.
All the modules from `ghc` seem to be invisible.

to build:

```
nix-build -L .#polysemy-901
nix-build -L .#polysemy-plugin-901

```


src/Polysemy/Internal.hs
```diff
@@ -267,7 +267,7 @@ instance Functor (Sem f) where
 
 
 instance Applicative (Sem f) where
-  pure a = Sem $ const $ pure a
+  pure a = Sem $ \_ -> pure a
```
KingoftheHomeless, line 270: ?

KingoftheHomeless (commented):
Didn't the GHC library receive a module structure rework? That would be it.

src/Polysemy/Internal.hs
KingoftheHomeless: Didn't mean for this comment to be submitted. There _are_ some minor changes here and there I'm not sure why you've made, but they shouldn't affect anything.

tek: All of those changes were prompted by GHC complaining. In this case, something about polytypes

tek:
```
    • Couldn't match type ‘b0’
                     with ‘forall x. Union f (Sem f) x -> m x’
      Expected: (forall x. Union f (Sem f) x -> m x) -> m a
        Actual: b0 -> m a
      Cannot instantiate unification variable ‘b0’
      with a type involving polytypes: forall x. Union f (Sem f) x -> m x
```

KingoftheHomeless: Oh! This is a side-effect of quick look impredicativity (or the loss of function contravariance?), I wager. In earlier versions, I bet `b0` was instantiated to `Union f (Sem f) x0 -> m x0` for some ambiguous `x0`, making it first-rank and thus valid. Now, `b0` is instead instantiated to the proper impredicative `(forall x. Union f (Sem f) x -> m x)`, which is rejected since quick look impredicativity isn't on.

tek: should we activate it conditionally?

tek: FYI works with `ImpredicativeTypes`

KingoftheHomeless: Are you running with GHC 9.2? Otherwise, quick-look isn't in yet.

If you are running with 9.2, why not. I expect `ImpredicativeTypes` to replace `RankNTypes` entirely.

tek: no, 9.0.1. Still fixes the error!

KingoftheHomeless: Ergh, actually, we want backwards compatibility, so I'd wait until GHC 9.6 before enabling it (since by then the "three most recent major versions" standard will not include any version prior to 9.2).

Avoid `ImpredicativeTypes` for the moment, unless you're willing to guard its use with CPP.

tek: that's what I meant with "conditionally". I'm also using CPP to compensate for the breaking change in `TH.Syntax`, so I would find it acceptable to be able to use the same code as before by using the extension

KingoftheHomeless: `ImpredicativeTypes` in 9.0.1 is still undefined. Don't use it before 9.2.

tek: ok, so we keep this change

KingoftheHomeless: Yup.

tek:
success! @KingoftheHomeless pls merge if you deem this sufficient

KingoftheHomeless:
The tests check out. Looks good enough to me.

tek:
wooo :partying_face:

felixonmars:
Not sure why but the tests are not passing for me:

```
Test suite polysemy-test: RUNNING...
test/TypeErrors.hs:81: failure in expression `:{
 let reinterpretScrub :: Sem (Output Int ': m) a -> Sem (State Bool ': Trace ': m) a
     reinterpretScrub = undefined
     foo :: Sem '[Output Int] ()
     foo = pure ()
     foo' = reinterpretScrub foo
     foo'' = runState True foo'
     foo''' = traceToIO foo''
  in runM foo'''
:}'
expected: ...
          ... Unhandled effect 'Embed IO'
          ...
          ... Expected type: Sem '[Embed m] (Bool, ())
          ... Actual type: Sem '[] (Bool, ())
          ...
 but got:
          <interactive>:160:15: error:
              • Unhandled effect 'Embed IO'
                Probable fix:
                  add an interpretation for 'Embed IO'
              • In the expression: traceToIO foo''
                In an equation for ‘foo'''’: foo''' = traceToIO foo''
                In the expression:
                  let
                    reinterpretScrub ::
                      Sem (Output Int : m) a -> Sem (State Bool : Trace : m) a
                    reinterpretScrub = undefined
                    foo :: Sem '[Output Int] ()
                    ....
                  in runM foo'''

          <interactive>:161:11: error:
                                ^
              • Couldn't match type: '[]
                               with: '[Embed m]
                Expected: Sem '[Embed m] (Bool, ())
                  Actual: Sem '[] (Bool, ())
              • In the first argument of ‘runM’, namely ‘foo'''’
                In the expression: runM foo'''
                In the expression:
                  let
                    reinterpretScrub ::
                      Sem (Output Int : m) a -> Sem (State Bool : Trace : m) a
                    reinterpretScrub = undefined
                    foo :: Sem '[Output Int] ()
                    ....
                  in runM foo'''
              • Relevant bindings include
                  it :: m (Bool, ()) (bound at <interactive>:154:2)

test/TypeErrors.hs:116: failure in expression `:{
 let foo :: Member Resource r => Sem r ()
     foo = undefined
  in runM $ lowerResource foo
:}'
expected: ...
          ... Couldn't match expected type ...
          ... with actual type ...
          ... Probable cause: ... is applied to too few arguments
          ...
 but got:
          <interactive>:208:13: error:
              • Couldn't match expected type: Sem '[Embed m] a
                            with actual type: Sem (Resource : r0) a0 -> Sem r0 a0
              • Probable cause: ‘lowerResource’ is applied to too few arguments
                In the second argument of ‘($)’, namely ‘lowerResource foo’
                In the expression: runM $ lowerResource foo
                In the expression:
                  let
                    foo :: Member Resource r => Sem r ()
                    foo = undefined
                  in runM $ lowerResource foo
              • Relevant bindings include
                  it :: m a (bound at <interactive>:206:2)

          <interactive>:208:27: error:
              • Couldn't match expected type: Sem r0 x -> IO x
          ..... Couldn't match expected type^
                            with actual type: Sem r1 ()
              • In the first argument of ‘lowerResource’, namely ‘foo’
                In the second argument of ‘($)’, namely ‘lowerResource foo’
                In the expression: runM $ lowerResource foo

Examples: 17  Tried: 17  Errors: 0  Failures: 2
```

Any ideas?

tek:
@felixonmars how are you running it?

tek:
pushed a fix!

felixonmars:
@tek Thanks a lot! (And sorry for the late response)

PR #411, opened 2021-04-20
Add combinator for adding effects at an index

tek:
opinions?

### DIFF
--- a/polysemy-plugin/test/InsertSpec.hs
+++ b/polysemy-plugin/test/InsertSpec.hs
@@ -1,0 +1,22 @@
+module InsertSpec where
+
+import Polysemy.State (State)
+import Polysemy.Reader (Reader)
+import Polysemy.Internal (insertAt)
+import Polysemy
+import Test.Hspec
+
+insertAtWithIndex :: Sem (e1 : e2 : e3 : e4 : r) a -> Sem (e1 : e2 : Reader i : e3 : e4 : r) a
+insertAtWithIndex = insertAt @2
+
+insertAtAndRaiseUnder :: Sem (e1 : r) a -> Sem (e1 : e2 : State s : e3 : Reader i : e4 : r) a
+insertAtAndRaiseUnder = raise2Under . insertAt @2 . raiseUnder
+
+insertAtEmpty :: Sem (e1 : e2 : r) a -> Sem (e1 : e2 : r) a
+insertAtEmpty = insertAt @2
+
+spec :: Spec
+spec = parallel $ do
+  describe "insert" $ do
+    it "insert" $ do
+      1 `shouldBe` 1
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -30,6 +30,7 @@ module Polysemy.Internal
   , Subsume (..)
   , subsume
   , subsumeUsing
+  , insertAt
   , Embed (..)
   , usingSem
   , liftSem
@@ -44,6 +45,8 @@ module Polysemy.Internal
 import Control.Applicative
 import Control.Monad
 #if __GLASGOW_HASKELL__ < 808
+
+
 import Control.Monad.Fail
 #endif
 import Control.Monad.Fix
@@ -53,10 +56,13 @@ import Data.Kind
 import Polysemy.Embed.Type
 import Polysemy.Fail.Type
 import Polysemy.Internal.Fixpoint
+import Polysemy.Internal.Index (InsertAtUnprovidedIndex, InsertAtIndex(insertAtIndex))
 import Polysemy.Internal.Kind
 import Polysemy.Internal.NonDet
 import Polysemy.Internal.PluginLookup
 import Polysemy.Internal.Union
+import Type.Errors (WhenStuck)
+import Polysemy.Internal.Sing (ListOfLength (listOfLength))
 
 
 -- $setup
@@ -540,6 +546,32 @@ subsumeUsing pr =
     go
 {-# INLINE subsumeUsing #-}
 
+------------------------------------------------------------------------------
+-- | Introduce a set of effects into 'Sem' at the index @i@, before the effect
+-- that previously occupied that position. This is intended to be used with a
+-- type application:
+--
+-- @
+-- let
+--   sem1 :: Sem [e1, e2, e3, e4, e5] a
+--   sem1 = insertAt @2 (sem0 :: Sem [e1, e2, e5] a)
+-- @
+--
+-- @since 1.6.0.0
+insertAt
+  :: forall index inserted head oldTail tail old full a
+   . ( ListOfLength index head
+     , WhenStuck index InsertAtUnprovidedIndex
+     , old ~ Append head oldTail
+     , tail ~ Append inserted oldTail
+     , full ~ Append head tail
+     , InsertAtIndex index head tail oldTail full inserted)
+  => Sem old a
+  -> Sem full a
+insertAt = hoistSem $ \u -> hoist (insertAt @index @inserted @head @oldTail) $
+  weakenMid @oldTail (listOfLength @index @head) (insertAtIndex @Effect @index @head @tail @oldTail @full @inserted) u
+{-# INLINE insertAt #-}
+
 
 ------------------------------------------------------------------------------
 -- | Embed an effect into a 'Sem'. This is used primarily via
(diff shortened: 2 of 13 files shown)


googleson78:
I wonder if you can't somehow reuse `Here/There` for this (?) :thinking:

src/Polysemy/Internal.hs
```diff
@@ -534,6 +540,143 @@ subsumeUsing pr =
...
+class KnownList l where
+  singList :: SList l
+
+instance KnownList '[] where
+  singList = SEnd
+  {-# INLINE singList #-}
+
+instance KnownList xs => KnownList (x ': xs) where
+  singList = SCons singList
+  {-# INLINE singList #-}
```
KingoftheHomeless, line 614: We should put all of the `SList` and `KnownList` boilerplate in a different module; I suggest `Polysemy.Internal.Membership`.
Also, replace the existing `KnownList` representation in `Polysemy.Internal.Bundle` with usages of this new one, instead.

src/Polysemy/Internal.hs
```diff
@@ -534,6 +540,143 @@ subsumeUsing pr =
...
+instance {-# INCOHERENT #-} (Build l1 s r ix head return, l ~ (e ': l1))
+      => Build l (e ': s) r ix head return where
+  singBuild = SCons (singBuild @_ @l1 @s @r @ix @head @return)
+
+instance {-# INCOHERENT #-} TypeError (InsertAtFailure r ix head return)
+       => Build l s r ix head return where
+  singBuild = undefined
+
+type family InsertAtUnprovidedIndex where
+  InsertAtUnprovidedIndex = TypeError ('Text "insertAt: index must be provided")
```
KingoftheHomeless, line 640: `Tide` and `Build` were placeholder names, and should be replaced. My suggestions are`KnownListOfLength` and `KnownPrefix`.

`Tide` isn't `insertAt`-specific, and should reside in `Polysemy.Internal.Membership`. A silly thing here is that the only thing that makes `Build` `insertAt` specific is the need to propagate information for error messages. I suppose that can't be helped.

src/Polysemy/Internal.hs
```diff
@@ -534,6 +540,143 @@ subsumeUsing pr =
...
+insertAt
+  :: forall n new top r s a
+   . ( Tide n top
+     , WhenStuck n InsertAtUnprovidedIndex
+     , s ~ Append new r, Build new s r n top (Append top s)
+     , Subsume (Append top r) (Append top s))
+  => Sem (Append top r) a
+  -> Sem (Append top s) a
+insertAt =
+  subsume_
```
KingoftheHomeless, line 663: Remember that `subsume_` won't work here. You need to steal [`weakenMid`](https://hackage.haskell.org/package/in-other-words-0.2.0.0/docs/src/Control.Effect.Internal.KnownList.html#weakenMid) from IOW and use that, instead.

Implementation should be something like this (haven't doublechecked):
```
insertAt = hoistSem $ \u -> hoist (insertAt @n @new @top @r @s) $
  weakenMid @r (singTide @n @top) (singBuild @_ @new @s @r @n @top @(Append top s)) u
```
Some type applications may be redundant.

src/Polysemy/Internal.hs
```diff
@@ -534,6 +540,143 @@ subsumeUsing pr =
...
+--
+-- @since TODO
+insertBefore ::
+  ∀ e es r r' a .
+  r' ~ InsertedBefore e es r =>
+  Subsume r r' =>
+  Sem r a ->
+  Sem r' a
+insertBefore = subsume_
+{-# inline insertBefore #-}
```
KingoftheHomeless, line 597: Given the problems I pointed out with `insertBefore` and `insertAfter`, my opinion remains that we shouldn't include them.

polysemy-plugin/test/InsertSpec.hs
```diff
@@ -0,0 +1,31 @@
...
+
+import Polysemy.Reader (Reader)
+import Polysemy.State (State)
+import Polysemy.Internal (Sem)
+import Polysemy.Internal (insertAt)
+import Polysemy
+import Test.Hspec
+
+insertSome :: ∀ e1 e2 i r a . Sem (e1 ': e2 ': r) a -> Sem (e1 ': e2 ': Reader i ': r) a
+insertSome = insertAt @2 @'[Reader i]
```
KingoftheHomeless, line 11: Drop `@'[Reader i]`, it shouldn't be necessary.

polysemy-plugin/test/InsertSpec.hs
```diff
@@ -0,0 +1,31 @@
...
+--   Sem (e1 ': e2 ': r) a ->
+--   Sem (e1 ': e2 ': State Int ': e4 ': Reader i ': r) a
+-- insertAtTest2 = insertAt @2 @'[State Int, e4, Reader i]
+
+
+spec :: Spec
+spec = parallel $ do
+  describe "insert" $ do
+    it "insert" $ do
+      1 `shouldBe` 1
```
KingoftheHomeless, line 31: If we can bear the _headache_, we should have tests for compile-time error messages like the `polysemy-plugin` tests do.

src/Polysemy/Internal.hs
KingoftheHomeless: That is, `insertBefore` and `insertAfter` plays hell with even partially polymorphic effects. This doesn't work:
```
insertAfterTest :: forall i r a. Sem (Reader Int ': Reader i ': r) a -> Sem (Reader Int ': Reader i ': State Int ':  r) a
insertAfterTest = insertAfter @(Reader i) @'[State Int]
```
This is unavoidable as long as we try to use effect equality in order to determine where the new effects should be inserted, making `insertAfter` and `insertBefore` fundamentally flawed.

src/Polysemy/Internal.hs
```diff
@@ -534,6 +540,143 @@ subsumeUsing pr =
...
+--   sem1 = insertAt @2 @[e3, e4] (sem0 :: Sem [e1, e2, e5] a)
+-- @
+--
+-- @since TODO
+insertAt
+  :: forall n new top r s a
+   . ( Tide n top
+     , WhenStuck n InsertAtUnprovidedIndex
+     , s ~ Append new r, Build new s r n top (Append top s)
+     , Subsume (Append top r) (Append top s))
```
KingoftheHomeless, line 659: No `Subsume`. Begone.

KingoftheHomeless (commented):
I realize this is all just placeholder for the snippets of code I threw out, but I figured I could place some signposts as to how we should proceed from here.

src/Polysemy/Internal.hs
```diff
@@ -342,7 +352,7 @@ hoistSem nat (Sem m) = Sem $ \k -> m $ \u -> k $ nat u
 
 
 ------------------------------------------------------------------------------
--- | Introduce an arbitrary number of effects on top of the effect stack. This
+-- | Introduce an arbitrary number of effects on head of the effect stack. This
```
KingoftheHomeless, line 355: the head

src/Polysemy/Internal.hs
```diff
@@ -464,7 +474,7 @@ raise3Under = hoistSem $ hoist raise3Under . weaken3Under
 
 
 ------------------------------------------------------------------------------
--- | Allows reordering and adding known effects on top of the effect stack, as
+-- | Allows reordering and adding known effects on head of the effect stack, as
```
KingoftheHomeless, line 477: the head

src/Polysemy/Internal.hs
```diff
@@ -540,6 +550,32 @@ subsumeUsing pr =
...
+--
+-- @since TODO
+insertAt
+  :: forall index inserted head oldTail tail old full a
+   . ( ListOfLength index head
+     , WhenStuck index InsertAtUnprovidedIndex
+     , old ~ Append head oldTail
+     , tail ~ Append inserted oldTail
+     , full ~ Append head tail
+     , InsertAtIndex index head tail oldTail full inserted)
```
KingoftheHomeless, line 572: Good lord... I know there's not much we can do to shorten this type signature, but still.

KingoftheHomeless (approved):
Finish the rewrite on `Bundle` and this looks good.

src/Polysemy/Internal.hs
tek: well those are basically local type binders…and an error message. Effectively it's just two constraints :upside_down_face:

src/Polysemy/Internal.hs
tek: whoops, that was a _replace all_ :sweat_smile:

tek: changing it back to the original wording

tek: maybe _stack_ is a misleading term, given that we're inserting effects in the middle of the list

PR #412, opened 2021-04-23
Add `Semigroup` and `Monoid` instances for `Sem`

Gabriel439:
I don't feel strongly about this change, but I frequently use these lifted `Monoid` and `Semigroup` instances in my own coding

I also wasn't sure what `@since` annotations to add since I don't know what the next release number will be.

### DIFF
--- a/src/Polysemy/Internal.hs
+++ b/src/Polysemy/Internal.hs
@@ -307,6 +307,11 @@ instance (Member Fail r) => MonadFail (Sem r) where
   fail = send . Fail
   {-# INLINE fail #-}
 
+instance Semigroup a => Semigroup (Sem f a) where
+  (<>) = liftA2 (<>)
+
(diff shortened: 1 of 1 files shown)


tek:
we use `@since TODO`!

tek (approved):

KingoftheHomeless (approved):
I honestly thought we already had these.

googleson78:
Hmm, should they have inline pragmas? Do those even work for instance methods?

lambdadog:
@googleson78 my understanding is that they shouldn't (unless GHC has reason to specialize the instance which, admittedly my memory isn't entirely clear on exactly when GHC chooses to take this step), since instance methods are essentially, implementation-wise, passed in as records to the functions calling them

although this is all based on my shaky memory of Alexis King's "Effects for Less" talk at ZuriHac 2020.

PR #414, opened 2021-07-10
remove type-errors-pretty to speed up release for 901

tek:


src/Polysemy/Internal/CustomErrors.hs
```diff
@@ -36,6 +37,19 @@ type family DefiningModuleForEffect (e :: k) :: Symbol where
   DefiningModuleForEffect (e a) = DefiningModuleForEffect e
   DefiningModuleForEffect e     = DefiningModule e
 
+-- These are taken from type-errors-pretty
```
googleson78, line 40: sounds good to me, just one thing - let's add a link to https://github.com/polysemy-research/polysemy/issues/401 so that it's documented why we pulled this in here.

tek: brilliant!

PR #424, opened 2021-10-20
Plugin: Solve ambiguous types based on instances in scope

### LINKED ISSUE #219: Plugin doesn't attempt to disambiguate two instances of the same effect with different type variables
```haskell
ambiguous :: Members '[
              State Int
            , State String
            ] r
          => Sem r ()
ambiguous = put 10
```

Leads to
```
Ambiguous use of effect 'State'
Possible fix:
  add (Member (State Integer) r) to the context of
        the type signature
```

This throws a wrench into my plans to add a `Tagged` effect for use to disambiguate (truly) identical effects, so I'll look into it and see if I can do something about it.

isovector:
This PR uses knowledge of in-scope [W] constraints to help solve otherwise ambiguous `Member` constraints. Eg if we have `Members [Embed IO, Embed Identity] r`, and are trying to solve `(Member (Embed m) r, MonadIO m)`, there is a unique choice of `m ~ IO` such that the program will typecheck.

Fixes #219

### DIFF
--- a/polysemy-plugin/polysemy-plugin.cabal
+++ b/polysemy-plugin/polysemy-plugin.cabal
@@ -77,6 +77,7 @@ test-suite polysemy-plugin-test
   type: exitcode-stdio-1.0
   main-is: Main.hs
   other-modules:
+      AmbiguousSpec
       BadSpec
       DoctestSpec
       ExampleSpec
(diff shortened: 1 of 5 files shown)


isovector:
This is not yet perfect. It doesn't work for polymorphic givens.

isovector:
This doesn't work for fundep-constrained things, but I'm OK with that.

PR #425, opened 2021-10-27
Loosen the type of weaving

isovector:
I'm trying to write shrinkable arbitrary instances of `Sem`, and need to pack arbitrary data inside the `m` in `e m a`. But weaving doesn't let me do that --- for no good reason.


isovector:
This turned out to be unnecessary.

PR #430, opened 2021-11-18
Fix a regression in the plugin where polymorphic given effects in interpreters could unify

tek:
This PR fixes a regression in #423 where the skolem set was being incorrectly created for interpreter-mode wanted constraints, leading to bad unifications.

### DIFF
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE CPP                        #-}
+{-# LANGUAGE DerivingStrategies         #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ViewPatterns               #-}
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


isovector:
Allegedly caused by #423. Will investigate.

isovector:
So the problem is the following set of unifications:

```
unifying    
  [[WD] hole{co_a9Nc} {1}:: Tagged () e_a9MF[tau:1]
                            ~# Tagged () (State ()) (CNonCanonical),
   [WD] hole{co_a9Ne} {1}:: Error () ~# e_a9MF[tau:1] (CNonCanonical),
   [WD] hole{co_a9Ng} {2}:: State () ~# e_a9MF[tau:1] (CNonCanonical)]
```

The first one is the one we want --- these other two are crap, but I'm not yet sure where they're coming from.

isovector:
@googleson78 is right, in that the plugin is now incorrectly getting itself into "interpreter mode":

```
member mode 
  [WD] hole{co_a7iP} {1}:: Tagged () e_a7ii[tau:1]
                           ~# Tagged () (State ()) (CNonCanonical)
interpreter mode
  Just [WD] hole{co_a7iR} {1}:: Error ()
                                ~# e_a7ii[tau:1] (CNonCanonical)
interpreter mode
  Just [WD] hole{co_a7iT} {2}:: State ()
                                ~# e_a7ii[tau:1] (CNonCanonical)
```

isovector:
The wanted context is very interesting when things fail:

```
wanted      
  [[WD] $dMember_a7iI {1}:: Member
                              (Tagged () e_a7ii[tau:1]) r_a7i0[sk:1] (CDictCan),
   [WD] $dMember_a7iK {1}:: Member
                              (Error ()) (e_a7ii[tau:1] : Error () : r_a7i0[sk:1]) (CDictCan),
   [WD] $dMember_a7iM {2}:: Member
                              (State ()) (e_a7ii[tau:1] : Error () : r_a7i0[sk:1]) (CDictCan)]
wanted finds
  [(effect name =  Tagged
    effect =  Tagged () e_a7ii[tau:1]
    row =  r_a7i0[sk:1]),
   (effect name =  Error
    effect =  Error ()
    row =  e_a7ii[tau:1] : Error () : r_a7i0[sk:1]),
   (effect name =  State
    effect =  State ()
    row =  e_a7ii[tau:1] : Error () : r_a7i0[sk:1])]
```

We're getting into interpreter mode because the rows have a `':` constructor. I don't know yet if this is expected behavior.

isovector:
The regression is this:

https://github.com/polysemy-research/polysemy/blob/95c534bb109f55bb542c26b7bca1916a3c5532d3/polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs#L96-L97

We need to add any effect skolems to the skolem set!

isovector:
mkWanted has this state when things go wrong:

```
mkWanted    
  (find_constraint = 
      (effect name =  Error
      effect =  Error ()
      row =  e_a7ii[tau:1] : Error () : r_a7i0[sk:1]),
   solve_context = (InterpreterUse True),
   given = e_a7ii[tau:1],
   wanted = Error ())
```

I think the trick is to just try to just get the skolem out of `given`

isovector:
Fixed. @tek PTAL.

PR #432, opened 2021-11-20
GHC 9.2 fix?

spacekitteh:
I only used the github editor; thus am relying on CI.


isovector:
Thanks for the patch! Included as part of #434

PR #433, opened 2021-11-20
Add GHC 9.2 to flake.nix

spacekitteh:
Note that currently polysemy-plugin doesn't yet compile with GHC 9.2, so that is expected to fail until #432 is finished.

Additionally, the flake.lock file needs to be updated, but I have no idea how to do that.


tek:
thanks! I set up the flake in https://github.com/polysemy-research/polysemy/tree/ghc921

isovector:
Thanks for the patch! Included as part of #434

PR #434, opened 2021-11-22
Support GHC 9.2

tek:
Caveat: The custom type error for `insertAt` has been disabled for this version since there appears to have been introduced a subtle change in the types used to match `InsertAtIndex` constraint, causing the incoherent instances to break.
Since the custom error feature isn't required for `insertAt` to work, I propose to postpone the fix for this.

Incorporates @spacekitteh's PRs #432 and #433.

### DIFF
--- a/polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
+++ b/polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
@@ -6,6 +6,7 @@ import           Data.Bool
 import           Data.Function (on)
 import           Data.Set (Set)
 import qualified Data.Set as S
+
 #if __GLASGOW_HASKELL__ >= 900
 import           GHC.Tc.Types.Constraint
 #elif __GLASGOW_HASKELL__ >= 810
@@ -72,16 +73,22 @@ unify solve_ctx = tryUnifyUnivarsButNotSkolems skolems
         InterpreterUse _ s -> s
         FunctionDef s      -> s
 
+#if __GLASGOW_HASKELL__ >= 902
+#define BINDME (const BindMe)
+#define APART (const Apart)
+#else
+#define BINDME BindMe
+#define APART Skolem
+#endif
 
 tryUnifyUnivarsButNotSkolems :: Set TyVar -> Type -> Type -> Maybe TCvSubst
 tryUnifyUnivarsButNotSkolems skolems goal inst =
   case tcUnifyTysFG
-         (bool BindMe Skolem . flip S.member skolems)
+         (bool BINDME APART . flip S.member skolems)
          [inst]
          [goal] of
     Unifiable subst -> pure subst
     _               -> Nothing
-
 
 ------------------------------------------------------------------------------
 -- | A wrapper for two types that we want to say have been unified.
--- a/flake.lock
+++ b/flake.lock
@@ -2,49 +2,66 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1618217525,
-        "narHash": "sha256-WGrhVczjXTiswQaoxQ+0PTfbLNeOQM6M36zvLn78AYg=",
+        "lastModified": 1637014545,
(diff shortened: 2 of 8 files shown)


polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
```diff
@@ -72,7 +71,16 @@ unify solve_ctx = tryUnifyUnivarsButNotSkolems skolems
         InterpreterUse _ s -> s
         FunctionDef s      -> s
 
-
+#if __GLASGOW_HASKELL__ >= 902
+tryUnifyUnivarsButNotSkolems :: Set TyVar -> Type -> Type -> Maybe TCvSubst
+tryUnifyUnivarsButNotSkolems skolems goal inst =
+  case tcUnifyTysFG
+         (bool (const BindMe) (const Apart) . flip S.member skolems)
```
isovector, line 78: This is the only bit that's different, no? For things like this I like to just define some CPP:

```haskell
#if __GLASGOW_HASKEL__ >= 902
#define BINDME (const BindMe)
#define APART (const Apart)
#else
#define BINDME BindMe
#define APART Skolem
#endif


tryUnifyUnivarsButNotSkolems :: Set TyVar -> Type -> Type -> Maybe TCvSubst
tryUnifyUnivarsButNotSkolems skolems goal inst =
  case tcUnifyTysFG
         (bool BINDME SKOLEM . flip S.member skolems)
         [inst]
         [goal] of
    Unifiable subst -> pure subst
    _               -> Nothing
```

polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
```diff
@@ -19,9 +19,8 @@ import           GHC.Core.Type
 import           GHC.Core.Unify
 import           GHC.Plugins (Outputable, ppr, parens, text, (<+>))
 #else
-import           Type
-import           Unify
-import           GhcPlugins (Outputable, ppr, parens, text, (<+>))
+import Type (Type, TyVar, eqType, nonDetCmpType, TCvSubst)
+import Unify (tcUnifyTysFG, BindFlag(Skolem, BindMe), UnifyResultM(Unifiable))
```
isovector, line 23: formatter gone rogue?

isovector (changes requested):
I'd prefer if `tryUnifyUnivarsButNotSkolems` had less duplication between GHC versions. Too easy to get out of sync as it is.

polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
tek: also missing an import :grimacing:

polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
tek: :heavy_check_mark:

polysemy-plugin/src/Polysemy/Plugin/Fundep/Unification.hs
tek: :heavy_check_mark:

isovector (approved):
lgtm!

spacekitteh:
So this still works while there isn't a GHC 9.0.2, but it will break then due to the typos.

tek:
if you're referring to the `__GLASGOW_HASKELL__` number, it only contains the first two parts of the version: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/intro.html#version-numbering

spacekitteh:
...huh. Ok, ignore that then!

spacekitteh:
Oh - polysemy-plugin needs to be pushed to hackage too!

isovector:
oops, yeah. nice catch!

isovector:
done https://github.com/polysemy-research/polysemy/commit/8f469182f04c555b4f9ac1931ee02e3bb2d6921e

PR #438, opened 2021-11-28
Remove lowering functions and forklift

isovector:


src/Polysemy/Async.hs
```diff
@@ -48,43 +46,6 @@ sequenceConcurrently :: forall t r a. (Traversable t, Member Async r) =>
...
-          pureT $ inspect ins <$> fa
-
-        Await a -> pureT =<< embed (A.wait a)
-        Cancel a -> pureT =<< embed (A.cancel a)
-    )  m
-{-# INLINE asyncToIO #-}
-
 ------------------------------------------------------------------------------
 -- | Run an 'Async' effect in terms of 'A.async' through final 'IO'.
 --
```
KingoftheHomeless, line 51: Documentation for `asyncToIOFinal` must be updated.

src/Polysemy/Resource.hs
```diff
@@ -138,42 +136,6 @@ resourceToIOFinal = interpretFinal $ \case
 {-# INLINE resourceToIOFinal #-}
```
KingoftheHomeless, line 136: The documentation for `resourceToIOFinal` must be updated.

src/Polysemy.hs
```diff
@@ -120,9 +120,6 @@ module Polysemy
   , reinterpret2H
   , reinterpret3H
 
-    -- * Combinators for Interpreting Directly to IO
-  , withLowerToIO
-
     -- * Kind Synonyms
   , Effect
   , EffectRow
```
KingoftheHomeless, line 125: `.@`, `.@@` should be removed.

KingoftheHomeless (changes requested):
The documentation needs a pass -- these are the ones I've been able to spot, but I suspect we've been mentioning lowering/forklift related stuff all over the place, and we should ensure such references go down the memory hole.

isovector:
@KingoftheHomeless PTAL. You were right, looking for the documentation to update found a few more things I should remove.

isovector:
I plan to merge this tomorrow unless there are any objections

PR #440, opened 2021-12-02
add Scoped

tek:
note: adds `BlockArguments`, which should be fine since we dropped 8.4.4


isovector:
I havent looked too closely at this particular PR, but I'm happy to get this merged. Green light from me!

tek:
cool, we will have to update it soon anyway when v2 is dropping

PR #442, opened 2021-12-21
Add link to "Porting to Polysemy"

eyeinsky:
Was searching for exactly this and today stumbled upon on reddit. I think it's a useful resource to link to given that mtl is *the* monad stacking library around, especially for those who aren't deep into effect libraries.


tek:
thanks!

PR #443, opened 2022-01-16
Copy docs of effect constructors to functions

TheMatten:
Uses new `putDoc` functionality to add documentation from constructors to generated functions. This may make it more practical to not expose original constructors at all.

### DIFF
--- a/src/Polysemy/Internal/TH/Effect.hs
+++ b/src/Polysemy/Internal/TH/Effect.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE CPP, TemplateHaskell #-}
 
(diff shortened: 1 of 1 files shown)


tek:
oh wow that's awesome

tek (approved):

googleson78:
Some thoughts I had:

* we should document this (and add it to the changelog) - I really hate it when libraries do some magic TH stuff without telling me
* is it possible that now the other `makeSem` which doesn't generate a type signature is redundant?
* related to the last one - are all the docs you can attach to a data constructor a subset of the ones you can attach to a function definition?

googleson78:
Sorry for the late comments, I totally missed the mail for the pr :sweat_smile:

tek:
please report to HR for sensitivity training

PR #445, opened 2022-01-19
Prevent `errorToIOFinal` from mixing errors up (#444)

### LINKED ISSUE #444: `errorToIOFinal` potentially mixes errors up
As per [what Alexis King found out about `Error` interpretation in `effectful`](https://github.com/arybczak/effectful/issues/1#issuecomment-867236416), I found that `polysemy` in fact is subject to the same problem for the `errorToIOFinal` handler. Take this `Filesystem` effect:

```haskell
data Filesystem :: Effect where
  ReadFile :: FilePath -> Filesystem m String
  WriteFile :: FilePath -> String -> Filesystem m ()
makeSem ''Filesystem
```

and an interpreter:
```haskell
newtype FsError = FsError String
  deriving stock (Show)
  deriving anyclass (Exception)

runFilesystemPure :: Member (Error FsError) es => Map FilePath String -> Sem (Filesystem ': es) a -> Sem es a
runFilesystemPure fs = evalState fs . reinterpret \case
  ReadFile path -> maybeM 
    (throw $ FsError $ "File not found: " ++ show path)
    pure $ gets (M.lookup path)
  WriteFile path contents -> modify $ M.insert path contents
```

Ideally, this test would give a `Left`:

```haskell
f :: IO (Either FsError (Either FsError String))
f = runFinal $ errorToIOFinal $ runFilesystemPure M.empty $ do
  errorToIOFinal $ readFile "nonexistent"
```

because the inner `errorToIOFinal` should not be able to see the implementation of `ReadFile`. But in fact we get

```haskell
>>> f
Right (Left (FsError "File not found: \"nonexistent\""))
```

This breaks effect encapsulation in a subtle way. It should be relatively easy to fix though; Polysemy can assign `OpaqueExc`s with `Unique` IDs on each `errorToIOFinal` calls.

re-xyr:
This hopefully fixes #444. I'm unexperienced with Polysemy's HO interpretation mechanism so I'm not sure if I impelemented this correctly; specifically I used `getInspectorS` and I'm unsure what to do at the `Nothing` case. It did pass the test in #444 though.

### DIFF
--- a/src/Polysemy/Error.hs
+++ b/src/Polysemy/Error.hs
@@ -25,14 +25,16 @@ module Polysemy.Error
   , errorToIOFinal
   ) where
 
-import qualified Control.Exception as X
+import qualified Control.Exception          as X
 import           Control.Monad
 import qualified Control.Monad.Trans.Except as E
-import           Data.Typeable
+import           Data.Unique                (Unique, hashUnique, newUnique)
+import           GHC.Exts                   (Any)
 import           Polysemy
 import           Polysemy.Final
 import           Polysemy.Internal
 import           Polysemy.Internal.Union
+import           Unsafe.Coerce              (unsafeCoerce)
 
 
 data Error e m a where
@@ -44,7 +46,7 @@ makeSem ''Error
 
 hush :: Either e a -> Maybe a
 hush (Right a) = Just a
-hush (Left _) = Nothing
+hush (Left _)  = Nothing
 
 
 ------------------------------------------------------------------------------
@@ -55,7 +57,7 @@ fromEither
     :: Member (Error e) r
     => Either e a
     -> Sem r a
-fromEither (Left e) = throw e
+fromEither (Left e)  = throw e
 fromEither (Right a) = pure a
 {-# INLINABLE fromEither #-}
 
@@ -103,7 +105,7 @@ fromExceptionVia
 fromExceptionVia f m = do
   r <- embed $ X.try m
   case r of
-    Left e -> throw $ f e
+    Left e  -> throw $ f e
     Right a -> pure a
 {-# INLINABLE fromExceptionVia #-}
 
@@ -137,7 +139,7 @@ fromExceptionSemVia f m = do
     s  <- getInitialStateS
     pure $ (fmap . fmap) Right m' `X.catch` \e -> (pure (Left e <$ s))
   case r of
-    Left e -> throw $ f e
+    Left e  -> throw $ f e
     Right a -> pure a
 {-# INLINABLE fromExceptionSemVia #-}
 
@@ -168,7 +170,7 @@ tryJust f m = do
       Right v -> return (Right v)
       Left e -> case f e of
                   Nothing -> throw e
-                  Just b -> return $ Left b
+                  Just b  -> return $ Left b
 {-# INLINABLE tryJust #-}
 
 ------------------------------------------------------------------------------
@@ -184,7 +186,7 @@ catchJust ef m bf = catch m handler
   where
       handler e = case ef e of
                     Nothing -> throw e
-                    Just b -> bf b
+                    Just b  -> bf b
 {-# INLINABLE catchJust #-}
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/Polysemy/Error.hs
```diff
@@ -242,7 +243,7 @@ mapError f = interpretH $ \case
 {-# INLINE mapError #-}
 
 
-newtype WrappedExc e = WrappedExc { unwrapExc :: e }
+data WrappedExc e = WrappedExc { getExcUid :: {-# UNPACK #-} !Unique, unwrapExc :: e }
```
KingoftheHomeless, line 246: Scrap the type argument. Use `GHC.Exts.Any` instead and `unsafeCoerce` as necessary, like `OpaqueExc` of in-other-words

src/Polysemy/Error.hs
```diff
@@ -261,35 +262,42 @@ instance (Typeable e) => X.Exception (WrappedExc e)
...
-    :: ( Typeable e
+    :: forall e r a
+    .  ( Typeable e
        , Member (Final IO) r
        )
     => Sem (Error e ': r) a
     -> Sem r (Either e a)
 errorToIOFinal sem = withStrategicToFinal @IO $ do
-  m' <- runS (runErrorAsExcFinal sem)
+  um <- liftS newUnique
```
KingoftheHomeless, line 272: Create the unique in the IO block instead.

```haskell
s <- getInitialStateS
m' <- bindS (`runErrorAsExcFinal` sem)
pure $ do
  uid <- newUnique
  (fmap . fmap) Right (m' (uid <$ s)) `catch` \exc@(WrappedExc uid' e) ->
    if uid == uid' then
      return $ Left (unsafeCoerce e <$ s)
    else
      X.throwIO exc
```
(untested code)

src/Polysemy/Error.hs
```diff
@@ -261,35 +262,42 @@ instance (Typeable e) => X.Exception (WrappedExc e)
 --
 -- @since 1.2.0.0
 errorToIOFinal
-    :: ( Typeable e
+    :: forall e r a
+    .  ( Typeable e
```
KingoftheHomeless, line 266: Once you've converted to using `Any`, this `Typeable` constraint can be dropped.

src/Polysemy/Error.hs
```diff
@@ -242,7 +243,7 @@ mapError f = interpretH $ \case
 {-# INLINE mapError #-}
 
 
-newtype WrappedExc e = WrappedExc { unwrapExc :: e }
+data WrappedExc e = WrappedExc { getExcUid :: {-# UNPACK #-} !Unique, unwrapExc :: e }
   deriving (Typeable)
 
 instance Typeable e => Show (WrappedExc e) where
```
KingoftheHomeless, line 249: When the type argument is dropped, the `Show` instance should be replaced. The very wordy instance from in-other-words is probably warranted (adapted to fit polysemy, of course).

KingoftheHomeless (changes requested):

src/Polysemy/Error.hs
googleson78: What's the reason for this?

googleson78: Ah, to drop the Typeable constraint? Didn't see the comments below. In what way is the `Typeable` constraint bad?

KingoftheHomeless: It's no as much that it's bad as it is completely unnecessary, and unnecessary things are bad. It pollutes constraints, making using it just that little bit more harder to use. As the catching logic can be completely driven by `Unique`s, `Typeable` has just become clutter.

re-xyr:
I'm not exactly sure why tests are failing now; it seems to me that all test cases have passed but the test suite somehow exited with a non-zero exit code.

src/Polysemy/Error.hs
```diff
@@ -261,35 +273,32 @@ instance (Typeable e) => X.Exception (WrappedExc e)
...
+runErrorAsExcFinal uid = interpretFinal $ \case
+  Throw e   -> pure $ X.throwIO $ WrappedExc uid (unsafeCoerce e)
   Catch m h -> do
     m' <- runS m
     h' <- bindS h
     s  <- getInitialStateS
-    pure $ X.catch m' $ \(se :: WrappedExc e) ->
-      h' (unwrapExc se <$ s)
+    pure $ catchWithUid uid m' $ \(WrappedExc _ e) ->
+      h' (unsafeCoerce e <$ s)
```
KingoftheHomeless, line 303: This is wrong, and will result in runtime corruption (likely why the test fails, it segfaults). It should be:
``
    pure $ catchWithUid uid m' $ \e -> h' (e <$ s)
```

Comment (re-xyr) [reply src/Polysemy/Error.hs]: Oops, missed that bit. Should be fine now.

**Comment by KingoftheHomeless** [comment]
This deserves a test, you can adapt [the one from `in-other-words`.}(https://github.com/KingoftheHomeless/in-other-words/blob/master/test/ErrorSpec.hs). You can drop test5 and test6.

**Comment by KingoftheHomeless** [comment]
Also, add an item to the changelog! This constitutes a breaking change as the semantics have changed.

**Comment by re-xyr** [comment]
*Strictly speaking* this will be a breaking change, but I think in spirit this falls into the cagegory of a bugfix? I don't see anyone will depend on the behavior of a handler mixing errors up.

**Comment by KingoftheHomeless** [comment]
Well, we shouldn't assume; regardless, it will be bundled together with other breaking changes, so it's not like this will upgrade a minor release to a major one.

File: src/Polysemy/Error.hs [hunk 6]
```diff
@@ -242,14 +244,24 @@ mapError f = interpretH $ \case
 {-# INLINE mapError #-}
 
 
-newtype WrappedExc e = WrappedExc { unwrapExc :: e }
-  deriving (Typeable)
+data WrappedExc = WrappedExc !Unique Any
```
KingoftheHomeless, line 247: Stick an `{-# UNPACK #-}` on the `Unique`. (Unless it's a no-op? I don't think it is for `Unique`s.)

KingoftheHomeless (commented):
This looks good now! Only thing I'm wondering about is the lack of `{-# UNPACK #-}` on the `Unique` in `WrappedExc`. I think I saw it in your earlier commits, but not sure. I don't remember if the representation of `Unique` is such that UNPACK has any effect. To be sure, compile the module with -O2 and the pragma -- you should get a warning that the pragma is useless if it is.
Once you know (if you don't already), either push up a commit with the pragma if it matters, or just comment if it doesn't, and I'll approve.

src/Polysemy/Error.hs
re-xyr: It is a no-op, since `Unique` is already a `newtype` by itself.

KingoftheHomeless (approved):
Alright, with that settled, this is ready to merge. Thanks for this!

PR #450, opened 2022-05-17
Fix formatting issue in 'tag' documentation

MorrowM:
https://hackage.haskell.org/package/polysemy-1.7.1.0/docs/Polysemy-Tagged.html#v:tag

Currently:
image

With this commit:

image


tek:
thanks!

PR #455, opened 2022-09-28
add call site parameter to Scoped

tek:
Since we haven't cut a release since we added `Scoped` and I've added some features in the meantime, I thought I'd upstream them now because I find them very useful.

I split them in two PRs so we can discuss separately.

This one concerns scope parameters that are passed to the interpreter from the call site, allowing the user to supply configuration for the resources, for example a process config:

```haskell
withSystemProcess ::
  Member (Scoped (ProcessConfig () () ()) resource SystemProcess) r =>
  ProcessConfig () () () ->
  InterpreterFor SystemProcess r
withSystemProcess =
  scoped
```

which can then be used by the interpreter to create the process:

```haskell
withProcess ::
  ProcessConfig () () () ->
  (Process Handle Handle Handle -> Sem r a) ->
  Sem r a

interpretSystemProcess :: InterpreterFor (Scoped (ProcessConfig () () ()) (Process Handle Handle Handle) SystemProcess) r
interpretSystemProcess =
  interpretScoped withProcess \ process -> \case
    ReadStdout -> hGetSome (getStdout process) 4096
```

Opinions? I think it's rather straightforward and I ran into a lot of situations where I needed it.
I kept a type alias `Scoped_` for the parameterless version, do you think that's sensible?

### DIFF
--- a/src/Polysemy/Internal/Scoped.hs
+++ b/src/Polysemy/Internal/Scoped.hs
@@ -25,26 +25,69 @@ import Polysemy.Internal.Combinators (transform)
 --
 -- The constructors are not intended to be used directly; the smart constructor 'scoped' is used like a local
 -- interpreter for @effect@.
-data Scoped (resource :: Type) (effect :: Effect) :: Effect where
-  Run :: ∀ resource effect m a . resource -> effect m a -> Scoped resource effect m a
-  InScope :: ∀ resource effect m a . (resource -> m a) -> Scoped resource effect m a
+-- 'scoped' takes an argument of type @param@, which will be passed through to the interpreter, to be used by the
+-- resource allocation function.
+data Scoped (param :: Type) (resource :: Type) (effect :: Effect) :: Effect where
+  Run :: ∀ param resource effect m a . resource -> effect m a -> Scoped param resource effect m a
+  InScope :: ∀ param resource effect m a . param -> (resource -> m a) -> Scoped param resource effect m a
+
+-- |A convenience alias for a scope without parameters.
+type Scoped_ resource effect =
+  Scoped () resource effect
 
 -- |Constructor for 'Scoped', taking a nested program and transforming all instances of @effect@ to
--- @Scoped resource effect@.
+-- @'Scoped' param resource effect@ and wrapping the result with 'InScope'.
+--
+-- This allows the effective interpreter to bracket the nested program with a resource from a distance.
+--
+-- The value @param@ is passed to the interpreter.
 scoped ::
-  ∀ resource effect r .
-  Member (Scoped resource effect) r =>
+  ∀ param resource effect r .
+  Member (Scoped param resource effect) r =>
+  param ->
   InterpreterFor effect r
-scoped main =
-  send $ InScope @resource @effect \ resource ->
-    transform @effect (Run resource) main
+scoped param main =
+  send $ InScope @param @resource @effect param \ resource ->
+    transform @effect (Run @param resource) main
+{-# inline scoped #-}
+
+-- |Constructor for 'Scoped_', taking a nested program and transforming all instances of @effect@ to
+-- @'Scoped_' resource effect@ and wrapping the result with 'InScope'.
+--
+-- This allows the effective interpreter to bracket the nested program with a resource from a distance.
+scoped_ ::
+  ∀ resource effect r .
+  Member (Scoped_ resource effect) r =>
+  InterpreterFor effect r
+scoped_ =
+  scoped @() @resource ()
+{-# inline scoped_ #-}
+
+-- |Transform the parameters of a 'Scoped' program.
+--
+-- This allows incremental additions to the data passed to the interpreter, for example to create an API that permits
+-- different ways of running an effect with some fundamental parameters being supplied at scope creation and some
+-- optional or specific parameters being selected by the user downstream.
+rescope ::
+  ∀ param0 param1 resource effect r .
+  Member (Scoped param1 resource effect) r =>
+  (param0 -> param1) ->
+  InterpreterFor (Scoped param0 resource effect) r
+rescope fp =
+  transform \case
+    Run res e ->
+      Run @param1 res e
+    InScope p main ->
+      InScope (fp p) main
+{-# inline rescope #-}
 
 -- |Helper for @runScoped@.
-interpretH' ::
+interpretWeaving ::
   ∀ e r .
   (∀ x . Weaving e (Sem (e : r)) x -> Sem r x) ->
   InterpreterFor e r
-interpretH' h (Sem m) =
+interpretWeaving h (Sem m) =
   Sem \ k -> m $ decomp >>> \case
     Right wav -> runSem (h wav) k
-    Left g -> k $ hoist (interpretH' h) g
+    Left g -> k $ hoist (interpretWeaving h) g
+{-# inline interpretWeaving #-}
--- a/src/Polysemy/Internal/Scoped.hs
+++ b/src/Polysemy/Internal/Scoped.hs
@@ -59,8 +59,10 @@ import Polysemy.Internal.Union (Weaving, decomp, hoist)
 --
 -- > interpretWriteFile :: Member (Embed IO) => InterpreterFor (Scoped FilePath Handle Write) r
 -- > interpretWriteFile =
--- >   interpretScoped (\ name use -> bracket (openFile name WriteMode) hClose use) \ handle -> \case
--- >     Write line -> embed (Text.hPutStrLn handle line)
+-- >   interpretScoped allocator handler
+-- >   where
+-- >     allocator name use = bracket (openFile name WriteMode) hClose use
+-- >     handler fileHandle (Write line) = embed (Text.hPutStrLn fileHandle line)
 --
 -- Essentially, the @bracket@ is executed at the point where @scoped@ was called, wrapping the following block.
 -- When the second @scoped@ is executed, another call to @bracket@ is performed.
--- a/src/Polysemy/Internal/Scoped.hs
+++ b/src/Polysemy/Internal/Scoped.hs
@@ -72,8 +72,8 @@ import Polysemy.Internal.Union (Weaving, decomp, hoist)
 --
 -- This makes it possible to use an in-memory interpreter for testing:
 --
--- > interpretWriteMap :: Member (Output (FilePath, Text)) r => InterpreterFor (Scoped FilePath FilePath Write) r
--- > interpretWriteMap =
+-- > interpretWriteOutput :: Member (Output (FilePath, Text)) r => InterpreterFor (Scoped FilePath FilePath Write) r
+-- > interpretWriteOutput =
 -- >   interpretScoped (\ name use -> use name) \ name -> \case
 -- >     Write line -> output (name, line)
 --
--- a/src/Polysemy/Internal/Scoped.hs
+++ b/src/Polysemy/Internal/Scoped.hs
@@ -3,12 +3,13 @@
 
 module Polysemy.Internal.Scoped where
 
-import Polysemy.Internal (send, Member, InterpreterFor, Sem(Sem), runSem)
-import Data.Kind (Type)
-import Polysemy.Internal.Kind (Effect)
-import Polysemy.Internal.Union (Weaving, hoist, decomp)
 import Control.Arrow ((>>>))
+import Data.Kind (Type)
+
+import Polysemy.Internal (InterpreterFor, Member, Sem (Sem), runSem, send)
 import Polysemy.Internal.Combinators (transform)
+import Polysemy.Internal.Kind (Effect)
+import Polysemy.Internal.Union (Weaving, decomp, hoist)
 
 -- |@Scoped@ transforms a program so that @effect@ is associated with a @resource@ within that program.
 -- This requires the interpreter for @effect@ to be parameterized by @resource@ and constructed for every program using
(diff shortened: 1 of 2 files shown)


src/Polysemy/Internal/Scoped.hs
```diff
@@ -25,26 +25,69 @@ import Polysemy.Internal.Combinators (transform)
...
+  Run :: ∀ param resource effect m a . resource -> effect m a -> Scoped param resource effect m a
+  InScope :: ∀ param resource effect m a . param -> (resource -> m a) -> Scoped param resource effect m a
+
+-- |A convenience alias for a scope without parameters.
+type Scoped_ resource effect =
+  Scoped () resource effect
 
 -- |Constructor for 'Scoped', taking a nested program and transforming all instances of @effect@ to
--- @Scoped resource effect@.
+-- @'Scoped' param resource effect@ and wrapping the result with 'InScope'.
```
googleson78, line 39: I don't think `InScope` should be mentioned in the docs - it's not exported from the non-`Internal` module

src/Polysemy/Internal/Scoped.hs
```diff
@@ -25,26 +25,69 @@ import Polysemy.Internal.Combinators (transform)
...
+  Member (Scoped param resource effect) r =>
+  param ->
+  InterpreterFor effect r
+scoped param main =
+  send $ InScope @param @resource @effect param \ resource ->
+    transform @effect (Run @param resource) main
+{-# inline scoped #-}
+
+-- |Constructor for 'Scoped_', taking a nested program and transforming all instances of @effect@ to
+-- @'Scoped_' resource effect@ and wrapping the result with 'InScope'.
```
googleson78, line 55: ditto

src/Polysemy/Internal/Scoped.hs
```diff
@@ -25,26 +25,69 @@ import Polysemy.Internal.Combinators (transform)
...
+--
+-- This allows the effective interpreter to bracket the nested program with a resource from a distance.
+scoped_ ::
```
googleson78, line 58: Perhaps `scoped_` docs should be something like
```
-- |Constructor for 'Scoped_'. Similar to 'scoped', but <explain differences and importantly give an example when you want this>
```
?
Or perhaps the example would be better attached to `type Scoped_`?

src/Polysemy/Internal/Scoped.hs
```diff
@@ -25,26 +25,69 @@ import Polysemy.Internal.Combinators (transform)
...
+scoped_ =
+  scoped @() @resource ()
+{-# inline scoped_ #-}
+
+-- |Transform the parameters of a 'Scoped' program.
+--
+-- This allows incremental additions to the data passed to the interpreter, for example to create an API that permits
+-- different ways of running an effect with some fundamental parameters being supplied at scope creation and some
+-- optional or specific parameters being selected by the user downstream.
+rescope ::
```
googleson78, line 71: I find the `re` prefix a bit confusing here, because most other `re-`s modify the effect on top of the stack, as opposed to "subsuming", e.g. https://hackage.haskell.org/package/polysemy-1.7.1.0/docs/Polysemy-Internal-Combinators.html#v:reinterpret

Actually this also makes me wonder why you used `transform` here instead of leaving the `Scoped` at the head of the effects - is this much more commonly practical/useful?

src/Polysemy/Scoped.hs
```diff
@@ -2,21 +2,86 @@ module Polysemy.Scoped (
...
-import Polysemy.Internal.Tactics (Tactical, runTactics)
+import Polysemy.Internal.Combinators (interpretH)
+import Polysemy.Internal.Scoped (Scoped (InScope, Run), interpretWeaving, rescope, scoped, scoped_)
+import Polysemy.Internal.Tactics (Tactical, runTactics, runTSimple, liftT)
+import Polysemy.Internal.Union (Weaving (Weaving), injWeaving)
+
+-- |Interpreter for 'Scoped', taking a @resource@ allocation function and a parameterized handler for the plain
+-- @effect@.
+--
+-- >>> runScoped withResource scopedHandler
```
googleson78, line 28: Is this code snippet related to `interpretScopedH`?

src/Polysemy/Scoped.hs
```diff
@@ -2,21 +2,86 @@ module Polysemy.Scoped (
...
+--
+-- @withResource@ is a callback function, allowing the user to acquire the resource for each program wrapped by 'scoped'
+-- using other effects, with an additional argument that contains the call site parameter passed to 'scoped'.
```
googleson78, line 31: Examples would be awesome if you can spare the time :sweat_smile: But I guess this is true for almost everything in the library, so it's not really very important to start adding them now.

src/Polysemy/Scoped.hs
```diff
@@ -26,75 +91,34 @@ import Polysemy.Internal.Tactics (Tactical, runTactics)
 -- @withResource@ is a callback function, allowing the user to acquire the resource for each program from other effects.
 --
 -- @scopedInterpreter@ is a regular interpreter that is called with the @resource@ argument produced by @scope@.
+-- This is mostly useful if you want to reuse an interpreter that you cannot easily rewrite (like from another library).
+-- If you have full control over the implementation, 'interpreterScoped' should be preferred.
```
googleson78, line 95: `interpreterScoped` is a typo, right?

src/Polysemy/Internal/Scoped.hs
tek: ditto

src/Polysemy/Internal/Scoped.hs
tek: I added an extended example to `Scoped` and referenced it here. satisfactory?

src/Polysemy/Internal/Scoped.hs
tek: leaving the `Scoped` at the head would fix your stack for the entire program. this combi is intended for use in business logic!

tek: regarding the naming, please make a suggestion!

src/Polysemy/Scoped.hs
tek: whoopsie!

src/Polysemy/Scoped.hs
tek: I added one to `Scoped`, and there is also the link to my blog post.
do you think we should reference the `Scoped` docs in each interpreter combi? instead of having any prose there?

src/Polysemy/Scoped.hs
tek: indeed it is!

src/Polysemy/Internal/Scoped.hs
googleson78: Maybe it's just good enough like this.

src/Polysemy/Internal/Scoped.hs
```diff
@@ -21,30 +22,138 @@ import Polysemy.Internal.Combinators (transform)
...
+-- As an example, imagine an effect for writing lines to a file:
+--
+-- > data Write :: Effect where Write :: Text -> Write m ()
+-- > makeSem ''Write
+--
+-- If we now have the following requirements:
+--
+-- 1. The file should be opened and closed right before and after the part of the program in which we write lines
+-- 2. The file name should be specifiable at the point in the program where writing begins
+-- 3. We don't want to commit to IO, lines should be stored in memory when running tests
```
googleson78, line 41: I think this example is great, it seems minimal and concretely shows under when `Scoped` is useful. Personally I would add something along the lines of `(notice how these requirements are very similar to what we have in regards to transactions and a connection in databases)`, but maybe that use case is not that prevalent.

src/Polysemy/Internal/Scoped.hs
```diff
@@ -21,30 +22,138 @@ import Polysemy.Internal.Combinators (transform)
...
+-- Here 'scoped' creates a prompt for an interpreter to start allocating a resource for @"file1.txt"@ and handling
+-- @Write@ actions using that resource.
+-- When the 'scoped' block ends, the resource should be freed.
+--
+-- The interpreter may look like this:
+--
+-- > interpretWriteFile :: Member (Embed IO) => InterpreterFor (Scoped FilePath Handle Write) r
+-- > interpretWriteFile =
+-- >   interpretScoped (\ name use -> bracket (openFile name WriteMode) hClose use) \ handle -> \case
+-- >     Write line -> embed (Text.hPutStrLn handle line)
```
googleson78, line 63: I think it would be better to pull out the allocator and interpter functions into a `let`, this is a bit syntactically noisy for a "tutorial"-like implementation imo.

src/Polysemy/Scoped.hs
```diff
@@ -26,75 +91,34 @@ import Polysemy.Internal.Tactics (Tactical, runTactics)
...
-          ex <$> runTactics s (raise . run . wv) ins (run . wv) (scopedHandler resource act)
-        InScope main ->
-          ex <$> withResource \ resource -> run (wv (main resource <$ s))
-
--- |Variant of 'runScoped' that takes a handler instead of an interpreter.
-interpretScoped ::
-  ∀ resource effect r .
-  (∀ x . (resource -> Sem r x) -> Sem r x) ->
-  (∀ r0 x . resource -> effect (Sem r0) x -> Sem r x) ->
-  InterpreterFor (Scoped resource effect) r
```
googleson78, line 79: This old version is effectively equivalent to what would be `interpretScoped_`, right? Perhaps it's worth keeping it under that moniker? But then it would make sense to duplicate all the interpreters for the non-parametrised version :thinking:

src/Polysemy/Internal/Scoped.hs
```diff
@@ -21,30 +22,138 @@ import Polysemy.Internal.Combinators (transform)
...
+-- When the second @scoped@ is executed, another call to @bracket@ is performed.
+--
+-- The effect of this is that the operation that uses @Embed IO@ was moved from the call site to the interpreter, while
+-- the interpreter may be executed at the outermost layer of the app.
+--
+-- This makes it possible to use an in-memory interpreter for testing:
+--
+-- > interpretWriteMap :: Member (Output (FilePath, Text)) r => InterpreterFor (Scoped FilePath FilePath Write) r
+-- > interpretWriteMap =
+-- >   interpretScoped (\ name use -> use name) \ name -> \case
```
googleson78, line 75: I'm not exactly sure how `use name` type checks here. In order to do this, we need `parameter ~ resource`, right? Do you *need* that to be true in order to write this program? I guess so that you can `output (name, line)` and have it type check as well?

src/Polysemy/Internal/Scoped.hs
googleson78: Although I guess in that case usually you reuse the same connection, hence `Scoped_` would be closer.

src/Polysemy/Internal/Scoped.hs
```diff
@@ -21,30 +22,138 @@ import Polysemy.Internal.Combinators (transform)
...
+--
+-- Essentially, the @bracket@ is executed at the point where @scoped@ was called, wrapping the following block.
+-- When the second @scoped@ is executed, another call to @bracket@ is performed.
+--
+-- The effect of this is that the operation that uses @Embed IO@ was moved from the call site to the interpreter, while
+-- the interpreter may be executed at the outermost layer of the app.
+--
+-- This makes it possible to use an in-memory interpreter for testing:
+--
+-- > interpretWriteMap :: Member (Output (FilePath, Text)) r => InterpreterFor (Scoped FilePath FilePath Write) r
```
googleson78, line 73: The name is a bit confusing, I would expect you to be writing to a `Map`. Do you mean `Map` as in "map `Write` to `Output`"? I think just `interpretWriteAsOutput` is good enough.

src/Polysemy/Internal/Scoped.hs
googleson78: Awesome

src/Polysemy/Scoped.hs
googleson78: Awesome.

Yeah I think it makes sense to point to `Scoped` from all the combinators, although I expect anyone looking at the combinators to also be looking at `Scoped` (unless you're looking at the smart constructor `scoped`, in which case it might be still necessary)

Come to think of it, I think I thought of asking about an example here is because you wrote more docs here in the first place, which I thought would be better suited to have at the `Scoped` type. But I think right now it's fine as is. 

So feel free to resolve if you think it's not necessary to add links to `Scoped` from all the combinators

src/Polysemy/Scoped.hs
```diff
@@ -26,75 +91,34 @@ import Polysemy.Internal.Tactics (Tactical, runTactics)
 -- @withResource@ is a callback function, allowing the user to acquire the resource for each program from other effects.
 --
 -- @scopedInterpreter@ is a regular interpreter that is called with the @resource@ argument produced by @scope@.
+-- This is mostly useful if you want to reuse an interpreter that you cannot easily rewrite (like from another library).
+-- If you have full control over the implementation, 'interpreteScoped' should be preferred.
 --
 -- /Note/: This function will be called for each action in the program, so if the interpreter allocates any resources,
 -- they will be scoped to a single action. Move them to @withResource@ instead.
```
googleson78, line 98: Should this warning be moved to the type perhaps? Or added to all the other combinators that do this as well?

src/Polysemy/Scoped.hs
tek: > Come to think of it, I think I thought of asking about an example here is because you wrote more docs here in the first place, which I thought would be better suited to have at the Scoped type.

That's what I assumed! It's a reasonable concern though, I always struggle to decide whether to duplicate docs for names that have almost identical purposes.

In the HLS workflow, you would maybe start typing a common prefix and then cycle through the completions, reading the docs for each of them to decide which one is appropriate.
In that case, referring to a name that isn't listed due to no common prefix (`scoped` vs `Scoped`), you would have to change to a different workflow.
On the other hand, if you already _know_ that you want `interpretScopedX`, then you probably already know the fundamentals of the concept, and the information you need is how the specific `X` differs from the others, so only describing the specifics of the interpreter variant would make most sense.

In the workflow in which you're browsing Hackage to learn how it works, the reference is easily followed, and duplication of the prose would probably only cause cognitive overhead for finding the information that you haven't already read on the previous Haddock.

What do you think?

src/Polysemy/Internal/Scoped.hs
tek: well the parameter could be something else, like a flag indicating rollback behaviour?

src/Polysemy/Internal/Scoped.hs
tek: used a `where`, satisfactory?

src/Polysemy/Scoped.hs
tek: yeah that's a dilemma. in particular with the other PR there would be a rather large number of possible combinations.

We could decide that only the most likely variants should be special-cased, like the one you suggest here.

There's also the issue of naming with `interpretScopedAs`, which I chose back then to convey that the bracketing function is a plain action, but now it is parameterized by `param`.

Not sure what to do here, but it's always possible to add convenience stuff later.

src/Polysemy/Internal/Scoped.hs
tek: well yeah, it's a degenerate case. we're simply not doing anything in the allocator.

src/Polysemy/Internal/Scoped.hs
tek: oops! I started out with `State (Map FilePath [Text])` but decided that it was unnecessarily complex for the example :sweat_smile:

src/Polysemy/Scoped.hs
tek: no, that is specific to `runScoped` because it uses a blackbox interpreter.
with a handler, the problem cannot occur.

src/Polysemy/Internal/Scoped.hs
googleson78: Right.

src/Polysemy/Internal/Scoped.hs
googleson78: awesome

src/Polysemy/Scoped.hs
googleson78: I agree with your assessment.

PR #456, opened 2022-09-28
add variants of interpretScoped that allow additional local effects

tek:
This builds upon the other PR #455.

Essentially, it permits a scope handler to use local effects that are interpreted by the resource allocation function.
The implementation is pretty hacky, so I would appreciate some advice, @KingoftheHomeless.

I'm pretty sure that the `error` can never be hit unless `Run` is sent manually, but it would surely be nice to avoid it nonetheless.

### DIFF
--- a/src/Polysemy/Scoped.hs
+++ b/src/Polysemy/Scoped.hs
@@ -1,3 +1,5 @@
+{-# language AllowAmbiguousTypes #-}
+
 module Polysemy.Scoped (
   -- * Effect
   Scoped,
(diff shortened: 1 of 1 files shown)


KingoftheHomeless:
I'll try to take a look at this tomorrow afternoon, but it might have to wait until sunday.

googleson78:
I changed the target branch of this pr to make the changes that are exclusive to this pr clearer.

src/Polysemy/Scoped.hs
```diff
@@ -82,6 +88,74 @@ interpretScopedAs ::
...
+{-# inline interpretScopedAs #-}
+
+-- |Higher-order interpreter for 'Scoped' that allows the handler to use additional effects that are interpreted by the
+-- resource allocator.
+--
+-- /Note/: It is necessary to specify the list of local interpreters with a type application; GHC won't be able to
+-- figure them out from the type of @withResource@:
+--
+-- > interpretScopedWithH @[AtomicState Int, Reader Bool] withResource \ _ -> \case
+-- >   SomeAction -> atomicPut . (> 0) =<< ask
```
googleson78, line 100: I'm not very clear on where these additional effects are interpreted in this example. Can you point it out?

tek: I planned to write a better example later, but I added `withResource` for you now. the types were also wrong :sweat_smile:

KingoftheHomeless: Make a higher-order example so tactical actually gets some use! And have a first-order example for `interpretScopedWith`.

tek: I adapted [this one](https://git.tryp.io/tek/ribosome/src/branch/main/packages/host/lib/Ribosome/Host/Interpreter/MState.hs#L56) for an example.
It would probably look nicer if I could also use `Lock` there, but it's not a core effect.

Don't know whether it's such a good example because the local effect part is purely for expressivity purposes – I could just as well use a raw `TVar` here.

tek: to really drive the point home you could make both `Lock` and `AtomicState` scoped effects as well, then there wouldn't be any `Embed IO` in the interpreter :sweat_smile: would also showcase the possibility of using `Scoped` in a resource allocator, which is pretty neat in my opinion

PR #459, opened 2022-10-02
Extend Scoped

KingoftheHomeless:
WIP cleanups to #455 and #456. Will be continually added upon. Keeping them separate so we can discuss if there's any change you have questions about or disagree with.

### DIFF
--- a/src/Polysemy/Scoped.hs
+++ b/src/Polysemy/Scoped.hs
@@ -1,3 +1,5 @@
+{-# language AllowAmbiguousTypes #-}
+
 module Polysemy.Scoped (
   -- * Effect
   Scoped,
@@ -12,15 +14,19 @@ module Polysemy.Scoped (
   interpretScopedH',
   interpretScoped,
   interpretScopedAs,
+  interpretScopedWithH,
+  interpretScopedWith,
+  interpretScopedWith_,
   runScoped,
   runScopedAs,
 ) where
 
-import Polysemy.Internal (InterpreterFor, Sem, liftSem, raise)
+import Polysemy.Internal (InterpreterFor, Sem, liftSem, raise, Append, insertAt)
 import Polysemy.Internal.Combinators (interpretH)
 import Polysemy.Internal.Scoped (Scoped (InScope, Run), interpretWeaving, rescope, scoped, scoped_)
 import Polysemy.Internal.Tactics (Tactical, runTactics, runTSimple, liftT)
 import Polysemy.Internal.Union (Weaving (Weaving), injWeaving)
+import Polysemy.Internal.Index (InsertAtIndex)
 
 -- |Interpreter for 'Scoped', taking a @resource@ allocation function and a parameterized handler for the plain
 -- @effect@.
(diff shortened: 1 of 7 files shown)


src/Polysemy/Scoped.hs
```diff
@@ -90,103 +95,112 @@ interpretScopedAs resource =
...
     inScope =
       interpretWeaving \case
         Weaving (InScope param main) s wv ex _ ->
-          insertAt @0 @extra (ex <$> withResource param \ resource -> inScope (wv (main resource <$ s)))
+          restack (extendMembershipLeft (singList @extra))
+            (ex <$> withResource param \resource ->
+                inScope (wv (main resource <$ s)))
         Weaving (Run resource act) s wv ex ins ->
-          ex <$> runTactics s (raise . inScope . wv) ins (inScope . wv) (scopedHandler resource act)
+          -- Smells dubious, ask. God I hate tactics.
```
googleson78, line 134: ! @tek

tek: what do you suspect to go wrong there?

KingoftheHomeless: Oh yeah, uhh, figured I'd ask this in a call. I don't think I've ever had to use `runTactics` outside of `interpretH`, but that could be when I've worked on `Exceptional` I've based the work on the core rewrite. If this is just code adapted from `interpretH`, then it's fine; otherwise, it seems dubious.

tek: hm yeah I think I just copied that from `interpretH`

KingoftheHomeless:
@tek In general I notice we have very different stylistic tastes -- I honestly don't like yours! (And I'm guessing you don't like mine, that I've used in other parts of polysemy and `in-other-words`.) Are you using any particular style enforcer?

tek:
no I format code by hand and some vim functions :sweat_smile: but feel free to change it to your liking.

I format everything to optimize for line-based manipulation with vim, so I'm aware that it probably seems exotic if that's not a constraint for you

tek:
ok, I hope this documentation rewrite is satisfactory now :slightly_smiling_face: please have a look, @KingoftheHomeless!

src/Polysemy/Scoped.hs
```diff
@@ -28,21 +28,21 @@ import Polysemy.Internal.Combinators
...
---
--- @scopedHandler@ is a handler like the one expected by 'interpretH' with an
--- additional parameter that contains the @resource@ allocated by
--- @withResource@.
+-- This combinator is analogous to 'interpretH' in that it allows the handler to
+-- use the 'Tactical' environment and transforms the effect into other effects
+-- on the stack.
 interpretScopedH ::
   ∀ param resource effect r .
+  -- | A callback function that allows the user to acquire a resource for each
```
KingoftheHomeless, line 40: Format for arguments is `-- ^` -- unless Haddock has improved since I last knew it.

tek: this is valid: https://haskell-haddock.readthedocs.io/en/latest/markup.html#documenting-parts-of-a-declaration

tek: my impression is that the point of `-- ^` is to allow the docstring to be on the same line

KingoftheHomeless: Well alright then!

KingoftheHomeless (commented):
@tek Merge at your leisure. You can get rid of the .cabal and stack.yaml changes if they inconvenience you for any reason, but sooner or later they'll have to be merged, anyway.

tek:
:partying_face:

PR #462, opened 2022-10-11
Scoped without resource param

tek:
Since the `*With` interpreter uses `error` for the case that `Run` is encountered without a wrapping `InScope`, I realized that this would allow the resource to be visible in the interpreter only.
What do you think, @KingoftheHomeless @googleson78 ?
Bit awkward for sure, but not having to haul that parameter around everywhere would certainly be quite the increase in ergonomics.

### DIFF
--- a/src/Polysemy/Internal/Scoped.hs
+++ b/src/Polysemy/Internal/Scoped.hs
@@ -7,10 +7,9 @@ import Data.Kind (Type)
 
 import Polysemy
 
--- | @Scoped@ transforms a program so that @effect@ is associated with a
--- @resource@ within that program. This requires the interpreter for @effect@ to
--- be parameterized by @resource@ and constructed for every program using
--- @Scoped@ separately.
+-- | @Scoped@ transforms a program so that an interpreter for @effect@ may
+-- perform arbitrary actions, like resource management, before and after the
+-- computation wrapped by a call to 'scoped' is executed.
 --
 -- An application for this is @Polysemy.Conc.Events@ from
 -- <https://hackage.haskell.org/package/polysemy-conc>, in which each program
@@ -20,6 +19,8 @@ import Polysemy
 -- database effect.
 --
 -- For a longer exposition, see <https://www.tweag.io/blog/2022-01-05-polysemy-scoped/>.
+-- Note that the interface has changed since the blog post was published: The
+-- @resource@ parameter is no longer necessary.
 --
 -- Resource allocation is performed by a function passed to
 -- 'Polysemy.Scoped.interpretScoped'.
@@ -46,7 +47,7 @@ import Polysemy
 --
 -- Then we can take advantage of 'Scoped' to write this program:
 --
--- > prog :: Member (Scoped FilePath resource Write) r => Sem r ()
+-- > prog :: Member (Scoped FilePath Write) r => Sem r ()
 -- > prog = do
 -- >   scoped "file1.txt" do
 -- >     write "line 1"
@@ -61,7 +62,7 @@ import Polysemy
 --
 -- The interpreter may look like this:
 --
--- > interpretWriteFile :: Members '[Resource, Embed IO] => InterpreterFor (Scoped FilePath Handle Write) r
+-- > interpretWriteFile :: Members '[Resource, Embed IO] => InterpreterFor (Scoped FilePath Write) r
 -- > interpretWriteFile =
 -- >   interpretScoped allocator handler
 -- >   where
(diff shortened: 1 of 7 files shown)


KingoftheHomeless:
Hmm. This doesn't scream Correct and Right to me... I'm going to have to look into and experiment with this. If I can convince myself this is without issue I'm 100% for this change.

tek:
yeah it seems pretty evil :sweat_smile:

KingoftheHomeless:
I've convinced myself this is legit. Polish as you like, and I'll review once you feel the PR is ready for merging.

tek:
sweet! I removed the references to `resource` from the docs, I think it's good to go!

src/Polysemy/Internal/Scoped.hs
```diff
@@ -20,6 +19,8 @@ import Polysemy
 -- database effect.
 --
 -- For a longer exposition, see <https://www.tweag.io/blog/2022-01-05-polysemy-scoped/>.
+-- Note that the interface has changed since the blog post was published: The
+-- @resource@ parameter is no longer necessary.
```
KingoftheHomeless, line 23: `is no longer present`/`no longer exists`.

src/Polysemy/Internal/Scoped.hs
```diff
@@ -7,10 +7,9 @@ import Data.Kind (Type)
 
 import Polysemy
 
--- | @Scoped@ transforms a program so that @effect@ is associated with a
--- @resource@ within that program. This requires the interpreter for @effect@ to
--- be parameterized by @resource@ and constructed for every program using
--- @Scoped@ separately.
+-- | @Scoped@ transforms a program so that an interpreter for @effect@ may
+-- perform arbitrary actions, like resource management, before and after the
+-- computation wrapped by a call to 'scoped' is executed.
```
KingoftheHomeless, line 12: Now that `resource` doesn't exist for disambiguation between `Scoped`s, perhaps we should remind people of `Tagged`(?)

src/Polysemy/Scoped.hs
```diff
@@ -36,26 +36,26 @@ import Polysemy.Internal.Tactics
...
   -- argument that contains the call site parameter passed to 'scoped'.
   (∀ x . param -> (resource -> Sem r x) -> Sem r x) ->
   -- | A handler like the one expected by 'interpretH' with an additional
   -- parameter that contains the @resource@ allocated by the first argument.
   (∀ r0 x . resource -> effect (Sem r0) x -> Tactical effect (Sem r0) r x) ->
-  InterpreterFor (Scoped param resource effect) r
+  InterpreterFor (Scoped param effect) r
 interpretScopedH withResource scopedHandler =
-  go
+  go (errorWithoutStackTrace "top level run")
```
KingoftheHomeless, line 49: Leave a TODO here for us to investigate whether the loopbreaker optimization is worth it here (i.e. this pattern):
```
go :: ...
go = ... // same as existing but replace all recursive references to go with go1
{-# INLINE go #-}
go1 :: ...
go1 = go
{-# NOINLINE go #-}
```
Alternatively, if you know it's not worth it, say so and don't add that TODO.

src/Polysemy/Scoped.hs
```diff
@@ -36,26 +36,26 @@ import Polysemy.Internal.Tactics
 -- use the 'Tactical' environment and transforms the effect into other effects
 -- on the stack.
 interpretScopedH ::
-  ∀ param resource effect r .
+  ∀ resource param effect r .
```
KingoftheHomeless, line 39: Does this param order change make sense any longer now `resource` is no longer part of the effect type? Or perhaps I'm getting confused and that's exactly why you made the change? If the latter, please explain your reasoning.

KingoftheHomeless (changes requested):

src/Polysemy/Scoped.hs
tek: my only reason was that most of the others had this order.
I guess the resource must always be fixed in an interpreter, while the param may be polymorphic. should we reorder them all to have `param` as the first one? and `resource` as the second or `effect`?

src/Polysemy/Internal/Scoped.hs
tek: not sure what scenario you're thinking of here. can you formulate something?

KingoftheHomeless: Actually, considering the resource typed param should be kept polymorphic people couldn't disambiguate using it anyway. So nevermind.

src/Polysemy/Scoped.hs
KingoftheHomeless: I feel the order doesn't have as much meaning now. As the one most experienced with `Scoped`, it's up to you what the order should be.

KingoftheHomeless (approved):
I'll leave you to decide how you want the typed parameter order.

src/Polysemy/Scoped.hs
tek: I did a grep and I'm not using type applications anywhere in my projects for those interpreters…I think we can leave it like it is.

tek:
:grin:

PR #464, opened 2022-12-02
failing test for scoped with interpreter switch

tek:
@KingoftheHomeless can you explain why this fails? Shouldn't the higher-order computation arrive in the handler untouched by the scope machinery?

### DIFF
--- a/src/Polysemy/Scoped.hs
+++ b/src/Polysemy/Scoped.hs
@@ -47,17 +47,19 @@ interpretScopedH ::
   (∀ r0 x . resource -> effect (Sem r0) x -> Tactical effect (Sem r0) r x) ->
   InterpreterFor (Scoped param effect) r
 interpretScopedH withResource scopedHandler =
-  -- TODO investigate whether loopbreaker optimization is effective here
-  go (errorWithoutStackTrace "top level run")
+  interpretWeaving $ \(Weaving effect s wv ex _) -> case effect of
+    Run _ -> errorWithoutStackTrace "top level run"
+    InScope param main -> withResource param \ resource ->
+          ex
+      <$> interpretH (scopedHandler resource) (go $ raiseUnder $ wv (main <$ s))
   where
-    go :: resource -> InterpreterFor (Scoped param effect) r
-    go resource =
+    -- TODO investigate whether loopbreaker optimization is effective here
+    go :: InterpreterFor (Scoped param effect) (effect ': r)
+    go =
       interpretWeaving \ (Weaving effect s wv ex ins) -> case effect of
-        Run act ->
-          ex <$> runTactics s (raise . go resource . wv) ins (go resource . wv)
-            (scopedHandler resource act)
-        InScope param main ->
-          withResource param \ resource' -> ex <$> go resource' (wv (main <$ s))
+        Run act -> liftSem $ injWeaving $ Weaving act s (go . wv) ex ins
+        InScope param main -> raise $ withResource param \ resource ->
+          ex <$> interpretH (scopedHandler resource) (go $ wv (main <$ s))
 {-# inline interpretScopedH #-}
 
 -- | Variant of 'interpretScopedH' that allows the resource acquisition function
@@ -159,24 +161,25 @@ interpretScopedWithH ::
 interpretScopedWithH withResource scopedHandler =
   interpretWeaving \case
     Weaving (InScope param main) s wv ex _ ->
-      ex <$> withResource param \ resource -> inScope resource $
-        restack
-          (injectMembership
-           (singList @'[Scoped param effect])
-           (singList @extra)) $ wv (main <$ s)
+      ex <$> withResource param \ resource ->
+        interpretH (scopedHandler resource) $ inScope $
+          restack
+            (injectMembership
+             (singList @'[Scoped param effect])
+             (singList @(effect ': extra))) $ wv (main <$ s)
     _ ->
       errorWithoutStackTrace "top level Run"
   where
-    inScope :: resource -> InterpreterFor (Scoped param effect) r1
-    inScope resource =
+    inScope :: InterpreterFor (Scoped param effect) (effect ': r1)
+    inScope =
       interpretWeaving \case
         Weaving (InScope param main) s wv ex _ ->
-          restack (extendMembershipLeft (singList @extra))
-            (ex <$> withResource param \resource' ->
-                inScope resource' (wv (main <$ s)))
+          restack
+            (extendMembershipLeft (singList @(effect ': extra)))
+            (ex <$> withResource param \resource ->
+                interpretH (scopedHandler resource) $ inScope $ wv (main <$ s))
         Weaving (Run act) s wv ex ins ->
-          ex <$> runTactics s (raise . inScope resource . wv) ins (inScope resource . wv)
-            (scopedHandler resource act)
+          liftSem $ injWeaving $ Weaving act s (inScope . wv) ex ins
 {-# inline interpretScopedWithH #-}
 
 -- | First-order variant of 'interpretScopedWithH'.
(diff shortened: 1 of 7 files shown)


KingoftheHomeless:
Oh neat, this is actually a bug with `interpretScopedH`. The way it works is that it recursively applies the provided handler to all `Run`s, and then just raises the resulting Sem to `Sem (e ': r)` so the handler will accept it. So the higher-order computations the handler gets are all already processed, and that's why it doesn't work.

What the more reasonable semantics would be here is that usages of `scoped` delimit applications of `interpretH` leveraging the handler. So:
```
run (interpretScopedH scopeHO (handleHO 1) $ scoped_ @HO (inc ret)) == 2
```
but
```
run (interpretScopedH scopeHO (handleHO 1) $ scoped_ @HO (inc (scoped_ @HO ret))) == 1
```

And that's possible, I've cooked up a quick fix for `interpretScopedH` specifically which is probably safe. During lunch I'll spread the fix to all other `Scoped` interpreters and create a PR.

KingoftheHomeless:
... In fact. This gives me an idea. Why not get rid of `resource` _entirely_? `interpretScopedH` can look like this:
```
newInterpretScopedH ::
  ∀ param effect r .
  (∀ x . param -> InterpreterFor effect r) ->
  InterpreterFor (Scoped param effect) r
```
And then the definition of `interpretScopedH` can be rewritten to this:
```
interpretScopedH withResource scopedHandler =
  newInterpretScopedH $ \param sem ->
    withResource param $ \resource -> interpretH (scopedHandler resource) sem
```

Our understanding of `Scoped` really does grow every day!

KingoftheHomeless:
... oh my god this solves abstract effect interpretation. Holy shit. A shame this can't be applied to `in-other-words`.

KingoftheHomeless:
Actually, it can _maybe_, _kind of_ be applied to `in-other-words`. This has given me a lot to think about, but I'll have to save it until the weekend.

KingoftheHomeless:
```

**Comment by KingoftheHomeless** [comment]
My celebration that it solves abstract effect interpretation may have been premature, after some experimentation. It comes close, though.

Abstract effect interpretation is the ability to say "I want to interpret this effect I'm using locally in application code, using some interpreter provided to me from the effect stack rather than my own." The most typical example of this is someone wanting to interpret a state or error effect locally, but instead of hard-coding the interpreter used  -- by e.g. using `runState` or -- god forbid -- `stateToIO` -- it is provided the interpreter from the context, via an effect from the effect stack. My rethink of `interpretScopedH` means `Scoped` _almost_ be used for this, but the trouble lies in the return type;  `State`, `Error`, and other such interpreters change the return type, which isn't possible with `Scoped` as it currently is. I experimented with:
```
data Scoped (param :: Type) (effect :: Effect) (modifier :: Type -> Type) :: Effect where
  Run :: ∀ param effect m a . effect m a -> Scoped param effect m a
  InScope :: ∀ param effect m a . (modifier a -> x) -> param -> m a -> Scoped param effect m x
```
... but I haven't managed to successfully interpret this yet, the `modifier` gets tangled up with the effectful state.

Still... my rethink is very powerful as-is. All of the various interpreters offered in Scoped can _all become redundant_ in the face of this one:
```runScoped ::
  ∀ param effect r .
  (param -> InterpreterFor effect r) ->
  InterpreterFor (Scoped param effect) r
runScoped innerInterpreter =
  interpretWeaving $ \(Weaving effect s wv ex _) -> case effect of
    Run _ -> errorWithoutStackTrace "top level run"
    InScope param main ->
      ex <$> innerInterpreter param (go $ raiseUnder (wv (main <$ s)))
  where
    go :: InterpreterFor (Scoped param effect) (effect ': r)
    go = interpretWeaving $ \(Weaving effect s wv ex ins) -> case effect of
      Run act -> liftSem $ injWeaving $ Weaving act s (go . wv) ex ins
      InScope param main ->
        ex <$> raise (innerInterpreter param (go (wv (main <$ s))))
{-# inline runScoped #-}
```
This can do bloody everything. Please point me to any of your usages of any of the `Scoped` interpreters and I will show you how.

KingoftheHomeless:
Hah! Interpreting `data Scoped (param :: Type) (effect :: Effect) (modifier :: Type -> Type) :: Effect` is possible _with the core rewrite_ together with the constraint `Traversable modifier`!
We have abstract effect interpretation!

KingoftheHomeless:
Look at it!:
```haskell
scoped :: forall param effect modifier r x
        . Member (Scoped param effect modifier) r
       => param -> Sem (effect ': r) x -> Sem r (modifier x)
scoped param main =
  send $ InScope @param @effect @modifier id param do
    transform @effect (Run @param @_ @modifier) main
{-# INLINE scoped #-}

runScoped :: forall param effect modifier r
           . Traversable modifier
          => (forall x. param -> Sem (effect ': r) x -> Sem r (modifier x))
          -> InterpreterFor (Scoped param effect modifier) r
runScoped innerInterpreter = interpretH $ \case
    Run _ -> errorWithoutStackTrace "top level run"
    InScope conv param main -> do
      Processor pr <- getProcessorH
      res <- raise $ innerInterpreter param (go $ raiseUnder $ pr main)
      conv <$> traverse restoreH res
  where
    go :: InterpreterFor (Scoped param effect modifier) (effect ': r)
    go = interpretH $ \case
      Run e -> propagate e
      InScope conv param main -> do
        Processor pr <- getProcessorH
        res <- raise $ raise $ innerInterpreter param (go $ pr main)
        conv <$> traverse restoreH res
{-# INLINE runScoped #-}
```
It's beautiful!
```haskell
type StateProvider s = Scoped s (State s) ((,) s)

runStateViaProvider :: Member (StateProvider s) r
                    => s
                    -> Sem (State s ': r) a
                    -> Sem r (s, a)
runStateViaProvider = scoped

stateProviderToIO :: Member (Embed IO) r
                  => Sem (StateProvider s ': r) a
                  -> Sem r a
stateProviderToIO = runScoped stateToIO

runStateProvider :: Sem (StateProvider s ': r) a
                 -> Sem r a
runStateProvider = runScoped runState
```

KingoftheHomeless:
That `traverse restoreH` though... messes with local/global state semantics, will lead to some strange interactions people familiar with `MonadBaseControl` will recognize. Oh well, it's the sacrifice you'll have to make.

KingoftheHomeless:
i can't believe after so much time spent on trying to figure out abstract effect interpretation suddenly i just stumble across the solution _years_ later. what the hell

KingoftheHomeless:
this certainly motivates me to finish the core rewrite. Wow.

tek:
:joy: I'm always amazed how some completely banal question I give you unlocks an enormous torrent of creativity in your brain

tek:
wow, that is really bananas

tek:
so we can `runError` with exceptions without having `Embed IO` at the call site?

KingoftheHomeless:
> so we can `runError` with exceptions without having `Embed IO` at the call site?

_YES!!!!!!!!!!!!!!!!!_

tek:
ok the `ErrorProvider` would be an additional effect though, but still

KingoftheHomeless:
Well, "without an additional effect" -- you'll need the effect interpreter provider.
```
type ErrorProvider e = Scoped () (Error e) (Either e)

runErrorViaProvider :: Member (ErrorProvider e) r
                    => Sem (Error e ': r) a
                    -> Sem r (Either e a)
runStateViaProvider = scoped ()

errorProviderToIO :: Member (Final IO) r
                  => Sem (ErrorProvider e ': r) a
                  -> Sem r a
errorProviderToIO = runScoped (\_ -> errorToIOFinal)

runErrorProvider :: Sem (ErrorProvider e ': r) a
                 -> Sem r a
runErrorProvider = runScoped (\_ -> runError)

**Comment by tek** [comment]
right, right

**Comment by tek** [comment]
I was thinking about introducing this pattern for `Resumable`, since it uses `ExceptT` everywhere, but I haven't attempted anything concrete yet. so that's nice :smile:

**Comment by tek** [comment]
why does the `modifier` have to be concrete in `StateProvider`?

**Comment by tek** [comment]
oh right, you're using it right away

**Comment by KingoftheHomeless** [comment]
ofc. there's still that type variable... `ErrorProvider e` and `StateProvider s`, that means having a bunch of provider effects for each unique state/error type you want to interpret locally... I may have something in mind to solve that, but it'll have to wait until I actually have the time to sit down with it, I've already spent too much of my working time on this, and will have to work a bit late to compensate.

**Comment by KingoftheHomeless** [comment]
I was thinking that in the meantime I'm brainstorming a complete rewrite of `Scoped`, I should plug this issue. Funny thing is I've had no issues fixing every interpreter _but one_ -- `interpretScopedH'`. Having `Tactical` thrown into the `withResource`, together with the _very weird_ signature of the `scopedHandler`, makes it much more difficult to apply the fix to it. I'm thinking of leaving that interpreter as-is and leaving a note -- how often do you use `interpretScopedH'` @tek ?

**Comment by tek** [comment]
I think I used it once but I can't find any reference of it anymore in my projects. no objection to getting rid of it

**Comment by KingoftheHomeless** [comment]
oh, pff, actually, no need to deprecate `interpretScopedH'`, the way it was written made it so the bug can't apply to it in the first place, precisely because its `scopedHandler` has the signature that it does.

**Comment by KingoftheHomeless** [comment]
@tek It's 1 regardless.

> 

oh no.

**Comment by KingoftheHomeless** [comment]
It being `1` is wrong. I know exactly why.

_Oh no._

**Comment by KingoftheHomeless** [comment]
so uhhhh this is why it was important for `resource` to be part of `Scoped`. I... completely missed this. Fuck.

**Comment by KingoftheHomeless** [comment]
This throws _quite the wrench_ into my plans. Ok, I need to brainstorm _even harder_ now to see if I can _possibly_ salvage this.

**Comment by tek** [comment]
oh no :see_no_evil:

**Comment by KingoftheHomeless** [comment]
Yes, it's salvageable. I think. You need to change `Scoped` to have an equivalent of the previously removed `resource` parameter, so you can keep track of what `Run` corresponds to what use of `InScope`. Fortunately, _I think and Hope_ you can keep track of this solely through the depth, so `Scoped` needs to be changed to:
```
data Scoped (param :: Type) (effect :: Effect) :: Effect where
  Run :: ∀ param effect m a . Word -> effect m a -> Scoped param effect m a
  InScope :: ∀ param effect m a . param -> (Word -> m a) -> Scoped param effect m a
```
And then what the `Scoped` interpreters do is that the `Word` they give to `InScope` -- and is thus used when `scoped` creates `Run`s -- corresponds to how deeply nested the current `InScope` is inside other `InScope`s. This allows the interpreter to tell `Run`s belonging to different `InScope`s apart, and the bug is fixed.

Two things to note here:
1. I may be wrong with the depth being enough. Perhaps `Unique`s are necessary, which will require us to do `unsafePerformIO`.
2. The signature of my Improved `runScoped` I showed off before doesn't _quite_ escape unscathed -- it needs a way to propagate `Run`s _not_ part of the current use of `scoped` the inner interpreter is currently running on, so the tail can't remain as purely `r`. Meaning `runScoped` will instead look like this:
```
runScoped ::
  ∀ param effect r .
  (forall e. param -> InterpreterFor effect (Opaque e ': r)) ->
  InterpreterFor (Scoped param effect) r
``` 
I don't think this will be that annoying in practice, but perhaps @tek will differ.

**Comment by KingoftheHomeless** [comment]
Forgot to say, I wrote up a _messy_ definition of `runScoped` using this principle that seems to work. But I have not tested it very extensively and it is too messy for me to be comfortable to share it.

**Comment by KingoftheHomeless** [comment]
what a bloody rollercoaster of emotions these last two days have been

**Comment by KingoftheHomeless** [comment]
well. I'll put my enthusiasm re. abstract effect interpretation on the backburner for the moment in favor of trying to make sure we can find a way to make `Scoped` not broken. If I'm wrong about the depth/unique-based approach, then we may be forced to revert to having the resource type parameter again.

**Comment by tek** [comment]
took me about four hours of type tetris but I managed to implement [interpretScopedResumableH](https://github.com/tek/polysemy-conc/blob/scoped-fix/packages/conc/lib/Polysemy/Conc/Interpreter/Scoped.hs#L285) with the new pattern :sweat_smile: not very pretty yet, but it works

**Comment by tek** [comment]
I converted all `Scoped`/`Resumable` interpreters to use your solution, tested in on my original use case, and everything is working out! If you have nothing to add, we can merge this.

File: src/Polysemy/Scoped.hs [hunk 1]
```diff
@@ -226,46 +229,34 @@ interpretScopedWith_ withResource scopedHandler =
...
--- The proper way to implement this would be to rewrite the resource allocation:
+-- /Note/: In previous versions of Polysemy, the wrapped interpreter was
+-- executed fully, including the initializing code surrounding its handler,
+-- for each action in the program. However, new and continuing discoveries
+-- regarding 'Scoped' has allowed the improvement of having the interpreter be
+-- used only once per use of 'scoped', and have it cover the same scope of
+-- actions that 'withResource' does.
 --
--- > runScoped (\ initial use -> use =<< embed (newTVarIO initial)) runAtomicStateTVar
+-- This renders 'withResource' practically redundant; for the moment, the API
```
KingoftheHomeless, line 239: wait is calling the first argument `withResource` established anywhere in any docs re. `Scoped`?
also this should be `@withResource@`, not `'withResource'`, my mistake

tek: one example creates a function of that name, but I guess that's not very explicit

tek: changed it to "the resource allocator"

PR #466, opened 2022-12-27
Fix scoped within scoped incoherency. Add `runScopedNew`

KingoftheHomeless:
The big question here is the `Opaque` pollution, which is breaking. There _is_ a way around that if we _desperately_ want to avoid it, but it won't be forward compatible with my Vision Of The Future Of `Scoped`, where `runScopedNew` (or something in its like) will be the primary `Scoped` interpreter.

### DIFF
--- a/src/Polysemy/Opaque.hs
+++ b/src/Polysemy/Opaque.hs
@@ -36,9 +36,10 @@ import Polysemy
 -- containing some polymorphic effect variables. By wrapping the polymorphic
 -- effect variables using 'Opaque', users of the function can use effects as
 -- normal, without having to use 'raise' or 'Polysemy.Membership.subsumeUsing'
--- in order to have 'Polysemy.Member' resolve.
+-- in order to have 'Polysemy.Member' resolve. The various interpreters of
+-- 'Polysemy.Scoped.Scoped' are examples of such usage of 'Opaque'.
 --
--- @since TODO
+-- @since 1.9.0.0
 newtype Opaque (e :: Effect) m a = Opaque (e m a)
 
 -- | Wrap 'Opaque' around the top effect of the effect stack
--- a/src/Polysemy/Opaque.hs
+++ b/src/Polysemy/Opaque.hs
@@ -1,0 +1,52 @@
+module Polysemy.Opaque (
+  -- * Effect
+  Opaque(..),
+
+  -- * Interpreters
+  toOpaque,
+  fromOpaque,
+  ) where
+
+import Polysemy
+
+-- | An effect newtype meant to be used to wrap polymorphic effect variables to
+-- prevent them from jamming up resolution of 'Polysemy.Member'.
+-- For example, consider:
+--
+-- @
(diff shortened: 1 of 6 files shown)


src/Polysemy/Scoped.hs
```diff
@@ -73,26 +71,28 @@ interpretScopedH' ::
     Tactical (Scoped param effect) (Sem r0) r x) ->
   InterpreterFor (Scoped param effect) r
 interpretScopedH' withResource scopedHandler =
-  go (errorWithoutStackTrace "top level run")
+  go 0 Empty
   where
-    go :: resource -> InterpreterFor (Scoped param effect) r
-    go resource =
+    go :: Word -> Seq resource -> InterpreterFor (Scoped param effect) r
+    go depth resources =
```
KingoftheHomeless, line 77: This is how we can avoid `Opaque` pollution in the existing interpreters; we crop this pattern, of having a `Seq` (or `Vector`) of the allocated resources, and a `Run`'s depth is used as an index to acquire the correct resource.

That, however, means the existing interpreters can't be written in terms of `runScopedNew`. I'm honestly leaning towards applying this pattern to preserve backwards compatibility, but one day we _will_ scrap all these interpreters in favor of `runScopedNew`, and there's an argument to be made that we don't want to encourage people to keep using the old ones because they have _one_ advantage (no `Opaque` pollution) compared to `runScopedNew`.

src/Polysemy/Opaque.hs
```diff
@@ -0,0 +1,53 @@
...
+--
+-- @since 1.9.0.0
+newtype Opaque (e :: Effect) m a = Opaque (e m a)
+
+-- | Wrap 'Opaque' around the top effect of the effect stack
+toOpaque :: Sem (e ': r) a -> Sem (Opaque e ': r) a
+toOpaque = rewrite Opaque
+{-# INLINE toOpaque #-}
+
+-- | Unwrap 'Opaque' around the top effet of the effect stack
```
KingoftheHomeless, line 50: oh NO i MISSPELLED

tek (approved):

src/Polysemy/Scoped.hs
```diff
@@ -237,35 +244,85 @@ interpretScopedWith_ withResource scopedHandler =
...
+-- | Run a 'Scoped' effect by specifying the interpreter to be used at every
+-- use of 'scoped'.
+--
+-- This interpretation of 'Scoped' is powerful enough to subsume all other
+-- interpretations of 'Scoped' (except 'interpretScopedH'' which works
+-- differently from all other interpretations) while also being much simpler.
+--
+-- Consider this a sneak-peek of the future of 'Scoped'. In the API rework
+-- planned for 'Scoped', the effect and its interpreters will be further
+-- expanded make 'Scoped' even more flexible.
```
KingoftheHomeless, line 278: missed a word. "expanded to"

src/Polysemy/Opaque.hs
tek: c'est francais!

PR #468, opened 2022-12-29
Allow inspection-testing 0.5

felixonmars:
Builds fine and all tests pass.

(I forgot to update also `package.yaml` and for `polysemy-plugin`. Both are fixed this time.)


tek:
thanks

PR #469, opened 2022-12-29
Core rewrite, a lot of V2 breaking and non-breaking changes

KingoftheHomeless:
Creating a PR now as I can now say with certainty that the interpretation API is final (except the naming of `Handling`/`Strategy`), and I've added all documentation. 

I'm gonna create another PR expanding upon this one featuring `Scoped`-As-I-Want-It-To-Be, whose implementation will also show why I need the `rPre` and `rPost` separation in the `Handling` environment. By the way, check out `Handling`! `subsume_` pulls a lot of heavy duty here, I love it! It solves all differences in opinion that me and @tek had re. flexibility of `runH` vs. flexibility of `liftWithH` without any compromises! It's even better than the solution I had before, _and_ it can not only be used to make `runH` flexible, but also `runH'`, `runExposeH`, and `runExposeH'`! @TheMatten I'd like to take this moment to thank you profusely for `subsume_`, it shouldn't have taken me two years to realize how good it is.

The doctests fail for me on 9.0.2, it fails to match upon the custom `insertAt` error message that is active on that version for some reason. Not related to this PR, it's just a curiosity.


KingoftheHomeless:
Oh, the doctest failure is this PR's fault, my bad, it was struck by the `raiseUnderN`/`raiseNUnder` renaming and I didn't catch it due to how doctest works.

KingoftheHomeless:
Closed in favor of #473

PR #470, opened 2022-12-30
WIP Scoped-As-I-Want-It-To-Be

KingoftheHomeless:


src/Polysemy/Internal/HigherOrder.hs
```diff
@@ -50,6 +51,25 @@ data HigherOrder z t e rPre rPost :: Effect where
   RestoreH
     :: forall z t e rPre rPost m a. t a -> HigherOrder z t e rPre rPost m a
 
+-- | A singleton datatype parametrized with type parameters corresponding to
```
googleson78, line 54:
```suggestion
-- | A proxy datatype parametrized with type parameters corresponding to
```

I think in Haskell "singleton" has the connotation of meaning "this type has exactly as many terms in it as there are type params for it/you can recover information about the type params by matching on this type"

KingoftheHomeless: Saying singleton here _is_ technically correct as singleton just means "a type with only one member". You're right however that the term "singleton" in Haskell is more commonly used to refer to the more specific "a type with only one member which you can pattern match on to receive information about the type", so I agree "proxy" is better.

KingoftheHomeless:
Closed in favor of #473

PR #473, opened 2023-01-07
Canonical branch for core rewrite, Meta, compile-time optimization

KingoftheHomeless:


KingoftheHomeless:
My skeleton of `Scoped` lacks the documentation of the old one. Both the docs of the old Scoped and WIP docs of the `Scoped1` (#470) could be used as a basis.

src/Polysemy/Meta.hs
```diff
@@ -0,0 +1,300 @@
...
+  ProcessMeta
+    :: forall eff z metaeffect t rH l m a
+     . Unique
+    -> ElemOf '(z, eff) l
+    -> t ()
+    -> z a
+    -> HandlingMeta metaeffect t rH l m
+        (Sem (Meta metaeffect ': HandlingMeta metaeffect t rH l ': rH) (t a))
+
+data Box where
```
googleson78, line 72: Is the value in the box ever used?

KingoftheHomeless: So `Box` is just what's used to carry an argument to `newUnique'`. `newUnique'` has an argument to ensure that the `unsafePerformIO (newUnique' ...)` _absolutely,  definitively_ does not get floated out of `processMeta`. If we used `newUnique` instead:
```haskell
processMeta pr t z = do
    let !uniq = unsafePerformIO newUnique
    ....
```
then GHC is free to transform this to:
```haskell
_newUnique :: Unique
_newUnique = unsafePerformIO newUnique

processMeta pr t z = do
    let !uniq = _newUnique
    ....
```
which is bad. the `NOINLINE` to `newUnique'` is so that it doesn't inline `newUnique' (Box z)`, to `newUnique` and is _then_ free to do the above.

Finally, providing and pattern-matching on `Box` instead of a generic, unevaluated `a` is just a last, extremely paranoid measure on my part: it avoids a theoretical GHC optimization that would detect if a function is always lazy in its arguments, and then rewrite `newUnique'` to:
```haskell
_newUnique' :: IO Unique
_newUnique' = newUnique
{-# NOINLINE newUnique'_ #-}

newUnique' :: a -> IO Unique
newUnique' _ = _newUnique'
{-# INLINE newUnique' #-}
```
... because then `unsafePerformIO (newUnique' z)`  would get inlined to `unsafePerformIO _newUnique', and then can get floated out.

tek: did you observe it being floated out without those measures?

PR #478, opened 2023-04-06
support GHC 9.6

tek:
needs https://github.com/isovector/type-errors/pull/7

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -57,11 +57,11 @@
     },
     "unstable": {
       "locked": {
-        "lastModified": 1672057183,
-        "narHash": "sha256-GN7/10DNNvs1FPj9tlZA2qgNdFuYKKuS3qlHTqAxasQ=",
+        "lastModified": 1680728981,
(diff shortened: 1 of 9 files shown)


tek:
`type-errors` is released, so this should be ready.
I'm wondering, is it common practice to use a conditional dependency when only the latest version of a dep works with a given GHC, but the library works with older versions?

googleson78:
If I'm understanding your question correctly, I believe a similar situation occurs in HLS, e.g. see the fourmolu plugin - https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-fourmolu-plugin/hls-fourmolu-plugin.cabal

But HLS might be considered a special case, so idk.

tek:
hmmmm. guess the question is whether the package set for 9.6 in stackage or nixpkgs will contain the latest version. which should probably be true for stackage, and therefore eventually also for nixpkgs. but even without a stackage resolver for 9.6, nixpkgs contains that version. not sure what the type-errors version will be then…but it probably won't matter much in any case what we specify here.

eh!

PR #481, opened 2023-09-15
Allow th-abstraction 0.6

Vekhir:
Builds and tests fine


tek:
thanks! you'll need to change the version in `package.yaml` as well though.

Vekhir:
Oh, right. I've updated `package.yaml`

